PR #11, opened 2016-08-12
fix: getObject for timestamp value  in ClickHouseResultSet

andr83:
There is a wrong getter getObject for timestamp value in ClickHouseResultSet.java. It return java.sql.Time instead java.sql.Timestamp


serebrserg:
Thanks.

PR #14, opened 2016-08-16
pass database from jdbc url to Driver

amaslak:
Default database name from jdbc url is not used by ClickHouseStatementImpl if connection was created using ClickHouseDriver.
```
Driver driver = new ClickHouseDriver();
Connection connection = driver.connect("jdbc:clickhouse://localhost:8123/test", new Properties());
Statement statement = connection.createStatement();
ResultSet rs = statement.executeQuery("select * from table limit 1"); // fails, but works with test.table
while (rs.next()) {
    System.out.println(rs.getMetaData().getColumnCount());
}
statement.close();
```


serebrserg:
Thanks for your help.
I did this a bit different to avoid modification of input properties object.
You can see it here https://github.com/yandex/clickhouse-jdbc/commit/5a4214a3edc5e710cb585486e07c9c8471c3d5de

PR #18, opened 2016-08-28
Inproved performance in batch operations

AndreevDm:


serebrserg:
Thanks. 
It is better to move the tests not using flights data set to a separate class. I'll rearrange them.

PR #19, opened 2016-09-07
Skip jdbc url to another DB

Somewater:
If project includes a series of database drivers it gets a problem because ClickHouse driver tries to handle jdbc url to another DB.


serebrserg:
Thanks. I think it is better to use acceptsURL instead of explicit substring checking.

Somewater:
Yes, I fix it

PR #20, opened 2016-09-07
Ability to use driver with Pentaho Report Designer

DrStep:
Changes that makes driver work correctly with Pentaho:
Added "statement"  attribute to ClickHouseResultSet and all constructors.
Added stub to  prepareStatement(String sql, int resultSetType, int resultSetConcurrency)


serebrserg:
Thanks.

PR #23, opened 2016-09-13
Driver.connect should return null for unsupported urls

zhurs:
According to java.sql documentation, connect should return null for foreign urls, current behaviour is throwing "java.lang.IllegalArgumentException: Incorrect ClickHouse jdbc url: ..." 
http://docs.oracle.com/javase/8/docs/api/java/sql/Driver.html#connect-java.lang.String-java.util.Properties-

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -81,6 +81,12 @@
             <artifactId>guava</artifactId>
             <version>19.0</version>
         </dependency>
+
+        <dependency>
+            <groupId>org.slf4j</groupId>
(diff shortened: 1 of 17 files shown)


serebrserg:
Thanks. 
I've already merged another PR on the problem, which introduced conflict with yours, but the test from yours is also useful. Could you update your PR? Then I'll merge it.

zhurs:
I will make another PR with test

PR #27, opened 2016-09-30
Support IPv6 in JDBC URL

amaslak:
[RFC 2732](http://www.ietf.org/rfc/rfc2732.txt) IPv6 Literal Addresses in URL's


serebrserg:
Thanks

PR #41, opened 2016-10-27
Add enum binding parameter support

popfalushi:
It is nice to be able to pass enum as a parameter for sql query. This minifix add's this possibility. Enum is transformed to String, because it is recommended way of saving enums to database since Archimedes. Have never seen anyone who uses ordinal for this.


popfalushi:
As it turns out, it is standard for JDBC drivers to not expect enums in binding parameters.

PR #45, opened 2016-10-31
Add check values in result set

KhurtinDN:


serebrserg:
This won't work for single-column resultset, however it is better than nothing. Thanks.

PR #68, opened 2017-02-15
Fix setting properties for datasource with ClickHouseProperties class

rasmus93:
In one of commits you change getting properties from `get` to `getProperty`. And it brakes setting properties with `ClickHouseProperties` class.

**Reason of this:**
In `getProperty` method there is checking that value `instanceof` String. But in `ClickHouseProperties` class you have properties(connection_timeout and other) that have time `int`, `boolean` etc. So for these values when you call `getProperty` method you will get `null` and default value would be set(like in test).

`String.valueOf` needed only for variables with primitive types, because objects can be null.


serebrserg:
Thanks. Included in 0.1.17.

PR #87, opened 2017-04-10
Allow to set boolean params as 0/1

nettoyeur:
In documentation bool params have 0/1 values but in Java I should keep in mind that they need to be replaced with false/true values.
This patch allows using 0/1 in boolean params and correctly parse them.


serebrserg:
Thanks.

PR #92, opened 2017-04-19
BalancedDataSource

Cadovvl:
Штука принимает на вход несколько хостов в jdbc урле и рандомно создает коннекшены к ним. 
Метод actualize вызывает пинг для каждого урла и неответившие переносит в disables, ответившие из disabled переносит в список урлов. 
Можно запустить шедулинг этих актуализаций, по дефолту вырублен.


Cadovvl:
Ой.... Я хотел сначала в свою репу пулл-реквест запилить, и между собой е8го обсудить. Не посмотрел, что в мастер Яндекса создаю реквест, сорри. 

Мне его пересоздать лил как?

serebrserg:
Можно пересоздать, или поставить пометку [WIP] в имени, а потом когда будет готово, убрать её и отдельным комментарием написать что можно смотреть.

Cadovvl:
Вроде теперь работает. Можно ревьюить. Если есть еще предложения - буду рад выслушать. 
Насчет добавления решения для Distributed таблиц (выполнить запрос по крайней мере на M машинах и N), то это можно потом отдельно записать. Эта штука позволит хотябы читать из реплик...

serebrserg:
Спасибо. Далее буду писать на английском чтобы все заинтересованные могли найти это обсуждение.
The implementation looks good, but use cases are not completely clear for me.
As I understood to have load balancing one have to retrieve new connection every time, which in current implementation of "connection" means creation of new http client with its own http connection pool.
I assume it might be better to create a "pool" of connections in the datasource, one for each url, and return a random instance from the pool on getConnection() call because current ClickHouseConnection is thread safe.

Both options are possible, so I don't mind your current solution if you carefully describe its properties in javadoc.

Cadovvl:
> current ClickHouseConnection is thread safe.
If it is, indeed it looks better to have a connection pool inside. 
But original ClickHouseDatasource creates new connection on 'getConnection' call. 
```java
    @Override
    public Connection getConnection() throws SQLException {
        return driver.connect(url, properties);
    }

    @Override
    public Connection getConnection(String username, String password) throws SQLException {
        return driver.connect(url, properties.withCredentials(username, password));
    }
```
I thought, that there was a purpose for that implementation. 

Maybe it would better to implement a connection pool, which takes a DataSource in constructor, generates list of connections from DataSource, randomly shuffle it and return connections in that order. 

Let's make a choose, and I'll implement it.

serebrserg:
> Maybe it would better to implement a connection pool, which takes a DataSource in constructor, generates list of connections from DataSource, randomly shuffle it and return connections in that order.
Yes, I think this kind of pool will be the easiest solution to understand and to use.

src/main/java/ru/yandex/clickhouse/BalancedClickhouseDataSource.java
```diff
@@ -0,0 +1,234 @@
...
+            throw new IllegalArgumentException(e);
+        }
+
+        for (final String url : urls) {
+            addUrl(url);
+        }
+    }
+
+    private void disableUrl(final String url) {
+        synchronized (urls) {
```
izebit: Why did you use synchronized blocks? I think that, usage of collections from concurrent package is better than  usage of synchronization.

src/main/java/ru/yandex/clickhouse/BalancedClickhouseDataSource.java
```diff
@@ -0,0 +1,234 @@
...
+    private static final org.slf4j.Logger log = LoggerFactory.getLogger(BalancedClickhouseDataSource.class);
+    private static final Pattern URL_TEMPLATE = Pattern.compile(JDBC_CLICKHOUSE_PREFIX + "//([a-zA-Z0-9_:,.]+)(/[a-zA-Z0-9_]+)?");
+
+    protected PrintWriter printWriter;
+    protected int loginTimeout = 0;
+    protected final ClickHouseDriver driver = new ClickHouseDriver();
+
+    private Random rnd = new Random();
+    private Set<String> disabledUrls = new HashSet<String>();
+    private Set<String> urls = new HashSet<String>();
```
izebit: You have written:
`synchronized (urls) {
   ...
}`
but field "urls" is not final. It seems to me that is a mistake.

src/main/java/ru/yandex/clickhouse/BalancedClickhouseDataSource.java
```diff
@@ -0,0 +1,234 @@
...
+                } catch (Exception e){
+                    log.error("Unable to actualize urls: " + e);
+                }
+            }
+        }, 0, rate, timeUnit);
+    }
+
+    private static class ScheduledActualizer {
+        private static final ScheduledExecutorService INSTANCE = Executors.newSingleThreadScheduledExecutor();
+    }
```
izebit: SingleThreadPool is already defined in ClickHouseDriver. I think that one thread is enought for jdbc driver.

src/main/java/ru/yandex/clickhouse/BalancedClickhouseDataSource.java
```diff
@@ -0,0 +1,234 @@
...
+import java.util.regex.Pattern;
+
+import static ru.yandex.clickhouse.ClickhouseJdbcUrlParser.JDBC_CLICKHOUSE_PREFIX;
+
+public class BalancedClickhouseDataSource implements DataSource {
+    private static final org.slf4j.Logger log = LoggerFactory.getLogger(BalancedClickhouseDataSource.class);
+    private static final Pattern URL_TEMPLATE = Pattern.compile(JDBC_CLICKHOUSE_PREFIX + "//([a-zA-Z0-9_:,.]+)(/[a-zA-Z0-9_]+)?");
+
+    protected PrintWriter printWriter;
+    protected int loginTimeout = 0;
```
izebit: It is not important remark. I suggest to rename this field  "loginTimeoutSeconds". it will look better.

src/main/java/ru/yandex/clickhouse/BalancedClickhouseDataSource.java
Cadovvl: I use synchronization because there is no need lock on read, only on write. But, probably, you are right. ConcurrentHashMap seems to be a better solution.

Cadovvl: Yes, probably, it was a serious mistake, I'll fix it.

src/main/java/ru/yandex/clickhouse/BalancedClickhouseDataSource.java
Cadovvl: It means, I have to do it "public" in a ClickhouseDriver. If it is acceptable, ok...

Cadovvl:
Add a connection pool. Remove second thread pool. Rewrite BalancedDataSource with concurrent container. 

repair some minor mistakes.

Cadovvl:
Looks like there is a mistake in a current realization of ConnectionPool.

Cadovvl:
Temporary remove a ConnectionPool. I'll create a separate pool request for it after fixing all troubles.

Cadovvl:
?

serebrserg:
Ok, let it be a separate PR. 
I don't mind to merge this PR. Could you add a javadoc for the BalancedClickhouseDataSource to describe its properties?

izebit:
I will try to write javadoc for this interface, but my english not very good, i will be very glad if you correct my possible mistakes)

Issue #95, opened 2017-05-04
Support ssl/tls/https connections


marcelcorso:
Right now there is no way (or we didn't find a way on the docs or by reading the code) to connect to clickhouse servers using ssl/tls.

Maybe because the protocol "http" is hardcoded here:

https://github.com/yandex/clickhouse-jdbc/blob/8d2e9fd7f1f6342e9baaa4b8e7a601d29afd1e8d/src/main/java/ru/yandex/clickhouse/ClickHouseStatementImpl.java#L544

and then the wrong type of factory is created here:

https://github.com/yandex/clickhouse-jdbc/blob/1d7561c23feae11eda13165ed08232bf5bac2988/src/main/java/ru/yandex/clickhouse/util/ClickHouseHttpClientBuilder.java#L48

It could be enabled by passing params on the jdbc connection url like:

`String url = "jdbc:clickhouse://clickhouse-000.example.com:8443/default?ssl=true&sslrootcert=ca.pem";`

like postgresql.

serebrserg:
The PR was released in 0.1.21.

PR #96, opened 2017-05-04
added the support of SSL/TLS connections.

dysosmus:
Related to the issue #95 
The new driver properties are :
 - ssl(true|false) to enable/disable ssl/tls
 - sslrootcert an ssl/tls file name of the root ssl certificate
 - sslmode the ssl verify mode (none|strict)


serebrserg:
Thanks.

Issue #107, opened 2017-06-15
ClickHouseRowBinaryStream doesn't support arrays


genuss:
There is a lack of implementation. I tried to implement it by myself but got no luck. Here is a sample method for string array:
````java
public void writeArray(String[] array) throws IOException {
        Preconditions.checkNotNull(array);
        writeUnsignedLeb128(array.length);
        for (String el : array) {
            writeString(el);
        }
}
````
With this implementation I get ``DB::Exception: Cannot read all data`` from clickhouse server version 1.1.54236.
Do you see what can be wrong with this code?

serebrserg:
@AndreevDm , do you see what needs to be done to support arrays in RowBinary?

AndreevDm:
@genuss you implementation is correct. Where is a small bug in writeString() function, i'll try to fix it today.

genuss:
Hah, today rechecked and it looks like my fault and this implementation actually works. Would you like me to submit a PR for all arrays string or numbers?

PR #108, opened 2017-06-21
Add arrays support

genuss:
Here you are. Complete (I hope) support of arrays in clickhouse-jdbc. This solves #107.
I'm relatively new to the codebase so feel free to comment/request for changes in my PR.

### DIFF
--- a/src/main/java/ru/yandex/clickhouse/response/ByteFragmentUtils.java
+++ b/src/main/java/ru/yandex/clickhouse/response/ByteFragmentUtils.java
@@ -15,12 +15,6 @@ public final class ByteFragmentUtils {
 
     public static final char ARRAY_ELEMENTS_SEPARATOR = ',';
     public static final char STRING_QUOTATION = '\'';
-    public static final DateFormat DATE_FORMAT = new SimpleDateFormat("yyyy-MM-dd");
-    public static final DateFormat DATE_TIME_FORMAT = new SimpleDateFormat("yyyy-MM-dd HH:mm:ss");
-    static {
-        DATE_FORMAT.setTimeZone(TimeZone.getTimeZone("UTC"));
-        DATE_TIME_FORMAT.setTimeZone(TimeZone.getTimeZone("UTC"));
-    }
 
     private ByteFragmentUtils() {
     }
@@ -151,12 +145,20 @@ public final class ByteFragmentUtils {
     }
 
     public static Object parseArray(ByteFragment value, Class elementClass) {
+        return parseArray(value, elementClass, null);
+    }
+
+    public static Object parseArray(ByteFragment value, Class elementClass, SimpleDateFormat dateFormat) {
         if (value.isNull()) {
             return null;
         }
 
         if (value.charAt(0) != '[' || value.charAt(value.length() - 1) != ']') {
             throw new IllegalArgumentException("not an array: " + value);
+        }
+
+        if ((elementClass == Date.class || elementClass == Timestamp.class) && dateFormat == null) {
+            throw new IllegalArgumentException("DateFormat must be provided for date/dateTime array");
         }
 
         ByteFragment trim = value.subseq(1, value.length() - 2);
@@ -202,7 +204,7 @@ public final class ByteFragmentUtils {
                 } else if (elementClass == Date.class) {
                     Date dateValue;
                     try {
-                        dateValue = new Date(DATE_FORMAT.parse(fragment.asString()).getTime());
+                        dateValue = new Date(dateFormat.parse(fragment.asString()).getTime());
                     } catch (ParseException e) {
                         throw new IllegalArgumentException(e);
                     }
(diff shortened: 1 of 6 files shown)


genuss:
It looks like something wrong with an openjdk6 travis tests.

src/main/java/ru/yandex/clickhouse/response/ByteFragmentUtils.java
```diff
@@ -4,11 +4,23 @@
...
+import java.text.DateFormat;
+import java.text.ParseException;
+import java.text.SimpleDateFormat;
+import java.util.TimeZone;
 
 public final class ByteFragmentUtils {
 
     public static final char ARRAY_ELEMENTS_SEPARATOR = ',';
     public static final char STRING_QUOTATION = '\'';
+    public static final DateFormat DATE_FORMAT = new SimpleDateFormat("yyyy-MM-dd");
```
serebrserg: SimpleDateFormat is not thread safe

src/main/java/ru/yandex/clickhouse/response/ByteFragmentUtils.java
```diff
@@ -181,7 +193,29 @@ public static Object parseArray(ByteFragment value, Class elementClass) {
...
+                    try {
+                        dateValue = new Date(DATE_FORMAT.parse(fragment.asString()).getTime());
+                    } catch (ParseException e) {
+                        throw new IllegalArgumentException(e);
+                    }
+                    java.lang.reflect.Array.set(array, index++, dateValue);
+                } else  if (elementClass == Timestamp.class) {
+                    Timestamp dateTimeValue;
+                    try {
+                        dateTimeValue = new Timestamp(DATE_TIME_FORMAT.parse(fragment.asString()).getTime());
```
serebrserg: fragment here represents datetime string in server timezone, I think it is incorrect to parse it with UTC time zone.

src/main/java/ru/yandex/clickhouse/response/ByteFragmentUtils.java
```diff
@@ -181,7 +193,29 @@ public static Object parseArray(ByteFragment value, Class elementClass) {
...
+                } else if (elementClass == Float.class) {
+                    Float floatValue = Float.parseFloat(fragment.asString());
+                    java.lang.reflect.Array.set(array, index++, floatValue);
+                } else if (elementClass == Double.class) {
+                    Double doubleValue = Double.parseDouble(fragment.asString());
+                    java.lang.reflect.Array.set(array, index++, doubleValue);
+                } else if (elementClass == Date.class) {
+                    Date dateValue;
+                    try {
+                        dateValue = new Date(DATE_FORMAT.parse(fragment.asString()).getTime());
```
serebrserg: fragment here represents date string in server timezone, I think it may be incorrect to parse it with UTC time zone.

src/main/java/ru/yandex/clickhouse/response/ByteFragmentUtils.java
genuss: I suppose I'd better use DateTimeFormatter from joda time then.

src/main/java/ru/yandex/clickhouse/response/ByteFragmentUtils.java
genuss: Is it possible to get server timezone? I can't see how to do that.

serebrserg: Timezone is a property of ClickhouseConnection. It is initialized on ClickhouseConnection initialization either by requesting it from server or from ClickHouseProperties (depending on properties).
The code is here ru.yandex.clickhouse.ClickHouseConnectionImpl#initTimeZone

Timezone is also propagated to ClickHouseResultSet by setting it to local SimpleDateFormat's.
ru.yandex.clickhouse.response.ClickHouseResultSet#initTimeZone

I guess you should somehow pass these DateFormat's to the parsing methods.

src/main/java/ru/yandex/clickhouse/response/ByteFragmentUtils.java
serebrserg: Anyway, you can't set timezone to static fields (because there may be several connections with different timezones), so you probably need to pass the formats from outside. As I mentioned in another comment, we already have SimpleDateFormats as object fields of ClickHouseResultSet with timezones already set, so it must be convenient to pass them to ByteFragmentUtils methods. This solution must be safe, at least it conforms with absence of thread safety in single ResultSet but doesn't allow different ResultSets to interfere.

src/main/java/ru/yandex/clickhouse/response/ByteFragmentUtils.java
genuss: Alright. Will code today :)

genuss:
Just added new commit. Now I use DateFormats from ClickhouseResultSet. The only thing I don't like is I had to create new SimpleDateFormat in ByteFragmentUtilsTest because I couldn't find a way to get the right one to the test. But it's a minor issue I think.

serebrserg:
Noticed that test for date arrays fails on my machine. Reverted, will investigate it separately.

serebrserg:
Will merge it to my branch here https://github.com/yandex/clickhouse-jdbc/pull/114 and investigate the local test failure. It is likely to be linked with timezones settings.
If you continue to work on this feature, please create PRs to branch #108_arrays_support . Meanwhile I will try to reproduce the problem at CI.

genuss:
OK

Issue #117, opened 2017-07-05
Error when using Spark - Syntax error: failed at position 8


sigurdurm:
Spark: 2.1.0 (Hadoop 2.7)
Java/Python

When querying using spark and clickhouse jdbc driver (newest source, head 164d10c), we got this error:
```
Exception in thread "main" org.apache.spark.SparkException: Job aborted due to stage failure: 
Task 0 in stage 0.0 failed 4 times, most recent failure: Lost task 0.3 in stage 0.0 (TID 3, 192.168.1.179, executor 2): 
ru.yandex.clickhouse.except.ClickHouseException: 
ClickHouse exception, code: 62, host: clickhouse, port: 8123; 
Code: 62, e.displayText() = DB::Exception: 
Syntax error: failed at position 8: "name","city","country" FROM default.spark FORMAT TabSeparatedWithNamesAndTypes;, 
expected not empty list of expressions, e.what() = DB::Exception
```

Caused by this line in the code:
https://goo.gl/73N8Ya

Our java spark code looks like this:
```
HashMap<String, String> clickhouseOptions = new HashMap<String, String>() {{
    put("driver", "ru.yandex.clickhouse.ClickHouseDriver");
    put("url", "jdbc:clickhouse://clickhouse:8123");
    put("dbtable", "default.spark");
}};

SparkSession spark = SparkSession.builder().getOrCreate();
Dataset<Row> test = spark.read().format("jdbc").options(clickhouseOptions).load();
test.show();
```

When calling ```show()``` the error happens.

The *default.spark* table has three columns *name, city, country*
The select query looks like this, coming form Spark:
```SELECT "name","city","country" FROM default.spark```

After clickhousifySql(sql), https://goo.gl/1wAuWb:
```SELECT "name","city","country" FROM default.spark FORMAT TabSeparatedWithNamesAndTypes;```

The problem lies in the **quotation marks**, removing them or replacing with a back quote, to support spaces in column names, solves the error:
```SELECT `name`,`city`,`country` FROM default.spark FORMAT TabSeparatedWithNamesAndTypes;```

serebrserg:
This must work with server >= 1.1.54245.
Quote from the linked pull request:
> BTW, we have added support of SQL double quotes in today version of ClickHouse (1.1.54245).

drauschenbach:
I'm still experiencing this with:
* Server 20.1.4.14 with revision 54431 via Docker (Feb 10 2020)
[...]

PR #118, opened 2017-07-05
#117 translate sql double quotes to back quotes

sigurdurm:
Spark adds double quotes to column names when executing a sql query, which is not currently supported in the clickhouse jdbc.

Sql string double quotes get replaced with back quotes, in the clickhousifySql function, e.g.:
```
"a", "a b", "a 1 b-c_d"
``` 
translates to 
```
`a`, `a b`, `a 1 b-c_d`
```


alexey-milovidov:
BTW, we have added support of SQL double quotes in today version of ClickHouse (1.1.54245).

serebrserg:
As I can see there are cases where implementation proposed in this PR fails, for example 
```sql
select 'there is "something" here'
```
Considering support of SQL double quotes had been implemented in server and that it is not that simple to add it in client, I propose to update the server and do nothing at client side.

PR #128, opened 2017-08-16
fix getCatalog/setCatalog

werehuman:
Now `getCatalog` always returns "default" even if specified database is not default. I fixed it and also added working `setCatalog` function.

### DIFF
--- a/src/main/java/ru/yandex/clickhouse/ClickHouseConnectionImpl.java
+++ b/src/main/java/ru/yandex/clickhouse/ClickHouseConnectionImpl.java
@@ -14,6 +14,7 @@ import ru.yandex.clickhouse.util.TypeUtils;
 import ru.yandex.clickhouse.util.guava.StreamUtils;
 
 import java.io.IOException;
+import java.net.URI;
 import java.net.URISyntaxException;
 import java.sql.Array;
 import java.sql.Blob;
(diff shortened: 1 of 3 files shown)


serebrserg:
Thanks for PR and excuse me for the delay.

Here is related issue with for Schemas -> Catalogs transition. https://github.com/yandex/clickhouse-jdbc/issues/94
There are some points that need clarification:
1) Is it allowed to have schemas equal to catalogs? Is using catalogs more convenient than using schemas?
2) Javadocs for both setSchema and setCatalog methods require no affect to already prepared Statement objects. Current implementations of both don't fit that. I think there must be some more copying of properties.

This is not completely clear for me, so discussion is welcome.

werehuman:
1. I don't know about difference between schema and catalog in abstract SQL database, but in MySQL [catalog is a database name](https://github.com/mysql/mysql-connector-j/blob/b3cda4f864902ffdde495b9df93937c3e20009be/src/com/mysql/jdbc/ConnectionImpl.java#L2616-L2631) (as I did here too) and [schemas are not supported too](https://github.com/mysql/mysql-connector-j/blob/b3cda4f864902ffdde495b9df93937c3e20009be/src/com/mysql/jdbc/ConnectionImpl.java#L5431-L5438).
2. This is a good remark, I fixed that. 30586a01a3d16bde27aa1567b0e41f85b00ae7ca

PR #130, opened 2017-08-30
Fix `ClickHousePreparedStatement.setObject(int, List<?>)` for Number

werehuman:
Imagine function:
```java
void mapObject(PreparedStatement statement, List<Supplier<?>> getters) {
    int count = 0;
    for (Supplier<?> getter : getters) {
        statement.setObject(++count, getter.get());
    }
}
```

It works for any types except numeric arrays. Array of double will be serialized as `['1.23','4.56']`, but clickhouse expects no quoting of number (`[1.23,4.56]`).


serebrserg:
Thanks

Issue #132, opened 2017-09-11
ClickHouseStatementImpl.getUpdateCount() doesn't signal no more results


peturrun:
The implementation of ClickHouseStatementImpl.getUpdateCount() will always return 0 if the previous statement didn't return a result set, no matter how often getMoreResults() is called. This causes a loop that tries to read all results from a Statement to loop forever since the following expression from the JavaDoc for Statement is always false:

`((stmt.getMoreResults() == false) && (stmt.getUpdateCount() == -1))`

serebrserg:
Thanks.
Merged your PR. Released in 0.1.28.

peturrun:
Thank you!

peturrun:
By the way, this was discovered when trying to use Flyway (http://flywaydb.org) for schema migration on ClickHouse. This is now working, and is available from here: https://github.com/activitystream/flyway

PR #133, opened 2017-09-11
Fix issue #132

peturrun:
This adds an extra field to hold the return value from getUpdateCount, since it depends not only on whether the last statement was a select, but also needs to change after getMoreResults has been called.

Also getMoreResults will close the current result set (if any).


serebrserg:
Thanks

PR #135, opened 2017-09-12
#123 & #134: handle null string values for bind parameters.

### LINKED ISSUE #123: setString and setObject have different behaviour for null values
When using prepared statements, setString and setObject have different behaviour for null values. The setString method uses ClickHouseUtil.escape(...) which produces the literal string 'NULL', whereas setObject produces \N

### LINKED ISSUE #134: PreparedStatement.execute() fails on null values
PreparedStatement.execute() fails if a null value has been bound to a parameter with setNull:

```
Caused by: java.lang.Throwable: Code: 62, e.displayText() = DB::Exception: Cannot parse expression of type Nullable(String) here: \N), e.what() = DB::Exception

	at ru.yandex.clickhouse.except.ClickHouseExceptionSpecifier.specify(ClickHouseExceptionSpecifier.java:53)
```

PreparedStatement.executeBatch() however works.

Example code:

```
        Connection connection = DriverManager.getConnection("jdbc:clickhouse://clickhouse:8123/default");
        Statement st = connection.createStatement();
        st.executeUpdate("create table if not exists test_null_insert(value Nullable(String)) engine = Memory");

        try {
            PreparedStatement ps = connection.prepareStatement("insert into test_null_insert(value) values(?)");

            // Works
            ps.setNull(1, Types.VARCHAR);
            ps.addBatch();
            ps.executeBatch();

            // Doesn't work
            ps.setNull(1, Types.VARCHAR);
            ps.execute();
        } finally {
            ResultSet rs = st.executeQuery("select value from test_null_insert");
            while (rs.next()) {
                String value = rs.getString(1);
                System.out.println(value != null ? "\"" + value + "\"" : "null");
            }

            st.executeUpdate("drop table test_null_insert");
            connection.close();
        }

```

peturrun:
There are two changes here:

* setString() adds an actual null value (\N) instead of the literal string 'NULL' when given a null parameter (fixes #123);
* A special case for null values is added to appendBoundValue (fixes #134).

### DIFF
--- a/src/main/java/ru/yandex/clickhouse/ClickHousePreparedStatementImpl.java
+++ b/src/main/java/ru/yandex/clickhouse/ClickHousePreparedStatementImpl.java
@@ -108,11 +108,11 @@ public class ClickHousePreparedStatementImpl extends ClickHouseStatementImpl imp
 
     private void appendBoundValue(StringBuilder sb, int num) {
         if (valuesQuote[num]) {
-            sb.append("'");
-        }
-        sb.append(binds[num]);
-        if (valuesQuote[num]) {
-            sb.append("'");
+            sb.append("'").append(binds[num]).append("'");
+        } else if (binds[num].equals("\\N")) {
+            sb.append("null");
+        } else {
+            sb.append(binds[num]);
         }
     }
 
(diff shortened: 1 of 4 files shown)


serebrserg:
Thanks

PR #141, opened 2017-09-26
Add support for boxed primitive types in arrays

Orhideous:
Similar behavior is implemented in the PostgreSQL driver


serebrserg:
Thanks

PR #142, opened 2017-10-05
Add position of missing placeholder in PreparedStatement

alex-krash:
**Goal**
Simplifying debugging process

**Current situation** 
When debugging bind issues, a debugger is needed for determining, what was missed

**Solution**
Added index of missing placeholder for PreparedStatement in exception message


serebrserg:
Thanks

PR #162, opened 2017-12-04
Setting Nullable Metadata based of Column Type.

SleepyThread:
Currently, we return isNull NO for all the columns in Click house table. This pull request checks the column type to be nullable and return the correct value.


serebrserg:
Thanks.

Issue #161, opened 2017-11-24
Implement WITH TOTALS support


IgnatBeresnev:
What I'm suggesting is we add a getter `getWithTotals()` to `ClickHouseResultSet` for those that really need it. This way, `WITH TOTALS` is supported and at the same time is not among the actual ResultSet results, so no unexpected information or behaviour there.

Supposedly, it'll be populated during one of `next()` calls (if you have any better ideas - let me know) and the user will be able to extract it after iterating over ResultSet.

I'll be happy to have a go at this and, if successful, send a PR.

serebrserg:
Hi, this would be nice.
You can also add getExtremes() with same functionality which returns min and max. By the way this can be configured through properties but is turned off by default because it is quite tricky to implement.

Yes, you can fill the values during iteration on main ResultSet. Totals and Extremes are in the end of the stream, so I think it must be stated that they are available only after the main ResultSet is exhausted.
Then one can call getTotals() or getExtremes() and either receive either new ResultSet with totals or extremes, or use the same ResultSet with some internal pointers reset.
So first option would look like
```java
ResultSet totalsRs = rs.getTotals();
if(totalsRs.next()) {
  // receive totals values
}
```
or second:
```java
rs.totals();
if(rs.next()) {
  // receive totals values
}
```
I have no strict preference between these two. 
Let me know if you have different ideas.

IgnatBeresnev:
I think first is better since it's more intuitive and scalable. I'll start working on it tonight.

hamsterready:
Hi guys, any progress on this one? Let me know if there is any active branch with this.

IgnatBeresnev:
@hamsterready still haven't had a chance to even look at it, a lot went down IRL. Hopefully these weekends I'll be able to have a go at this issue 

If you want to start doing it earlier, feel free to, just let us know :)

IgnatBeresnev:
@serebrserg it indeed is tricky to implement. 

The biggest problem arises when it's only 1 column in the result.  For insatnce, if we select 1 column where the result looks like 

column | 
------------ | 
Something | 
Something | 
(empty) | 
Something | 

it'll be sent as `Something\nSomethin\n\nSomething`, which is exactly the same as if we do `SELECT ... WITH TOTALS` on 1 column.

[...]

PR #168, opened 2018-01-10
Fix for #161

hamsterready:
Hi guys, since I need this feature quite badly - here's my implementation :) Please review and let me know. Cheers


serebrserg:
Thanks. Will be in 0.1.35, which should be available tomorrow.

Issue #167, opened 2018-01-08
get colums error


liyongcun:
java.lang.NumberFormatException: For input string: "16)"
	at ru.yandex.clickhouse.response.ByteFragmentUtils.parseInt(ByteFragmentUtils.java:57) ~[clickhouse-jdbc-0.1.34.jar:na]
	at ru.yandex.clickhouse.response.ClickHouseResultSet.getInt(ClickHouseResultSet.java:299) ~[clickhouse-jdbc-0.1.34.jar:na]
	at org.apache.commons.dbcp.DelegatingResultSet.getInt(DelegatingResultSet.java:225) ~[commons-dbcp-1.4.jar:1.4]
	at org.apache.ddlutils.platform.MetaDataColumnDescriptor.readColumn(MetaDataColumnDescriptor.java:123) ~[ddlutils-1.0.jar:1.0]
 this colum type is Nullable(Int16)

serebrserg:
must be fixed in referenced PR

Issue #160, opened 2017-11-23
Problem with ClickHouseDatabaseMetadata


yuragr76:
In method getColumns code
`// decimal digits`
`if (sqlType == Types.INTEGER || sqlType == Types.BIGINT && type.contains("Int")) {`
`                String bits = type.substring(type.indexOf("Int") + "Int".length());`
`                row.add(bits);`
`            } else {`
`                row.add(null);`
`            }`

must return '16)' if column has type Nullable(UInt16)

serebrserg:
Must have been fixed in related PR.

PR #172, opened 2018-02-06
Metadata column fixes

chris-baynes:
Fixes:
1. Decimal digits - this now returns the number of fractional digits (as per the jdbc spec), and also works for nullable types.
2. Column size - the precision of the type, or the width in characters if a Date/DateTime


chris-baynes:
This appears to fix issues: #167 and #160

Issue #184, opened 2018-03-13
Trouble inserting arrays of arrays for nested structures


milktea02:
I'm having trouble inserting an array of array of strings into a nested structure:

```
ru.yandex.clickhouse.except.ClickHouseException: ClickHouse exception, code: 27, host: localhost, port: 32782; Code: 27, e.displayText() = DB::Exception: Cannot parse input: expected [ before: \'[String1, String2]\',\'[String1, String2]\']

name: test.array_of_strings,    type: Array(Array(String)),                                         parsed text: "["ERROR
```

Eg. if my table includes something like:
```
CREATE TABLE test
(
    Period Date, 
    QueryID UInt32, 
    Fingerprint String, 
    nest_structure Nested(
        name String
        array_of_strings Array(String), 
)
ENGINE = MergeTree(Period, QueryID, 8192)
```

```
List<List<String>> arrayOfArrayOfStrings = new ArrayList<>();
List<String> arrayOfStrings = new ArrayList<>();
arrayOfStrings.add("String1");
arrayOfStrings.add("String2");
arrayOfArrayOfStrings.add(arrayOfStrings);
arrayOfArrayOfStrings.add(arrayOfStrings);

...

statement.setObject(5, arrayOfArrayOfStrings);

```

It seems like the ~~first~~ second '[' somehow doesn't get parsed.

It works fine when using the clickhouse-client
```
insert into test values('2017-08-17',5,'abbc',['name1','name2'], [['str1', 'str2'],['str1', 'str2']])
```

using jdbc version 0.1.38
clickhouse-server 1.1.54327

---

Problem is that the arraybuilder is not smart enough to know if it's building something other than an array of strings or an array of ints.

https://github.com/yandex/clickhouse-jdbc/blob/master/src/main/java/ru/yandex/clickhouse/util/ClickHouseArrayUtil.java#L116

https://github.com/yandex/clickhouse-jdbc/blob/master/src/main/java/ru/yandex/clickhouse/util/ClickHouseArrayUtil.java#L120

serebrserg:
released in 0.1.39

PR #185, opened 2018-03-14
#184 - Fix to parse array of arrays for nested structure

milktea02:
Fix for #184


serebrserg:
Thanks.
The code looks good. Could you add a test as well?

milktea02:
@serebrserg will do!

PR #186, opened 2018-03-15
Maven shade plugin for uber jars with shaded dependencies

f1llon:
The `clickhouse-jdbc` strictly depends on such libraries like `com.google.guava` & `com.fasterxml.jackson.core` well known of its issues with backward/forward compatibilities.

I faced the classpath issues when tried to use `clickhouse-jdbc` with Apache Spark 2.3 (depends on Guava v14). Adding the dependency on shaded jar will solve the issue.
Such approach is widely used for the artifacts with dependcies on Guava, Jackson, etc. (see [google](https://search.maven.org/#search%7Cga%7C1%7Ccom.google.cloud.bigdataoss) provides shaded jars for its dependencies)

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -176,6 +176,37 @@
                     </descriptorRefs>
                 </configuration>
             </plugin>
+            <plugin>
+              <groupId>org.apache.maven.plugins</groupId>
+              <artifactId>maven-shade-plugin</artifactId>
+              <version>3.1.0</version>
+              <executions>
+                <execution>
+                  <phase>package</phase>
+                  <goals>
+                    <goal>shade</goal>
+                  </goals>
+                  <configuration>
+                    <minimizeJar>true</minimizeJar>
+                    <relocations>
+                      <relocation>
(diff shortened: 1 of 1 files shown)


tkroman:
as a spark user, i'm really grateful for this

volyx:
Had the same problem, nice PR

serebrserg:
Thanks, I will check how is it built.

f1llon:
@serebrserg, looks like it was a bad idea to shade the `slf4j`. Being a facade it tries to obtain the logger implementations at runtime and, of course, never finds the implementation having `ru.yandex.repackaged.org.slf4j.impl` package. 
My proposition - is to shade only Guava and Jackson, as widely conflicting artifacts, other libraries to be available transitively and excluded if required. 
I could prepare new PR to the yandex:shade_plugin if you are ok.

tkroman:
@f1llon please don't forget joda-time too :)

serebrserg:
@f1llon ok, thanks.

PR #187, opened 2018-03-19
fix minor errors with an encoding

izebit:


serebrserg:
Thanks.
To be consistent I think it is better to use StreamUtils.UTF_8 in all places.

izebit:
@serebrserg  I used the String type, because the test where I made changes was written like this way and I didn't want to make a mess. But if you mentioned it, I will fix it.

Issue #59, opened 2016-12-10
executeBatch implementation


Badya:
@serebrserg @AndreevDm 

Currently i am working on integration between [metabase](https://github.com/metabase/metabase) and [ClickHouse](https://github.com/yandex/clickhouse).

It seems that clickhouse-jdbc driver doesn't provide support for formats like:
`INSERT INTO test.batch_insert (s, i) VALUES (?, ?), (?, ?)`
and
`INSERT INTO test.batch_insert (s, i) VALUES (?, 101), (?, 102)`

Looks like it is supported by ClickHouse (i've wrote some tests to theese formats and fixed 'em), so can it be supported in JDBC?

serebrserg:
Hi.
As far as I understand the problem, it works:
```java
PreparedStatement statement = connection.prepareStatement("INSERT INTO test.batch_insert (s, i) VALUES (?, ?), (?, ?)");
statement.setString(1, "string1");
statement.setInt(2, 21);
statement.setString(3, "string2");
statement.setInt(4, 22);
statement.execute();
```

The second example:
```java
PreparedStatement statement = connection.prepareStatement("INSERT INTO test.batch_insert (s, i) VALUES (?, 22), (?, 23)");
statement.setString(1, "string1");
statement.setString(2, "string2");
statement.execute();
```

If I misunderstood the question, please attach your examples.

Badya:
`PreparedStatement statement = connection.prepareStatement("INSERT INTO test.batch_insert (s, i) VALUES (?, ?), (?, ?)");
statement.setString(1, "str101");
statement.setInt(2, 101);
statement.setString(3, "str102");
statement.setInt(4, 102);
statement.addBatch();
statement.executeBatch();`

That's how JDBC API wraps PreparedStatement execution. (in my case i call it from closure.java.jdbc as jdbc/execute!)
And in this case we got: 
`DB::Exception: Cannot parse input: expected \n before: \t`
Which caused by method:
`ru.yandex.clickhouse.ClickHousePreparedStatementImpl#executeBatch`

I suppose because it throws away all "VALUES (?, ?) .." part of query instead of some standart PreparedStatement to SQLString convertion with parameters binding.

lloonngg:
[...]

PR #190, opened 2018-03-21
added support of batch insert

izebit:
I tried to resolve this issue #59 
I will be glad to see any comments and remarks about my pr.

### DIFF
--- a/src/main/java/ru/yandex/clickhouse/ClickHousePreparedAbstractStatementImpl.java
+++ b/src/main/java/ru/yandex/clickhouse/ClickHousePreparedAbstractStatementImpl.java
@@ -1,0 +1,447 @@
+package ru.yandex.clickhouse;
+
+import org.apache.http.entity.AbstractHttpEntity;
+import org.apache.http.impl.client.CloseableHttpClient;
+import ru.yandex.clickhouse.response.ClickHouseResponse;
+import ru.yandex.clickhouse.settings.ClickHouseProperties;
+import ru.yandex.clickhouse.settings.ClickHouseQueryParam;
+import ru.yandex.clickhouse.util.ClickHouseArrayUtil;
+import ru.yandex.clickhouse.util.guava.StreamUtils;
+
+import java.io.IOException;
+import java.io.InputStream;
+import java.io.OutputStream;
+import java.io.Reader;
+import java.math.BigDecimal;
+import java.math.BigInteger;
+import java.net.URL;
+import java.sql.*;
+import java.sql.Date;
+import java.text.SimpleDateFormat;
+import java.util.*;
+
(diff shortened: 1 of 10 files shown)


src/main/java/ru/yandex/clickhouse/ClickHousePreparedStatementFactory.java
```diff
@@ -0,0 +1,60 @@
...
+
+        String queryWithoutComments = removeCommentsFrom(query);
+        if (queryWithoutComments.regionMatches(true, 0, INSERT, 0, INSERT.length()))
+            return new ClickHousePreparedInsertStatementImpl(httpClient, connection, properties, queryWithoutComments, timeZone);
+        else
+            return new ClickHousePreparedStatementImpl(httpClient, connection, properties, queryWithoutComments, timeZone);
+
+    }
+
+    static String removeCommentsFrom(String str) {
```
serebrserg: Quotation is not handled, and more or less the same logic is there ru.yandex.clickhouse.ClickHousePreparedStatementImpl#parseSql

Also I don't think removing comments is a good idea because it makes it harder to find queries in query_log and trace execution. I think that queries must be modified as little as possible.

src/main/java/ru/yandex/clickhouse/ClickHousePreparedAbstractStatementImpl.java
```diff
@@ -0,0 +1,447 @@
...
+    private void initTimeZone(TimeZone timeZone) {
+        dateTimeFormat.setTimeZone(timeZone);
+        if (properties.isUseServerTimeZoneForDates())
+            dateFormat.setTimeZone(timeZone);
+    }
+
+
+    protected abstract void setParameter(int index, String value, boolean isQuote);
+
+    protected abstract String getSqlQuery() throws SQLException;
```
serebrserg: I think buildSql() says more, because one may misread getSqlQuery as a simple getter while it isn't.
Maybe there is a better name but it must define that it is no simple getter.

src/main/java/ru/yandex/clickhouse/ClickHousePreparedInsertStatementImpl.java
```diff
@@ -0,0 +1,204 @@
...
+            }
+
+            return sb
+                    .toString()
+                    .getBytes(StreamUtils.UTF_8);
+        }
+
+
+        private static List<String[]> getTuples(String sql) {
+            Matcher matcher = ARGUMENT_PATTERN.matcher(sql);
```
serebrserg: Quoted values are not handled here.
For example VALUES ('hello)', ?)

serebrserg:
As I see most of the feature was implemented and merged in https://github.com/yandex/clickhouse-jdbc/pull/246 
If you have something to add, please rebase or create a separate PR based on what is in master. 
I close the PR at the moment. Thanks for your effort.

Issue #164, opened 2017-12-14
ClickHouseResultSetMetaData nullable check always returns false


svetlana-kulygina:
I am trying to use clickhouse with Spark SQL.
Here is nullable field check from `JdbcUtils.getSchema` :

`val nullable = rsmd.isNullable(i + 1) != ResultSetMetaData.columnNoNulls`

But `ClickHouseResultSetMetaData.isNullable` implementation does not really respect column type:
```
    @Override
    public int isNullable(int column) throws SQLException {
        return columnNoNulls;
    }
```
Seems it should be like:
```
String clickhouseType = resultSet.getTypes()[column - 1]
return clickhouseType.startsWith("Nullable(") && clickhouseType.endsWith(")") ? columnNullable : columnNoNulls;
```

serebrserg:
Must be fixed in #266

PR #195, opened 2018-03-27
add implementation for isNullable #164

izebit:
it is a fix for issue #164

### DIFF
--- a/src/main/java/ru/yandex/clickhouse/response/ClickHouseResultSet.java
+++ b/src/main/java/ru/yandex/clickhouse/response/ClickHouseResultSet.java
@@ -5,6 +5,7 @@ import ru.yandex.clickhouse.ClickHouseStatement;
 import ru.yandex.clickhouse.except.ClickHouseExceptionSpecifier;
 import ru.yandex.clickhouse.settings.ClickHouseProperties;
 import ru.yandex.clickhouse.util.TypeUtils;
+import ru.yandex.clickhouse.util.Utils;
 
 import java.io.IOException;
 import java.io.InputStream;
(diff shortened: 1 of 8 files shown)


src/main/java/ru/yandex/clickhouse/response/ClickHouseResultSet.java
```diff
@@ -592,32 +646,113 @@ public BigDecimal getBigDecimal(int columnIndex, int scale)  {
...
+        }
+        if (index == start)
+            return null;
+        else
+            return query.substring(start, index);
     }
 
 
+    private static boolean isWithTotals(String sql) {
+        return sql.startsWith("select") && sql.contains(" with totals ");
```
serebrserg: Quotations and comments are not handled

src/main/java/ru/yandex/clickhouse/response/ClickHouseResultSet.java
```diff
@@ -592,32 +646,113 @@ public BigDecimal getBigDecimal(int columnIndex, int scale)  {
...
     }
 
     @Override
     public void setFetchSize(int rows) throws SQLException {
-        // ignore perfomance hint
+        // ignore performance hint
+    }
+
+    static Map<String, String> getTableNames(String query, String[] columnNames) {
+        Map<String, String> tableNames = new HashMap<String, String>(columnNames.length);
```
serebrserg: There also must be a more robust query parser handling comments and quoted strings.
I guess there must be one for the driver that parses whole query to tokens(maybe even generated with some parser generator).

serebrserg: As I see it is a separate feature from isNullable, so it is better to extract it to a separate PR if possible.

serebrserg:
In meanwhile the original issue was fixed in #266 with smaller footprint.
If you have things to add, please rebase and propose your changes.
I close the PR at the moment.

PR #203, opened 2018-04-13
Returning true value of IsNullable from ClickHouseResultSetMetaData

SleepyThread:
ResultSetMetaData::isNullable method should return nullability of values in the designated column.
Currently, Clickhouse returns columnNoNulls for all columns including Nullable. This patch fixes the issue.


SleepyThread:
@serebrserg Any feedback on this pull request?

SleepyThread:
@serebrserg  any updates?

PR #214, opened 2018-06-01
Add ResultSet isLast() implementation

devoxi:


src/main/java/ru/yandex/clickhouse/response/ClickHouseResultSet.java
```diff
@@ -580,6 +580,15 @@ public BigDecimal getBigDecimal(String columnLabel, int scale)  {
         return getBigDecimal(asColNum(columnLabel), scale);
     }
 
+    @Override
+    public boolean isLast() throws SQLException {
+        if (getType() == TYPE_FORWARD_ONLY) {
+            return !hasNext();
+        } else {
+            throw new UnsupportedOperationException();
```
AndreevDm: ResultSet Interface recommends throw SQLFeatureNotSupportedException, but type is always TYPE_FORWARD_ONLY so you can skip the check
https://github.com/ContentSquare/clickhouse-jdbc/blob/5f5ab8cf01a6464c0e91e979a90af26aa31dfe6a/src/main/java/ru/yandex/clickhouse/response/ClickHouseResultSet.java#L515

devoxi: If you want me to integrate the test inside the others instead of having a separate test, let me know.

PR #238, opened 2018-09-29
Implementation `ClickHousePreparedStatementImpl#getMetaData()`

tac0x2a:
Hi, I implement `ClickHousePreparedStatementImpl#getMetaData()`.

This method needs to execute query if before execution query for getting ResultSet via `getResultSet()`.


serebrserg:
The option with limit was implemented in #288 . Thanks for your effort.

PR #241, opened 2018-10-09
adding support for writing UUIDs in the binary stream

tsl-karlp:


tkroman:
Hi. Would be nice to have this, but is the 1.7 bump necessary? Looks like it broke the build plus doesn't look like having UUIDs depends on this bump. Maybe you could split this one into 2 PRs so that it's easier to get UUID pars merged?

tsl-karlp:
@tkroman : Sorry, looks like the upgrade to Java 1.7 was unnecessary.  I also fixed the test.

PR #242, opened 2018-10-14
bump dependency versions

tkroman:
and fix a couple of warnings

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -43,66 +43,74 @@
 
     <properties>
         <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
+        <jackson.version>2.9.7</jackson.version>
+        <httpclient.version>4.5.6</httpclient.version>
+        <lz4.version>1.5.0</lz4.version>
+        <testng.version>6.14.3</testng.version>
+        <guava.version>26.0-jre</guava.version>
+        <slf4j-api.version>1.7.25</slf4j-api.version>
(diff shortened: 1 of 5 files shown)


pom.xml
```diff
@@ -43,66 +43,74 @@
...
             <artifactId>httpmime</artifactId>
-            <version>4.5.2</version>
+            <version>${httpclient.version}</version>
         </dependency>
         <dependency>
-            <groupId>net.jpountz.lz4</groupId>
-            <artifactId>lz4</artifactId>
-            <version>1.3.0</version>
+            <groupId>org.lz4</groupId>
+            <artifactId>lz4-java</artifactId>
```
tkroman: same artifact, moved to new coordinates in 1.4.0 https://github.com/lz4/lz4-java/blob/master/CHANGES.md#140

src/main/java/ru/yandex/clickhouse/settings/ClickHouseProperties.java
```diff
@@ -203,6 +203,8 @@ public ClickHouseProperties(ClickHouseProperties properties) {
         setSessionCheck(properties.sessionCheck);
         setSessionId(properties.sessionId);
         setSessionTimeout(properties.sessionTimeout);
+        setPreferredBlockSizeBytes(properties.preferredBlockSizeBytes);
```
tkroman: these were not copied from original properties instance, i assume it's an omission, not intentional

PR #244, opened 2018-10-17
Allow to use dash in the database name

ei-grad:


ei-grad:
Closed in flavor of #247

PR #252, opened 2018-10-25
no joda dependency

ivanrudniev:
Joda-time library is used only for calculation of daysSinceEpoch variable if one class. Removing this dependency gives less clutter for java 8 applications that use new date-time classes.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -97,6 +97,7 @@
             <groupId>joda-time</groupId>
             <artifactId>joda-time</artifactId>
             <version>2.9.9</version>
+            <scope>test</scope>
         </dependency>
 
         <dependency>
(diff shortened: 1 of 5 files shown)


enqueue:
I applaud the effort, there is no need to include this library. Is there a way to adjust the unit tests, as well? Please also double-check the indentation in your code.

ivanrudniev:
Since method void writeDate(LocalDate date) is deleted, there should be some migration note added. Users of this method need to:
manually calculate daysSinceEpoch and use method writeUInt16():
- for joda-time: Days.daysBetween(LocalDate.parse("1970-01-01"), date).getDays() 
- for java 8: LocalDate.toEpochDay())
or use conversion to Date:
- for joda-time: localDate.toDate()
- for java 8: java.sql.Date.valueOf(LocalDate date)

serebrserg:
Thanks. I will add it to changelog on release.

PR #267, opened 2018-11-21
Add property for setting HTTP authorization header.

morimekta:
This very simple workaround for handling auth protected ingress / proxy
before a clickhouse instance. An even better solution would be to be
able to inject the Apache TargetAuthenticationStrategy, but since the
properties system is this simple I'm hesitant to add such logic for
handling something unrelated to clickhouse itself.

- Setting setHttpAuthorization() in the ClickhouseProperties will use
  the value to set the HTTP "Authorization" header value. E.g.:

```java
properties.setHttpAuthorization("Basic " + Base64.encodeToString("username:password"));
ClickHouseDataSource dataSource = new ClickHouseDataSource(url, properties);
dbi = new DBI(dataSource);
try (Handle handle = dbi.open()) {
  // do stuff.
}
```

Has been tested with HTTP ingress with enforced basic auth.

### DIFF
--- a/src/main/java/ru/yandex/clickhouse/settings/ClickHouseProperties.java
+++ b/src/main/java/ru/yandex/clickhouse/settings/ClickHouseProperties.java
@@ -37,10 +37,10 @@ public class ClickHouseProperties {
 
     // queries settings
     private Integer maxParallelReplicas;
-    private String totalsMode;
-    private String quotaKey;
+    private String  totalsMode;
+    private String  quotaKey;
     private Integer priority;
-    private String database;
+    private String  database;
     private boolean compress;
     private boolean decompress;
     private boolean extremes;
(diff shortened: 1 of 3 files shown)


pom.xml
```diff
@@ -115,6 +115,23 @@
 
     <build>
         <plugins>
+            <plugin>
+                <groupId>org.apache.maven.plugins</groupId>
```
serebrserg: I think this is not connected to the PR main goal.
I'm not sure this is needed and that it requires nothing extra, so it is better to separate it to another PR with specific description.

morimekta: Moved to separate PR

morimekta:
Note that I have not added tests for this (yet), as I have not yet managed to make the tests run locally, and this is handling systems (possibly) outside of clickhouse itself. If you require tests for this feature, I will make the effort to do so.

serebrserg:
I think this feature is quite specific and quite simple, that having a proxy set up in test environment is too much effort for it.
Thanks.

PR #273, opened 2018-11-27
Add writeNullable to RowBinaryStream

kanterov:
Following up on discussion in https://github.com/apache/beam/pull/7006#discussion_r232695315

@serebrserg can you please take a look?

### DIFF
--- a/src/main/java/ru/yandex/clickhouse/util/ClickHouseRowBinaryStream.java
+++ b/src/main/java/ru/yandex/clickhouse/util/ClickHouseRowBinaryStream.java
@@ -274,8 +274,23 @@ public class ClickHouseRowBinaryStream {
         }
     }
 
-    public void writeNullable(boolean value) throws IOException {
-        writeByte(value ? (byte) 1 : (byte) 0);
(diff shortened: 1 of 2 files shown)


src/main/java/ru/yandex/clickhouse/util/ClickHouseRowBinaryStream.java
```diff
@@ -274,6 +274,10 @@ public void writeFloat64Array(double[] doubles) throws IOException {
         }
     }
 
+    public void writeNullable(boolean value) throws IOException {
```
serebrserg: At first I expected here to be writeNullable methods for each primitive input type with wrappers instead of primitives. I think this would be the clearest option from client point of view.
However this option is ok as well, but I think this should be called markNextNullable(boolean isNullable), because it doesn't actually write nullable value, but marks that next is nullable.

kanterov: @serebrserg agree, I added javadoc with an example, and renamed method

kanterov:
@serebrserg thanks for the review

PR #275, opened 2018-12-03
Support for Scrollable ResultSet

edonin:
Hi,

For a project, I need a scrollable ResultSet for clickhouse (beeing able to call previous(), first(), last()... methods). I change the ResultSet for making it always scrollable. I do not know if there is a need for still having a "forward only" ResultSet.

Feel free to discuss with me about this change.

Best Regards

### DIFF
--- a/src/main/java/ru/yandex/clickhouse/ClickHouseConnectionImpl.java
+++ b/src/main/java/ru/yandex/clickhouse/ClickHouseConnectionImpl.java
@@ -149,8 +149,8 @@ public class ClickHouseConnectionImpl implements ClickHouseConnection {
     @Override
     public ClickHouseStatement createStatement(int resultSetType, int resultSetConcurrency,
                                                int resultSetHoldability) throws SQLException {
-        if (resultSetType != ResultSet.TYPE_FORWARD_ONLY && resultSetConcurrency != ResultSet.CONCUR_READ_ONLY
-            && resultSetHoldability != ResultSet.CLOSE_CURSORS_AT_COMMIT) {
+        if (resultSetType == ResultSet.TYPE_SCROLL_SENSITIVE || resultSetConcurrency != ResultSet.CONCUR_READ_ONLY
+            || resultSetHoldability != ResultSet.CLOSE_CURSORS_AT_COMMIT) {
             throw new SQLFeatureNotSupportedException();
         }
         return createStatement();
--- a/src/main/java/ru/yandex/clickhouse/response/ClickHouseResultSet.java
+++ b/src/main/java/ru/yandex/clickhouse/response/ClickHouseResultSet.java
@@ -36,6 +36,8 @@ public class ClickHouseResultSet extends AbstractResultSet {
     private final String[] types;
 
     private int maxRows;
+    
+    private List<ByteFragment[]> lines;
 
     // current line
     private ByteFragment[] values;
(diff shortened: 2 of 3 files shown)


serebrserg:
Hi,
There is a need for a forward only ResultSet when the whole result can't fit in memory and need to be streamed right to some other sink.
I think there should be a setting to enable scrollable ResultSet, leaving forward only as default. 
I think it is cleaner not to give considerable memory footprint in driver code, and if you need a scrollable one, in most cases the easier option is to map forward only result set into a list of business objects and scroll over them, having explicitly allocated memory for it, not hidden in the driver.

edonin:
Hi Serguey,

I understand the use case. Is it acceptable to have by default a "forward only" result set, and, if during the creation of the statement, you specify a "scrollable type" result set, you will have a scrollable one ?

it will requiere a bit more changes, but it is still acceptable.

serebrserg:
Yes, it is.

edonin:
It needed more changes, but it should be better now.

src/main/java/ru/yandex/clickhouse/response/ClickHouseScrollableResultSet.java
```diff
@@ -0,0 +1,139 @@
...
+		} else {
+			int current = rowNumber;
+			// We have to check the number of total rows
+			while(hasNext()) {
+				next();
+			}
+			if(-row > lines.size()) {
+				// there is not so many rows
+				// Put back the cursor where it was.
+				absolute(current);
```
serebrserg: according to javadoc, cursor must be before the first row in this case.

```
 * <p>An attempt to position the cursor beyond the first/last row in
 * the result set leaves the cursor before the first row or after
 * the last row.
```

src/test/java/ru/yandex/clickhouse/response/ClickHouseScrollableResultSetTest.java
```diff
@@ -0,0 +1,424 @@
...
+    public Object[][] longArrays() {
+        return new Object[][]{
+                {"[0]", new long[]{0}},
+                {"[324000111222,123,-456]", new long[]{324000111222L, 123, -456}},
+                {"[]", new long[]{}},
+        };
+    }
+
+    @Test(dataProvider = "longArrays")
+    public void toLongArrayTest(String str, long[] expected) throws Exception {
```
serebrserg: I'm not sure the copied part must be exactly copied. Is it possible to do some parameterized test for common scenarios?

serebrserg:
Looks good. Have some comments/questions.

src/main/java/ru/yandex/clickhouse/response/ClickHouseScrollableResultSet.java
edonin: Thanks for the review,. I am correcting that, and I found similar issues with at the limit values.

src/test/java/ru/yandex/clickhouse/response/ClickHouseScrollableResultSetTest.java
edonin: Hi, I am not sure to understand you here. Do you want me to change the hardcoded response/values between the two test classes? Or do you want me to have shared tests between the two classes where the ResultSet instanciation should be done depending of the class were the test is called? 

Thanks.

serebrserg: The latter one - to have some shared test for both ClickhouseResultSet and ClickHouseScrollableResultSet, and a special one for ClickHouseScrollableResultSet covering its specific features.

edonin: The common tests are now at only one place.

edonin:
Logically, all of the request changes are done.

PR #277, opened 2018-12-05
Metadata minor issues

enqueue:
* JDBC URL may contain trailing slash
* Use driver version from MANIFEST instead of constant "0.1"
* Memory is a valid TABLE engine
* Decimal type should be treated as such

### DIFF
--- a/src/main/java/ru/yandex/clickhouse/ClickHouseDatabaseMetadata.java
+++ b/src/main/java/ru/yandex/clickhouse/ClickHouseDatabaseMetadata.java
@@ -11,13 +11,12 @@ import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Iterator;
 import java.util.List;
-import java.util.regex.Matcher;
-import java.util.regex.Pattern;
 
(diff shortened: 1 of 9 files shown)


src/main/java/ru/yandex/clickhouse/ClickHouseDatabaseMetadata.java
```diff
@@ -97,17 +101,34 @@ public String getDriverName() throws SQLException {
...
+            return 0;
+        }
+        Matcher m = VERSION_NUMBER_PATTERN.matcher(v);
+        return m.matches() ? Integer.parseInt(m.group(1)) : 0;
     }
 
     @Override
     public int getDriverMinorVersion() {
-        return 1;
+        String v;
```
serebrserg: I suppose the common part from these methods may be extracted to something returning either array or version object to make it clearer and avoid duplications

enqueue: good suggestion!

PR #280, opened 2018-12-10
Support for queries with RowBinary result

bgranvea:
I've added ClickHouseStatement.executeQueryClickhouseRowBinaryStream methods which return a ClickHouseRowBinaryInputStream to parse result. I've reused integration test in RowBinaryStreamTest.

A quick benchmark shows a performance gain lower than what I expected, maybe because of the little endian to big endian conversion. It also varies depending if the client code runs on the same machine as Clickhouse server.

### DIFF
--- a/src/main/java/ru/yandex/clickhouse/util/ClickHouseRowBinaryInputStream.java
+++ b/src/main/java/ru/yandex/clickhouse/util/ClickHouseRowBinaryInputStream.java
@@ -99,7 +99,9 @@ public class ClickHouseRowBinaryInputStream implements Closeable {
 	}
 
 	/**
-	 * Warning: the result is negative in Java if UInt8 > 0x7f
+	 * Warning: the result is negative in Java if UInt8 &gt; 0x7f
+	 * @return next UInt8 value as a byte
+	 * @throws IOException
 	 */
 	public byte readUInt8AsByte() throws IOException {
 		return in.readByte();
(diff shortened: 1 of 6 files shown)


src/main/java/ru/yandex/clickhouse/util/ClickHouseRowBinaryInputStream.java
```diff
@@ -0,0 +1,375 @@
...
+		int value = readUInt8();
+		validateInt(value, 0, 1, "boolean");
+		return value != 0;
+	}
+
+	public short readUInt8() throws IOException {
+		return (short) in.readUnsignedByte();
+	}
+
+	/**
```
serebrserg: I see the build had failed because of these javadoc-like comments. Documentation plugin expected javadoc to have specific fields. These should be simple comments or must be written with regards to javadoc rules.

Here is the build log with errors
https://api.travis-ci.org/v3/job/466062930/log.txt

src/main/java/ru/yandex/clickhouse/util/ClickHouseRowBinaryInputStream.java
```diff
@@ -0,0 +1,383 @@
...
+			offset += read;
+			length -= read;
+		}
+	}
+
+	public int readByte() throws IOException {
+		return in.read();
+	}
+
+	public boolean isNull() throws IOException {
```
serebrserg: Maybe readIsNull() will be more clear (to mark that we actually read data and move forward).

PR #282, opened 2018-12-11
Enable to get a bound statement in ClickHousePreparedStatement

h6ak:
I have implemented `ClickHousePreparedStatement.toString()`.
This method returns bound statement when `buildSql()` succeeds, otherwise unbound statement.

### DIFF
--- a/src/main/java/ru/yandex/clickhouse/ClickHousePreparedStatement.java
+++ b/src/main/java/ru/yandex/clickhouse/ClickHousePreparedStatement.java
@@ -25,4 +25,6 @@ public interface ClickHousePreparedStatement extends PreparedStatement, ClickHou
     ResultSet executeQuery(Map<ClickHouseQueryParam, String> additionalDBParams, List<ClickHouseExternalData> externalData) throws SQLException;
 
     int[] executeBatch(Map<ClickHouseQueryParam, String> additionalDBParams) throws SQLException;
+
+    String toString();
 }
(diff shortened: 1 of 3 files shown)


src/main/java/ru/yandex/clickhouse/ClickHousePreparedStatement.java
```diff
@@ -25,4 +25,6 @@
     ResultSet executeQuery(Map<ClickHouseQueryParam, String> additionalDBParams, List<ClickHouseExternalData> externalData) throws SQLException;
 
     int[] executeBatch(Map<ClickHouseQueryParam, String> additionalDBParams) throws SQLException;
+
+    String toString();
```
enqueue: This is not required in Java; the class can simply override Object#toString().

enqueue:
@h6ak  what do you want t to achieve? Is this just for debugging purposes? `toString` is a very special method in Java, as you perhaps know.

h6ak:
@enqueue 
Thank you for your comment.

> what do you want t to achieve? Is this just for debugging purposes?

Yes, the goal is to enable us to check bound statements in our program for debugging.

Should I rename the implemented method ?

enqueue:
@h6ak I think this would be useful, but `toString` is called in places you do not expect / cannot control, and it might be expensive when you have statements with thousands of parameter values. I would suggest to use a different method name. If you still want to use `toString`, you can remove the method declaration from the interface `ClickHousePreparedStatement` because you will just override `Object#toString`.

h6ak:
@enqueue I understand what you say. I've renamed the implemented method to `asSql`.

PR #285, opened 2018-12-28
Metabase fixes

serebrserg:


Slach:
@ei-grad and @serebrserg guys! thanks for your work
will Metabase work with latest version clickhouse-jdbc drivers?

i just reference this pull request on original feature requests
https://github.com/metabase/metabase/issues/3332

PR #308, opened 2019-02-23
Implementation of cancel method

### LINKED ISSUE #305: statement cancel method has not been implemented.
I have a scenario where I can cancel the query at any time ,such as I write a wrong sql on terminal, I can press CTRL+C

izebit:
fix #305

### DIFF
--- a/src/main/java/ru/yandex/clickhouse/ClickHouseStatementImpl.java
+++ b/src/main/java/ru/yandex/clickhouse/ClickHouseStatementImpl.java
@@ -77,7 +77,7 @@ public class ClickHouseStatementImpl implements ClickHouseStatement {
 
     private final boolean isResultSetScrollable;
 
-    private final String queryId;
+    private volatile String queryId;
 
     /**
      * Current database name may be changed by {@link java.sql.Connection#setCatalog(String)}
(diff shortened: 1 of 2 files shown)


src/main/java/ru/yandex/clickhouse/ClickHouseStatementImpl.java
```diff
@@ -828,13 +842,22 @@ public void closeOnCompletion() throws SQLException {
...
     	if(isResultSetScrollable) {
     		return new ClickHouseScrollableResultSet(is, bufferSize, db, table, usesWithTotals, statement, timezone, properties);
     	} else {
     		return new ClickHouseResultSet(is, bufferSize, db, table, usesWithTotals, statement, timezone, properties);
     	}
     }
+
+    private Map<ClickHouseQueryParam, String> addQueryIdTo(Map<ClickHouseQueryParam, String> parameters) {
+        if (parameters.containsKey(ClickHouseQueryParam.QUERY_ID))
+            return parameters;
```
serebrserg: In this case queryId field should be set with value form parameters to be able to kill a query with user-defined QUERY_ID

izebit: I have added an implementation for the case when a user set `QUERY_ID` parameter.

Issue #307, opened 2019-02-23
Is there any way to provide IN clause?


doom369:
Didn't find any way to put a list for IN clause.

`where id in (?)`

Driver always accepts list as array and result is always `where id in ([1,2,3])`

doom369:
Fixed in https://github.com/blynkkk/clickhouse4j

PR #309, opened 2019-02-23
#307 provide a way to build values for IN clause

doom369:
Currently clickhouse jdbc driver doesn't provide any way to specify dynamic number of values for IN clause. For example, 

```
List<Integer> list = Arrays.asList(1,2,3,4,5);
statement.setObject(1, list);
```
for `where id in (?)` will always build an array:

`WHERE id IN ([1,2,3,4,5])`

while result should be:

`WHERE id IN (1,2,3,4,5)`

Fixes https://github.com/yandex/clickhouse-jdbc/issues/307.

This PR breaks back compatibility for those who uses `setObject` method with list. They need to use array instead of the list in order to preserve previous behaviour.

### DIFF
--- a/src/main/java/ru/yandex/clickhouse/util/ClickHouseArrayUtil.java
+++ b/src/main/java/ru/yandex/clickhouse/util/ClickHouseArrayUtil.java
@@ -134,18 +134,17 @@ public class ClickHouseArrayUtil {
         return builder.build();
     }
 
-    public static String toString(Collection collection) {
-        return toString(collection, true);
-    }
-
     /**
      * Convert collection to its ClickHouse-string representation.
      *
      * @param collection the collection to transform
-     * @param escape     enable or disable escaping of the collection elements
      */
-    public static String toString(Collection collection, boolean escape) {
-        return toString(collection.toArray(), escape);
+    public static String toString(Collection collection) {
+        ArrayBuilder builder = new ArrayBuilder(needQuote(collection.toArray()), false, false);
+        for (Object value : collection) {
+            builder.append(value);
+        }
+        return builder.build();
     }
 
     private static boolean needQuote(Object[] objects) {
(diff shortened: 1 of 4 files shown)


Au6ojlut:
Well, there is one way to do so. You can convert list of objects for `in` clause to string like `1,2,3,4,5`, then convert it to `byte[]` and `setObject(index, byte[])`.


Yeah, it looks awful but it works...

doom369:
Yes, there are workarounds, however current clickhouse flow is not the same as in other drivers like postgres, so merging this PR will bring the same (and expected) flow as in other drivers. So migration from other solutions would be even simpler.

alex-krash:
Binding multiple values for IN() section is a common problem for any JDBC driver.
We are not going to merge this PR, moreover it brakes backward compatibility.

doom369:
@alex-krash no problem.

For those who are interested in this and some other fixes/improvements we are gonna release own fork soon of this jdbc driver here https://github.com/blynkkk/clickhouse4j.

We already removed guava, jackson dependencies and working on making pluggable http client in order to make driver almost dependency free.

himgang:
> Well, there is one way to do so. You can convert list of objects for `in` clause to string like `1,2,3,4,5`, then convert it to `byte[]` and `setObject(index, byte[])`.
> 
> Yeah, it looks awful but it works...

query  = "select * from table where platform IN (?) limit 10;"

StringBuilder sb = new StringBuilder();
		sb.append("1").append(",").append("2");
statement.setObject(3, sb.toString().getBytes());
@Au6ojlut 
I am unable to get this to work for IN clause.
getting exception : 
Caused by: java.lang.Throwable: Code: 62, e.displayText() = DB::Exception: Syntax error: failed at position 189: \x28\x31\x2C\x32\x29  group by url, domain ,groupedSections limit 1 FORMAT TabSeparatedWithNamesAndTypes;. Unrecognized token (version 19.17.6.36 (official build))


Please Suggest alternative or correct format, If I am making a mistake.

Issue #312, opened 2019-02-26
No way to execute FORMAT CSVWithNames query


doom369:

zhicwu:
`select 1,2 format JSONEachRow` works too. However, this is currently limited to http implementation.

PR #313, opened 2019-02-26
#312 Allow CSV format in select clause

doom369:
This PR partially fixes https://github.com/yandex/clickhouse-jdbc/issues/312 and prepares the code for further refactoring/improvements (make clickhouse result set aware of `Format` type).

This change is required in order to allow in format like `CSV` that will allow (after further fixes) to export full csv from select clause.


doom369:
Will be implemented soon in alternative fork - https://github.com/blynkkk/clickhouse4j

PR #327, opened 2019-04-08
Make sure timezone is initialized

kornilova-l:
Timezone is not initialised if none of `use_server_time_zone` and `use_time_zone` is set.
As consequence client gets parser exception when retrieving dates. See https://youtrack.jetbrains.com/issue/DBE-7776


serebrserg:
Thanks

PR #330, opened 2019-04-12
Improve Detection of DateTime Columns

### LINKED ISSUE #329: Wrong column type for DateTime with explicit timezone
The driver does not understand `DateTime` columns with explicit timezone (see yandex/ClickHouse#1456). 

Test like this:

```
SELECT toDateTime(myDateTimeColumn) FROM myTable
```

Result looks like this on the wire:

```
toDateTime(myDateTimeColumn)
DateTime(\'UTC\')
2019-02-08 12:45:00
```

JDBC driver does not recognize this as a `DateTime` column.

For now we should treat these columns just like any other `DateTime` column.

enqueue:
Support DateTime columns with explicit timezones. Fix #329

### DIFF
--- a/src/main/java/ru/yandex/clickhouse/util/TypeUtils.java
+++ b/src/main/java/ru/yandex/clickhouse/util/TypeUtils.java
@@ -177,7 +177,7 @@ public class TypeUtils {
         } else if (type.equals("Date")) {
             // number of chars in '2018-01-01'
             return 10;
-        } else if (type.equals("DateTime")) {
+        } else if (type.startsWith("DateTime")) {
             // number of chars in '2018-01-01 01:02:35'
             return 19;
         } else if (type.startsWith("FixedString(")) {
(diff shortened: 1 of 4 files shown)


serebrserg:
Thanks for PR, it looks good.
Questions I don't have answer for yet:
Do we need parse DateTime values using specific timezone from column type? May timezone in column type differ from server timezone?

enqueue:
I agree that we might have an issue with parsing `DateTime` values in different time zones. I first wanted to fix the detection of the _type_ of the column.

Yes, it is possible to have `DateTime` with a different time zone than server time zone, e.g.:

```
SELECT timezone()

┌─timezone()─┐
│ UTC        │
└────────────┘
```

```
SELECT toDateTime('2019-04-19 13:37:00', 'Asia/Yekaterinburg')
FORMAT TabSeparatedWithNamesAndTypes

toDateTime(\'2019-04-19 13:37:00\', \'Asia/Yekaterinburg\')
DateTime(\'Asia/Yekaterinburg\')
2019-04-19 13:37:00
```
In that case, the JDBC driver will parse using a wrongly initialized date format :cry:

serebrserg:
OK, this doesn't brake anything, but enables us to do future work on supporting zoned DateTimes. Thanks.

Issue #333, opened 2019-04-17
ERROR: text "<SINGLE QUOTE>2016-08-1" is not like DateTime


simPod:
This is the error I see in clickhouse when trying to batch insert

```
2019.04.17 00:08:45.729957 [ 264 ] {0e04b028-f581-4c85-a668-1b3c1d744863} <Error> HTTPHandler: Code: 27, e.displayText() = DB::Exception: Cannot parse input: expected \t before: \'2016-08-12 13:21:32\'\t32767\t\'163.172.36.163\'\t\'185.229.191.169\'\t1522\t1\t0\t0\t0\t1000003\t1000010\t8080\t12724\t16\t\'128110002367193\'\t\'116530121835\'\t968\t968\t968\t\'\'\t\'Amste: (at row 1)

Row 1:
Column 0,   name: received_at,     type: DateTime, ERROR: text "<SINGLE QUOTE>2016-08-1" is not like DateTime
```

I'm using the same code as is in test here https://github.com/yandex/clickhouse-jdbc/blob/d75f1594b9eebcab200b27bc426b136233b1b927/src/test/java/ru/yandex/clickhouse/integration/BatchInsertsTest.java#L99 Even the same value.

[Thus I blame the newer ClickHouse version I'm running (19.4.3.11)](https://github.com/yandex/clickhouse-jdbc/issues/332).

simPod:
Also this doesn't work:

`statement.setString(1, format.format(new Date(long * 1000L)));` same error

I changed the DateTime to UInt32 and all inserts pass through without problem `statement.setLong(1, long);`

simPod:
I just noticed it inserts string values with quotes. Like `'163.172.83.181'` instead of `163.172.83.181`. That will be the problem.

simPod:
And this works:

```java
SimpleDateFormat format = new SimpleDateFormat("yyyy-MM-dd hh:mm:ss");
statement.setBytes(1, format.format(new Date(long * 1000L)).getBytes());
```

So the problem is in escaping I guess https://github.com/yandex/clickhouse-jdbc/blame/149967fd91d3d1191254188603951df09c6061bb/src/main/java/ru/yandex/clickhouse/ClickHousePreparedStatementImpl.java#L208 Can't help further tho, don't really undestand what happens there.

Baoqi:
I also encountered this issue today, as it turns out, when you use prepared statement to batch insert values, if you see

```
[...]

PR #339, opened 2019-04-26
Fix multi-line statement parsing for prepared statements

enqueue:
Fix for #333


enqueue:
Please also review #317

PR #347, opened 2019-05-21
Add max_insert_block_size, insert_deduplicate, insert_distributed_sync

kanterov:
Otherwise, JDBC properties aren't propagated.


kanterov:
@serebrserg please take a look

Issue #283, opened 2018-12-14
Inserting Array(FixedString(N))


hamsterready:
Hi,

It looks to me that it is not possible to use the JDBC driver to insert data into Array(FixedString(N)) column.

[The code](https://github.com/yandex/clickhouse-jdbc/blob/4697970bbffc4729889e9ba6436432c039920bed/src/main/java/ru/yandex/clickhouse/util/ClickHouseArrayUtil.java#L118) looks like this:

```
        ArrayBuilder builder = new ArrayBuilder(needQuote(values));
        for (Object value : values) {
            builder.append(value);
        }
```

And needQuote will always return true.

Here's full example including notes from [this post](https://groups.google.com/forum/#!topic/clickhouse/V_sRCCQdvHM).

```
import java.nio.ByteBuffer
import java.sql.Types
import java.util.UUID

import com.google.common.io.BaseEncoding
import ru.yandex.clickhouse.ClickHouseArray
import ru.yandex.clickhouse.util.ClickHouseArrayUtil
import ru.yandex.clickhouse.util.guava.StreamUtils


def toBytes(uuid: UUID): Array[Byte] = {
  val bts = ByteBuffer.
    allocate(16).
    putLong(uuid.getMostSignificantBits).
    putLong(uuid.getLeastSignificantBits).
    array
  val hex = "\\x" + BaseEncoding.base16.withSeparator("\\x", 2).encode(bts)
  hex.getBytes
}


val u1 = UUID.randomUUID()
val u2 = UUID.randomUUID()
val arr1 = List(u1, u2).map(toBytes).toArray
val arr2 = List(u1, u2).map(toBytes).map(new String(_, StreamUtils.UTF_8)).toArray

ClickHouseArrayUtil.arrayToString(new ClickHouseArray(Types.BINARY, arr1).getArray)

ClickHouseArrayUtil.arrayToString(new ClickHouseArray(Types.BINARY, arr2).getArray)

ClickHouseArrayUtil.arrayToString(arr2)

```

And the output:

```

toBytes: (uuid: java.util.UUID)Array[Byte]

u1: java.util.UUID = efbca4dd-cb17-48d8-92e3-26655d029a9a
u2: java.util.UUID = a776fc26-bf69-4dcb-b9f5-fef6e16836c3
[...]

Issue #153, opened 2017-11-02
Problem with Array(DateTime) in PreparedStatement


TanVD:
In current implementation "Timestamp.toString()" is used in setValue for arrays. It leads to inability to  insert Array(DateTime).
And this code will not work:
``` statement.setArray(index, statement.connection.createArrayOf("Array(DateTime)", array)) ```
with following exception (with example DateTime)
```  String is too long for DateTime: 2001-01-01 11:11:11.0 ```

serebrserg:
Thanks for report.
In fact arrays are currently serialized different from single values of corresponding types set with setObject. For example timezones are not considered for Date and Timestamp in arrays.

Blackmorse:
Is it the same issue?

`Timestamp[] arr = {...};`
`statement.setArray(2, new ClickHouseArray(type, Array));`

throws an exception:

` DB::Exception: Cannot parse input: expected \' before: .358\',\'2019-04-03 16:52:43.358\']\n2019-04-03 13:52:43\t[\'2019-04-03 15:52:43.358\',\'2019-04-03 16:52:43.358\']\n: (at row 1)`

TanVD:
Looks like

zhicwu:
The driver has transitioned to the RowBinary data format starting from v0.3.2. It's now recommended to utilize input functions and appropriate objects for insertion, such as LocalDateTime[] in this instance.

PR #352, opened 2019-06-16
Harmonize Parameters

### LINKED ISSUE #325: Return incorrect type for LowCardinality(Nullable(Int64))
I have a table like this:

```
show create table bin_.t1;

SHOW CREATE TABLE bin_.t1

┌─statement──────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┐
│ CREATE TABLE bin_.t1 ( t1 LowCardinality(Nullable(Int64)),  _partition_day Date MATERIALIZED today(),  _insertTime DateTime MATERIALIZED now(),  uuid UUID MATERIALIZED generateUUIDv4()) ENGINE = Distributed(bip_ck_cluster, shardt, bin__t1_replica, rand()) │
└────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┘
```

The column type is LowCardinality(Nullable(Int64)), I got type 12(VARCHAR) from jdbc driver:
```
resultSet = connection.getMetaData().getColumns(null, dbName, tableName, null);
``` 
ResultSet:
```
TABLE_CAT -> default
TABLE_SCHEM -> bin_
TABLE_NAME ->  t1
COLUMN_NAME ->  t1
DATA_TYPE ->  12
TYPE_NAME ->  LowCardinality(Nullable(Int64))
```

enqueue:
* Replace `TypeUtils` with `ClickHouseColumnInfo`, explicit enum for data types instead of inconsistent ad-hoc String parsing (`ClickHouseDataType`) -- This change should improve the maintainability because the code is now easier to understand and to test. Fix #325 
* Try to honour the timezone of a `DateTime` column when parsing its values. Currently, ClickHouse is a little bit inconsistent when it comes to using date times in queries (see [yandex/ClickHouse#5206], but this inconsistency helps us, too.
* Harmonize behavior for normal and array parameters (related to issues #153)
* Change (!) behavior for `byte[]` parameters. In Java, we usually use byte-Arrays for binary data, so it's a bit different to other types of arrays. This change will hopefully reduce the headaches when trying to mangle some binary data into ClickHouse FixedString columns (related to issue #283)
* Better support for UUID data type

Sorry that this PR is so large. I started with the `ClickHouseDataType` but then found out that the other issues were rather easy to address along the way.


enqueue:
Hi @serebrserg could you perhaps pick up this PR? If you are not comfortable with any of the changes, please let me know, and I can do the necessary adjustments.

alexey-milovidov:
We have a new maintainer: @alex-krash

enqueue:
Sorry, did not know that. :wave: @alex-krash

src/main/java/ru/yandex/clickhouse/ClickHousePreparedStatementParameter.java
```diff
@@ -0,0 +1,49 @@
...
+public final class ClickHousePreparedStatementParameter {
+
+    private final String stringValue;
+    private final boolean quoteNeeded;
+
+    public static ClickHousePreparedStatementParameter fromObject(Object x,
+        TimeZone dateTimeZone, TimeZone dateTimeTimeZone)
+    {
+        if (x == null) {
+            return new ClickHousePreparedStatementParameter(null, false);
```
alex-krash: 1.) Let's make this object a static final field, and use it, in order to avoid extra allocations.
2.) Provide public static method ClickHousePreparedStatementParameter::null(), returning this field.
(use here and in PreparedStatement.setNull())

src/main/java/ru/yandex/clickhouse/response/ClickHouseResultSet.java
```diff
@@ -74,13 +78,23 @@
...
         this.usesWithTotals = usesWithTotals;
-        initTimeZone(timezone);
+        this.dateTimeTimeZone = timeZone;
+        this.dateTimeZone = properties.isUseServerTimeZoneForDates()
+            ? timeZone
+            : TimeZone.getDefault();
+        dateTimeFormat = new SimpleDateFormat(DATE_TIME_PATTERN);
+        dateTimeFormat.setTimeZone(dateTimeTimeZone);
+        dateFormat = new SimpleDateFormat(DATE_PATTERN);
+        dateFormat.setTimeZone(dateTimeZone);
```
alex-krash: Let's move this fields initialization back to field declaration

src/main/java/ru/yandex/clickhouse/util/ClickHouseArrayUtil.java
```diff
@@ -158,20 +170,29 @@ private static boolean needQuote(Object[] objects) {
...
     private static class ArrayBuilder {
-        private final StringBuilder builder = new StringBuilder();
+
+        private final StringBuilder builder;
         private final boolean quote;
-        private final boolean explicitEscape;
+        private final TimeZone dateTimeZone;
+        private final TimeZone dateTimeTimeZone;
         private int size = 0;
         private boolean built = false;
```
alex-krash: Please, fix built = true; in method build();

src/main/java/ru/yandex/clickhouse/response/ClickHouseColumnInfo.java
```diff
@@ -0,0 +1,179 @@
...
+    private int precision;
+    private int scale;
+
+    public static ClickHouseColumnInfo parse(String typeInfo, String columnName) {
+        ClickHouseColumnInfo column = new ClickHouseColumnInfo(typeInfo, columnName);
+        int currIdx = 0;
+        while (typeInfo.startsWith(KEYWORD_ARRAY, currIdx)) {
+            column.arrayLevel++;
+            column.clickHouseDataType = ClickHouseDataType.Array;
+            currIdx += 6;
```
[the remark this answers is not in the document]
enqueue: Using the constants' lengths + 1 for parenthesis.

src/main/java/ru/yandex/clickhouse/response/ClickHouseResultSet.java
enqueue: This is inconsistent with the other fields, but OK...

enqueue:
Thank you for your review. I incorporated the changes requested.

Please double-check that the behaviour for `byte[]` is OK. I think it is more intuitive this way, and most users are not interested in sending individual bytes as an array, but in filling `FixedString` columns or something similar instead. However, this change is not backwards-compatible.

alex-krash:
> Please double-check that the behaviour for byte[] is OK.

@enqueue , could you, please, explain by example?

enqueue:
@alex-krash  I will try. There are two parts to this

1. Single values

Biggest part of the change is in  `ClickHouseValueFormatter.formatBytes`.

See the tests `ClickHousePreparedStatementTest.testBytes`, `ClickHousePreparedStatementTest.testSetBytesNormal` etc.

**Old Behavior**

Failure due to missing quotes. If we added missing quotes: `INSERT INTO foo (bar) VALUES ('foo')`

**New Behavior**

Bytes are transmitted as a string of hex-encoded byte values: `INSERT INTO foo (bar) VALUES ('\x66\x6F\x6F')`

2. Arrays/lists of `byte[]`

Biggest part of the change is in `ClickHouseArrayUtil.toString(byte[] values)`

See test `ClickHouseArrayUtilTest.testCollectionToString`

```
List<byte[]> listOfByteArrays = new ArrayList<byte[]>();
listOfByteArrays.add("foo".getBytes("UTF-8"));
listOfByteArrays.add("bar".getBytes("UTF-8"));
```

**Old Behavior**

Array/List of `byte[]` is transmitted as array of array of individual byte String representations: `[[102,111,111],[98,97,114]]`

**New Behavior**

Array/List of `byte[]` is transmitted as array of sequence of hex-encoded byte values: `['\x66\x6F\x6F','\x62\x61\x72']`

alex-krash:
@enqueue , thanks!

PR #355, opened 2019-06-29
Add getTimestamp with Calendar

xhochy:
These methods are only declared as abstract but aren't implemented.


xhochy:
*bump* could a maintainer have a look at this?

PR #367, opened 2019-08-20
Add support for 307 redirects

Meegooo:
Currently it will use GET requests to check for redirects, because Apache HTTP client does not support early responses, and because of difficulties with reproducing non-repeatable entities.

Off by default, can be turned on by setting `checkForRedirects` to `true`.


alex-krash:
@Meegooo, could you, please, clarify use case for it?
ClickHouse cluster behind balancer?

Meegooo:
@alex-krash 
For instance. Combined with #368 you can setup a load balancer that will redirect requests to different servers.

PR #375, opened 2019-08-29
Fix ClickHouse docker image version

simPod:
To prevent different results when running CI


zhicwu:
Thanks @simPod. Now it will test against LTS versions starting from 19.14 as well as latest, please see [this](../blob/develop/.github/workflows/build.yml#L28) for details.

PR #377, opened 2019-08-30
Fix default column expression

enqueue:
Retrieving DB MetaData for table which contains columns with defaults values, results in Exception `no column default_expresseion in columns list database table name type default_type default_expression`


enqueue:
Hi @alex-krash I think the actual issue has been fixed meanwhile, but please cherry-pick the test.

zhicwu:
The issue has been fixed and the unit test is merged into #538.

Issue #304, opened 2019-02-14
Contributing Guidelines / Code Style


enqueue:
Please add some [contributing guidelines](https://help.github.com/articles/setting-guidelines-for-repository-contributors) for this project, including code style.

Since the current code styles are different between the contributors, and there are patches with many useless changes in the import order, it might be advisable to enforce some rules via a [checkstyle](http://checkstyle.sourceforge.net/) configuration. Perhaps Yandex has already such a configuration, else simply start with one of the popular standards, e.g. [Sun / Oracle Code Conventions](https://www.oracle.com/technetwork/java/javase/documentation/codeconvtoc-136057.html) or [Google Java Styleguide](http://google.github.io/styleguide/javaguide.html).

I am volunteering to help bring the current code base in shape once such a script is installed in the CI.

enqueue:
I would be happy to help here. Do you guys have any preference regarding Java style, import orders etc.? Else I would suggest to implement regular CheckStyle with Sun rules.

chernser:
Closed by https://github.com/ClickHouse/clickhouse-java/pull/2859

PR #385, opened 2019-09-05
Speeded up setBoolean 10x times

alex-krash:
Following pipeline:
1.) String value = bool ? "1" : "0"
2.) Create new parameter holder for value
Replaced with:
bool ? trueParameter : falseParameter.
Got performance increase 10x times for setBoolean() sequential call.
@enqueue , FYI.


enqueue:
@alex-krash yes, good to see such performance gains by simple "tricks" :+1: 

However, the PR highlights for me the need for some very basic coding rules: Changes in import order, and the use of the wildcard import "litter" this diff. Can we make an attempt to get #304 on the way?

PR #387, opened 2019-09-06
Added isBeforeFirst, isFirst, isAfterLast methods from JDBC spec.

alex-zaitsev:


alex-krash:
@alex-zaitsev , thanks!

PR #390, opened 2019-09-15
Nested arrays support

timurib:
It makes array parser recursive (we aware of the array depth in advance, and leaf nodes are always on the same level).

Just in case, the max depth is restricted: ≤ 32.


alex-krash:
@timurib , thanks!

PR #393, opened 2019-10-08
Doesn't quote generated arrays if a collection was given.

andyHa:
If a collection is passed in it will automatically be converted
into an array literal. However, like for normal arrays, this
generated string must not be quoted again when being
inserted into a prepared statement.

ClickHouseValueFormatter#formatObject (L155)
will happy convert Collections into appropriate array
literals but these will be quoted again and then rejected
by the Clickhosue server as a true array instead of a
string literal is expected.


alex-krash:
@andyHa , thanks!

PR #396, opened 2019-10-08
read FixedString from BinaryInputStream

ayudovin:
`ClickHouseRowBinaryInputStream` can't read FixedString(n) and throw exception in this case, therefore the functionality for reading FixedString(n) has been added.

this [issue](https://github.com/ClickHouse/clickhouse-jdbc/issues/394)


alex-krash:
@ayudovin , please add a test, checking behavior for zero-padded string (e.g. when FixedString(10) and inserted "hello", should return "hello\0\0\0\0\0"

ayudovin:
@alex-krash, I have added changes by your suggestions.

orloffv:
insert with fixedString is already supported?

ayudovin:
> insert with fixedString is already supported?

general write string should work for you

alex-krash:
@ayudovin , thanks!

PR #410, opened 2019-12-10
Bugfix for Boolean type formatting

Fox0117:
Boolean -> ClickHousePreparedStatementParameter using ClickHousePreparedStatementParameter.fromObject produces value with single quotes (is recognized as String by ClickHouse, not as Uint8, as it should be)


alex-krash:
@Fox0117 , can we have some test for this, ensuring everything works fine?

alex-krash:
@AndreevDm , @Fox0117 , thanks for contribution!

Issue #406, opened 2019-11-18
JDBC 4.2 Support


enqueue:
The driver should support JDBC 4.2 API, because clients expect driver to work with API as included in Java SE 8. 

* Major changes: [JDBC 4.2](https://docs.oracle.com/javase/8/docs/technotes/guides/jdbc/jdbc_42.html)
* Spec: [JSR 221 maintenance release 2](https://jcp.org/aboutJava/communityprocess/mrel/jsr221/index2.html)

This also means dropping support for Java SE 6.

zhicwu:
@enqueue, I hope it's not too late to get back to this. I'm working on `0.2.5` release which is just about small changes like but fixing and small enhancement. I hope it can be released in mid of Jan, and then if you have time, maybe we can make JDBC 4.2 support as part of `0.3.0` in Feb/Mar?

zhicwu:
Will timezone issue #590 in PR #589.

PR #418, opened 2020-01-11
Prepare for JDBC 4.2 #406

enqueue:
Enhance parsing and retrieving of date time values, especially using `java.time` objects (issue #406 )

* Remove support for Java 1.6, add Java 11 checks (this will also allow us to use newer library versions soon, and perhaps get rid of some altogether)
* Update maven plugins
* Fix some Javadocs to make it work with Java 11 build
* Implement "old" methods in `PreparedStatement`
* Support `java.time` Objects
* Breaking change: SQL type FLOAT is strange, `Float32` is actually a SQL type REAL according to my understanding. I do not have access to any of the SQL standards, but see for example [PostgreSQL manual](https://www.postgresql.org/docs/10/datatype-numeric.html), [MS SQL Server manual](https://docs.microsoft.com/en-us/sql/odbc/reference/appendixes/sql-data-types?view=sql-server-ver15), [Oracle Datatypes](https://docs.oracle.com/cd/B28359_01/server.111/b28286/sql_elements001.htm#SQLRF00213)
* Deprecate non-JDBC methods
* Remove `ByteFragmentUtils`
* Add a test for JDBC Table B-3 conformance
* Add a test for JDBC Table B-1 conformance
* Added two tables to explain how the PreparedStatement and ResultSet implementations deal with different date time values (see docs/datetime.md).

I am sorry this PR is so large. I tried to create individual commits at least for _some_ specific parts. If you have a good suggestion on how to split the PR, please let me know.


enqueue:
@alex-krash did you have time to take a look at this PR? I am happy to discuss any issues. Have a good day!

alex-krash:
@enqueue , sorry for delay. Will take a look now.
Looks like it's time to start new release (0.3), with breaking Java6 support, and incorporating your changes.

pom.xml
```diff
@@ -4,7 +4,7 @@
 
     <groupId>ru.yandex.clickhouse</groupId>
     <artifactId>clickhouse-jdbc</artifactId>
-    <version>0.2.4</version>
+    <version>0.2.4-enqueue</version>
```
alex-krash: Please, avoid changes like this, in order not to pollute master branch.

src/main/java/ru/yandex/clickhouse/response/parser/ClickHouseDateValueParser.java
```diff
@@ -0,0 +1,175 @@
...
+            case Int8:
+            case Int16:
+            case Int32:
+            case Int64:
+            case UInt8:
+            case UInt16:
+            case UInt32:
+            case UInt64:
+                try {
+                    long l = Long.parseLong(s);
```
alex-krash: 1. We may have UInt64 overflow here
2. Why all these things in class `ClickHouseDateValueParser`? That's really confusing.

src/main/java/ru/yandex/clickhouse/response/parser/ClickHouseLocalTimeParser.java
```diff
@@ -0,0 +1,63 @@
...
+
+    private ClickHouseLocalTimeParser() {
+        super(LocalTime.class);
+    }
+
+    @Override
+    LocalTime parseDate(String value, ClickHouseColumnInfo columnInfo,
+        TimeZone timeZone)
+    {
+        return LocalTime.MIDNIGHT;
```
alex-krash: Are you sure, that it is ok?

enqueue: It is the most straightforward response in my opinion. The ClickHouse column is of type Date, e.g. "2020-03-07". If we adjusted midnight server time zone into JVM time zone, we would perhaps cause more confusion.

pom.xml
enqueue: Sorry, this was a mistake

src/main/java/ru/yandex/clickhouse/response/parser/ClickHouseDateValueParser.java
enqueue: 1. Thanks, I had not thought of this. New implementation truncates to milliseconds.
2. The code should not be confusing, let's see if we can improve it.  `ClickHouseDateValueParser` provides basic functionality for all date/time parsing. This makes the individual implementations much easier to read. We _could_ move the "utility" methods (lines 146 ff.) into the actual implementations. Do you think that this would help overall simplicity?

enqueue:
Hi @alex-krash , thank you for taking the time to review this very large PR :heart: 

Moving to Java 8 is an important step that will also help to keep libraries up to date. 0.2 releases containing the state of development as of now will still available for die-hard Java 6 users.

hopsoft:
Any updates on when this PR might get merged and released?

docs/datetime.md
```diff
@@ -0,0 +1,69 @@
...
+
+The methods which take a [Calendar]((https://docs.oracle.com/javase/8/docs/api/java.base/java/util/Calendar.html) argument behave the same as the corresponding methods without such an argument. The API documentation says something like
+
+> This method uses the given calendar to construct an appropriate millisecond value for the _x_ if the underlying database does not store timezone information.
+
+For Date and DateTime fields, the JDBC driver has enough time zone related information available, so these methods would only be relevant for String or other typed fields. There might be valid use cases, but for now we think that adding such an option would make things even more complicated.
+
+Requested Type | Number | Date | DateTime | Other 
+---------------| ----------------------------------
```
pan3793, line 54: missing `|`

enqueue: Thank you @pan3793 ! After almost a year, someone stumbles upon wrongly formatted tables in the documentation... :smiley:  I added a couple of pipes, and the table looks better now in the preview.

pan3793: There are so many tricks about timezone, especially in legacy Java `Date` API. I also encountered some issues when migrated to `java.time` API from legacy `Date` API in another ClickHouse JDBC driver implement in native protocol. And would you mind me pick up some code and doc from this PR?

enqueue: Yes, life would be boring without time zones and leap seconds. Feel free to use the code in this PR, but please let me know when you find any error, so I can fix it. Have a good weekend!

alongls:
Hello
is there any update regarding the driver compatibility to the current release of Metabase ?
thanks.

enqueue:
Hi @alongls 

this PR is for improvements in the official ClickHouse JDBC driver. Please follow  enqueue/metabase-clickhouse-driver#58 for an updated Metabase driver.

We use this branch for the driver, so it would be nice to see the PR reviewed and merged and use the official distribution, but perhaps it's too much work for maintainer due to size or other reasons.

zhicwu:
Thanks @enqueue. This is really a large PR. I quickly walked through the changes today, but I still need a few more days to play with the code(and maybe benchmark too).

zhicwu:
@verify 21.2.3.15

zhicwu:
@enqueue, the build break is caused by below code snippets in `TimeZoneTest`:
```java
Long offset = Long.valueOf(
    TimeUnit.MILLISECONDS.toHours(
        connectionServerTz.getTimeZone().getOffset(currentTime)) - 1);
```
I'm not sure why we need to deduct one here - it works if I remove that part. Could you take a look?

zhicwu:
I'll take the risk to merge this large PR in order to unblock future development. I'll create separate PR to fix above issues as well as existing ones in `ClickHousePreparedStatementTest` and those introduced in PR #571.

zhicwu:
@enqueue, I'm not sure if `Date` is handled properly, but now `Time` and `Timestamp` seem good. I think we probably need some more time to compare the tests against clickhouse-client and other drivers. As you may noticed, I removed the hard-coded timezone in pom.xml and enhanced build process to test a few selected timezones. In addition to that, you can now use below command to test more scenarios:
```bash
mvn -Duser.timezone=Etc/UTC -DclickhouseTimezone=Europe/Moscow clean verify
```

Anyway, I'll restructure directories tomorrow and start to add more features in the next following days. If you have time, you may want to review the changes I made in PR #580 and the 3 test cases I temporarily disabled(search `@Test(enabled = false)`).

PR #443, opened 2020-04-16
I moved all string related methods from Utils to already existing StringUtils

strogiyotec:
methods:
```
startsWithIgnoreCase
retainUnquoted
```
were moved from Utils to StringUtils
Tests are also povided inside
```
ru.yandex.clickhouse.util.apache
```
package


zhicwu:
Close outdated PR.

FYI, the driver has been rewritten and the package name will be changed to `com.clickhouse`.

PR #475, opened 2020-07-10
fix bug when select with FORMAT JSONEachRow

peggypig:
if sql is a select with FORMAT JSONEachRow, result has no headerFragment and typesFragment from clickhouse server. but first and second json data will be used for headerFragment and typesFragment  in ru.yandex.clickhouse.response.ClickHouseResultSet,this will result in the loss of two records.


peggypig:
retry

peggypig:
close and retry

MadDoxXimba:
Hello, may i know why this issue has been closed ?

I'm experiencing exactly the same issue !

PR #485, opened 2020-08-16
Expose the pointer for the LZ4 Compression buffer

mdonkers:
Not sure if this is wanted / useful for others.

But in our case, we aren't using the full JDBC library, just this class to compress the `INSERT` queries (because we're doing some batching of records, then push as one batch to a HTTP output stream).

Because of also other buffers, it is useful to know how much data is buffered into the LZ4 compression buffers, preventing having to call `flush()` regularly.


CodingFabian:
@alex-krash sorry for the ping. Any chance you pick up this contribution? Would help us to eliminate a manually patched jar, and potentially beneficial to others

zhicwu:
> @alex-krash sorry for the ping. Any chance you pick up this contribution? Would help us to eliminate a manually patched jar, and potentially beneficial to others

Could you change target to develop? I can help to merge the PR and make it part of the upcoming 0.2.5 release.

mdonkers:
Thank you!
I've changed base to `develop`.

PR #499, opened 2020-10-05
Restore public getColumnNames method

Okapist:
We use it


zhicwu:
Sorry for the breaking change. Could you change target to `develop`? I can then merge the PR and make it part of the upcoming 0.2.6 release.

Okapist:
Thanks. I missed previous message.

PR #500, opened 2020-10-05
update supported JDK

Okapist:
Fix Travis ci builds


Okapist:
jdk11 is broken in Travis CI. There 11.0.2 hardcoded version.

PR #502, opened 2020-10-08
Add the query_id in the ClickHouseProperties

alifirat:
# Description 

I had an use case in which I need to set the ClickHouse query ID when executing the query. In the release that I was using, when setting the ClickHouse query_id, this ones was override in ClickHouse by an another UUID.  

The goal of this PR is to add the `query_id` in the ClickHouseProperties and to be able to find into ClickHouse when requesting the `query_log` table.

### DIFF
--- a/src/main/java/ru/yandex/clickhouse/settings/ClickHouseProperties.java
+++ b/src/main/java/ru/yandex/clickhouse/settings/ClickHouseProperties.java
@@ -93,6 +93,7 @@ public class ClickHouseProperties {
     private Boolean insertDeduplicate;
     private Boolean insertDistributedSync;
     private Boolean anyJoinDistinctRightTableKeys;
+    private String queryId;
 
 
     public ClickHouseProperties() {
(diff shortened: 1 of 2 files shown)


zhicwu:
Thank you @alifirat for the pull request. `ClickHouseProperties` can be shared among multiple connections, so the change may introduce duplicated query id. You may pass `Map<ClickHouseQueryParam, String>` to `ClickHouseStatement` in order to use custom query id or any other query parameters.

LatchShun:
In version 0.3.2, the `ClickHouseStatement` can't pass `Map<ClickHouseQueryParam, String>` parameter, or choose http request.

zhicwu:
> In version 0.3.2, the `ClickHouseStatement` can't pass `Map<ClickHouseQueryParam, String>` parameter, or choose http request.

`ClickHouseQueryParam` no longer exists since the driver has been rewritten. If you need to customize settings, you probably should use `clickhouse-client` or `clickhouse-http-client` instead of the JDBC driver. In case you just need a quick workaround and don't mind mixed API usage, you may try below: 
```java
statement.unwrap(ClickHouseRequest.class)
  // .set(<setting>, <value>)
  .query(<your query>, <query id>)
  .execute()
```

PR #509, opened 2020-10-19
fix bug when select with FORMAT JSONEachRow

peggypig:
if sql is a select with FORMAT JSONEachRow, result has no headerFragment and typesFragment from clickhouse server. but first and second json data will be used for headerFragment and typesFragment in ru.yandex.clickhouse.response.ClickHouseResultSet,this will result in the loss of two records.


zhicwu:
Close as this has been fixed on develop branch. All text formats without header including `TSV` will be treated this way: 1) each line is a row(0x0A as separator); 2) one column `results Nullable(String)`.

PR #526, opened 2020-12-08
implement ResultSet findColumn() method issue#525

dangerous1990:
```
    /**
     * Maps the given <code>ResultSet</code> column label to its
     * <code>ResultSet</code> column index.
     *
     * @param columnLabel the label for the column specified with the SQL AS clause.  If the SQL AS clause was not specified, then the label is the name of the column
     * @return the column index of the given column name
     * @exception SQLException if the <code>ResultSet</code> object
     * does not contain a column labeled <code>columnLabel</code>, a database access error occurs
     *  or this method is called on a closed result set
     */
    int findColumn(String columnLabel) throws SQLException;
```

### DIFF
--- a/src/main/java/ru/yandex/clickhouse/response/ClickHouseResultSet.java
+++ b/src/main/java/ru/yandex/clickhouse/response/ClickHouseResultSet.java
@@ -607,6 +607,11 @@ public class ClickHouseResultSet extends AbstractResultSet {
         }
     }
 
+    @Override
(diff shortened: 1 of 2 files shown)


zhicwu (commented):
Could you change target to `develop` and add unit tests accordingly?

dangerous1990:
> Could you change target to `develop` and add unit tests accordingly?

done

PR #534, opened 2021-01-03
fix #528 Error while extract schema

### LINKED ISSUE #528: Error while extract schema
Hello,

I found if using a function that contains keywords, like _**fromUnixTimestamp64Milli**_, I got an error result when extract db and table name.

Here is code snippet:

```
private String extractDBAndTableName(String sql) {
        if (Utils.startsWithIgnoreCase(sql, "select")) {
            String withoutStrings = Utils.retainUnquoted(sql, '\'');
            int fromIndex = withoutStrings.indexOf("from");
            if (fromIndex == -1) {
                fromIndex = withoutStrings.indexOf("FROM");
            }
            if (fromIndex != -1) {
                String fromFrom = withoutStrings.substring(fromIndex);
                String fromTable = fromFrom.substring("from".length()).trim();
                return fromTable.split(" ")[0];
            }
        }
        if (Utils.startsWithIgnoreCase(sql, "desc")) {
            return "system.columns";
        }
        if (Utils.startsWithIgnoreCase(sql, "show")) {
            return "system.tables";
        }
        return "system.unknown";
    }
```

den-crane:
fixes #528 

added support for queries with  fromUnixTimestamp64Milli : `/*qq*/ SELECT fromUnixTimestamp64Milli(time)from table`

### DIFF
--- a/src/main/java/ru/yandex/clickhouse/ClickHouseStatementImpl.java
+++ b/src/main/java/ru/yandex/clickhouse/ClickHouseStatementImpl.java
@@ -36,6 +36,8 @@ import java.sql.ResultSet;
 import java.sql.SQLException;
 import java.sql.SQLWarning;
 import java.util.*;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
 
 
 public class ClickHouseStatementImpl implements ClickHouseStatement {
@@ -491,7 +493,7 @@ public class ClickHouseStatementImpl implements ClickHouseStatement {
         return false;
     }
 
-    private String extractTableName(String sql) {
+    static String extractTableName(String sql) {
         String s = extractDBAndTableName(sql);
         if (s.contains(".")) {
             return s.substring(s.indexOf(".") + 1);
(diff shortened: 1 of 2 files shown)


zhicwu:
Thanks @den-crane for fixing the issue. I noticed quite some bugs related to SQL parsing, so I'd suggest to use [antlr4](https://github.com/antlr/antlr4) instead. Meanwhile, I like the idea of creating a short-term workaround. Can we avoid using regular expression and try [this approach](https://github.com/ClickHouse/clickhouse-jdbc/blob/release_0.2.4/src/main/java/ru/yandex/clickhouse/ClickHouseStatementImpl.java#L474) instead for better performance?

zhicwu:
I added an experimental parser in #563 and hopefully it can fix this kind of issue. I'll merge your tests and do more testing in the next following days.

zhicwu:
Test cases are merged into `ClickHouseSqlParserTest` in #563 with minor changes to ensure it's a valid SQL.

Issue #480, opened 2020-07-23
ClickHouseDatabaseMetadata.getColumns doesn't return column comments


serge-rider:
It just doesn't read it from system.columns.

According to JDBC API it should be returned in REMARKS column.

Please add column comments (they are very handy for various UI/CLI tools).

PR #538, opened 2021-01-07
Make column comments available via DatabaseMetaData REMARKS

enqueue:
Fix for #480

Column comments have been available for some years in ClickHouse server, so I assume it is safe to refer to the column. If we want to support older server versions, please give me a hint for which minimum version to check.

### DIFF
--- a/src/main/java/ru/yandex/clickhouse/ClickHouseDatabaseMetadata.java
+++ b/src/main/java/ru/yandex/clickhouse/ClickHouseDatabaseMetadata.java
@@ -803,10 +803,10 @@ public class ClickHouseDatabaseMetadata implements DatabaseMetaData {
         StringBuilder query;
         if (connection.getServerVersion().compareTo("1.1.54237") > 0) {
             query = new StringBuilder(
-                "SELECT database, table, name, type, default_kind as default_type, default_expression ");
+                "SELECT database, table, name, type, default_kind as default_type, default_expression, comment ");
         } else {
             query = new StringBuilder(
-                "SELECT database, table, name, type, default_type, default_expression ");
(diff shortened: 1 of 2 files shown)


zhicwu:
Thanks @enqueue. Do you mind to change target to `develop`? It's probably inconvenient, but I hope we can give new code some more time staying on develop branch and only merge released code to master.

Moreover, if you have extra time, I hope you can help to review my PR as well :)

src/main/java/ru/yandex/clickhouse/ClickHouseDatabaseMetadata.java
```diff
@@ -856,7 +856,9 @@ public ResultSet getColumns(String catalog, String schemaPattern, String tableNa
                 ? String.valueOf(columnNullable)
                 : String.valueOf(columnNoNulls));
             //remarks
-            row.add(null);
+            row.add("".equals(descTable.getString("comment"))
```
zhicwu, line 859: Why we treat empty string as null? If that's required, can use a local variable to store comment so that we don't have to call getString twice?

zhicwu (commented):
Could you also merge the test from PR #377? The issue was fixed by PR #382 but no test.

src/main/java/ru/yandex/clickhouse/ClickHouseDatabaseMetadata.java
enqueue: ClickHouse returns an empty String for columns without a comment (column is not `Nullable`). Why I think that translating this empty String to `null` is a good behavior in this case:

1. A typical Java user would expect this to be `null` when no comment has been set.
2. Returning an empty String for a column which does not have a comment set would be a change in this driver's behavior. This is covered by an existing test, too.

I removed the duplicate call to `getString`.

Issue #535, opened 2021-01-03
SQL parser instead of regular expression


zhicwu:
In order to completely address SQL parsing issues, let's reuse the same antlr4 grammar([Lexer](https://github.com/ClickHouse/ClickHouse/blob/master/src/Parsers/New/ClickHouseLexer.g4) & [Parser](https://github.com/ClickHouse/ClickHouse/blob/master/src/Parsers/New/ClickHouseParser.g4)) from ClickHouse server.

Related bugs: #479, #528, #533

Update:
ANTLR is overkill and it will introduce 300KB runtime, which is the opposite of what we need(reducing dependencies). As a result, we'll need to implement a simplified parser which is able recognize case-insensitive keywords, and understand if the given query contains multiple statements and type of each statement(DDL or DML, query or mutation etc.).

Issue #495, opened 2020-09-29
NULL constant in PreparedStatement


bgranvea:
it seems to be related to #261.

I cannot use NULL constant in PreparedStatement, I get an error "Cannot parse expression of type Nullable(String) here: \N)"

Test case:

```
create table test (str Nullable(String)) ENGINE=MergeTree ORDER BY tuple();

@Test
public void testRequest() throws SQLException {
    try (Connection c = DriverManager.getConnection("jdbc:clickhouse://localhost:8123", "xxxx", "xxxx") {
        // works
        Statement stmt = c.createStatement();
        stmt.executeUpdate("INSERT INTO test(str) VALUES (null)");

        // doesn't work
        PreparedStatement ps = c.prepareStatement("INSERT INTO test(str) VALUES (null)");
        ps.executeUpdate();
    }
}
```

Here is what is sent to the server:

```
2020/09/29 10:01:09:724 CEST [DEBUG] wire - http-outgoing-0 >> "POST /?compress=1&query_id=26eca5df-cdcc-44cb-8c37-e62291f7d835&database=default&password=xxxx&user=xxxx HTTP/1.1[\r][\n]"
2020/09/29 10:01:09:725 CEST [DEBUG] wire - http-outgoing-0 >> "Content-Length: 33[\r][\n]"
2020/09/29 10:01:09:725 CEST [DEBUG] wire - http-outgoing-0 >> "Content-Type: text/plain; charset=UTF-8[\r][\n]"
2020/09/29 10:01:09:725 CEST [DEBUG] wire - http-outgoing-0 >> "Host: localhost:8123[\r][\n]"
2020/09/29 10:01:09:725 CEST [DEBUG] wire - http-outgoing-0 >> "Connection: Keep-Alive[\r][\n]"
2020/09/29 10:01:09:726 CEST [DEBUG] wire - http-outgoing-0 >> "User-Agent: Apache-HttpClient/4.5.3 (Java/1.8.0_201)[\r][\n]"
2020/09/29 10:01:09:726 CEST [DEBUG] wire - http-outgoing-0 >> "[\r][\n]"
2020/09/29 10:01:09:726 CEST [DEBUG] wire - http-outgoing-0 >> "INSERT INTO test(str) VALUES (\N)"
2020/09/29 10:01:09:764 CEST [DEBUG] wire - http-outgoing-0 << "HTTP/1.1 400 Bad Request[\r][\n]"

```
[...]

PR #541, opened 2021-01-08
Static null parameters for non-batch PreparedStatement

enqueue:
Fix for #495 It's not beautiful, but should fix the bug (at least the tests are green). I expect some major clean-up / refactoring in `ClickHousePreparedStatementImpl` and friends after 0.2.5 (e.g. via #418 and perhaps #535)


zhicwu:
Sorry for the late response. I was evaluating ANTLR4. Overall, I think it's overkill not to mention it will introduce another 300KB dependency. I'm good for the workaround.

Issue #522, opened 2020-11-20
The user name and password are printed in the debug level log


huldarchen:
ClickHouseStatementImpl：L578。

TTcheng:
![image](https://user-images.githubusercontent.com/19683420/102975239-b27eb480-453a-11eb-9056-d3ae775de026.png)
I found the same issue occured in my project logs,It seems like that the request uses plain text password, this is extremely dangerous

enqueue:
I see a couple of short-term, i.e. 0.2.5 target, options here. What are your opinions?

1. Hard-coded RegExp filter which can be used in `ClickhouseStatementImpl`, and perhaps one or five more places
2. Try to find a way to install RegExp filter somehow. I don't think there is a way built-into SLF4J without interfering with users' implementation settings, though. I don't like the idea of introducing something like `ClickHouseLogger.getLogger()` (which would include such a filter) and forcing all developers to use that.
3. Use [HTTP Basic Auth](https://clickhouse.tech/docs/en/interfaces/http) and get rid of these query parameters. I am unsure because original driver authors chose otherwise. Do you see any compatibility issues?

zhicwu:
> I see a couple of short-term, i.e. 0.2.5 target, options here. What are your opinions?

Thanks @enqueue. I agree with you that the first two are not ideal, while the third is better. Actually, in a private version of ClickHouse JDBC driver, I added `authMode` setting and made basic auth as default. It was not only for security reason but also to ensure the driver works with clickhouse server(s) behind proxy like httpd/nginx/traefik. I think we should go with the third. However, we may still need to keep query parameter for older version. We can add `19.14` into [this](blob/develop/.github/workflows/build.yml#L28) and add regression test case accordingly. By the way, there'll be one or two cases failed on 19.14 ;)

@alexey-milovidov / @den-crane / @filimonov, do you have impression of why we use query parameter to pass credential instead of basic auth in http header? Any specific version of clickhouse server should be added into regression?

filimonov:
> do you have impression of why we use query parameter to pass credential instead of basic auth in http header? 

Have no idea. Basic auth is totally ok (and always was).

zhicwu:
@enqueue, I hope you don't mind that I merged PR #546 just now. Now `19.14` is added into regression. If you have docker installed, you can run tests against specific version of ClickHouse on develop branch by issue command line below:
```bash
mvn -DclickhouseVersion=19.17 clean verify
```

[...]

PR #498, opened 2020-10-05
Give access to stats returned by server in X-ClickHouse-Summary

bgranvea:
Clickhouse can return interesting statistics in a X-ClickHouse-Summary header if send_progress_in_http_headers parameter is set (see https://github.com/ClickHouse/ClickHouse/pull/5116). 

In my use case, I need this to know how many rows are inserted when executing a INSERT ... SELECT ...

PR #548, opened 2021-01-24
Use Basic auth for credentials instead of query params

enqueue:
Fix for #522

### DIFF
--- a/src/main/java/ru/yandex/clickhouse/ClickHouseStatementImpl.java
+++ b/src/main/java/ru/yandex/clickhouse/ClickHouseStatementImpl.java
@@ -1,32 +1,5 @@
 package ru.yandex.clickhouse;
 
-import com.google.common.base.Strings;
-import org.apache.http.Header;
-import org.apache.http.HttpEntity;
-import org.apache.http.HttpResponse;
-import org.apache.http.NameValuePair;
-import org.apache.http.client.methods.HttpGet;
-import org.apache.http.client.methods.HttpPost;
-import org.apache.http.client.utils.URIBuilder;
-import org.apache.http.entity.ContentType;
-import org.apache.http.entity.StringEntity;
-import org.apache.http.entity.mime.MultipartEntityBuilder;
-import org.apache.http.impl.client.CloseableHttpClient;
-import org.apache.http.message.BasicNameValuePair;
-import org.apache.http.util.EntityUtils;
-import org.slf4j.Logger;
-import org.slf4j.LoggerFactory;
-import ru.yandex.clickhouse.domain.ClickHouseFormat;
-import ru.yandex.clickhouse.except.ClickHouseException;
-import ru.yandex.clickhouse.except.ClickHouseExceptionSpecifier;
-import ru.yandex.clickhouse.response.*;
-import ru.yandex.clickhouse.settings.ClickHouseProperties;
-import ru.yandex.clickhouse.settings.ClickHouseQueryParam;
-import ru.yandex.clickhouse.util.ClickHouseRowBinaryInputStream;
-import ru.yandex.clickhouse.util.ClickHouseStreamCallback;
-import ru.yandex.clickhouse.util.Utils;
-import ru.yandex.clickhouse.util.guava.StreamUtils;
-
 import java.io.ByteArrayInputStream;
 import java.io.IOException;
 import java.io.InputStream;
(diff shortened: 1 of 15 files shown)


zhicwu:
Thank you @enqueue! Will it work when `httpAuthorization` is set?

enqueue:
> Thank you @enqueue! Will it work when `httpAuthorization` is set?

Good point! I will check it.

src/main/java/ru/yandex/clickhouse/util/ClickHouseHttpClientBuilder.java
```diff
@@ -209,4 +240,31 @@ private KeyStore getKeyStore()
...
+
+    private static HttpHost getTargetHost(ClickHouseProperties props) {
+        return new HttpHost(
+            props.getHost(),
+            props.getPort(),
+            props.getSsl() ? "https" : "http");
+    }
+
+    private static boolean isConfigurationValidForAuth(ClickHouseProperties props) {
+        return props.getHost() != null
```
zhicwu, line 266: I'm not sure why we need to check host here - it's always `null` in `BalancedClickhouseDataSource`. Will this cause authentication issue when connecting to a cluster?

src/test/java/ru/yandex/clickhouse/util/ClickHouseHttpClientBuilderTest.java
```diff
@@ -0,0 +1,147 @@
...
+            "basic");
+    }
+
+
+    @Test(dataProvider = "authUserPassword")
+    public void testHttpAuthParametersCombination(String authorization, String user,
+        String password, String expectedAuthHeader) throws Exception
+    {
+        ClickHouseProperties props = new ClickHouseProperties();
+        props.setHost("localhost");
```
zhicwu, line 106: Could you add one more test when host is null?

src/main/java/ru/yandex/clickhouse/ClickHouseStatementImpl.java
```diff
@@ -15,35 +31,33 @@
...
-import java.util.*;
-
 
 public class ClickHouseStatementImpl implements ClickHouseStatement {
 
     private static final Logger log = LoggerFactory.getLogger(ClickHouseStatementImpl.class);
 
     private final CloseableHttpClient client;
 
+    private final HttpClientContext httpContext;
```
zhicwu, line 59: Why not make it protected and create the context in constructor?

zhicwu:
Thank you @enqueue for making this pull request. I think we're close to get it merged. Mean time, can we remove HttpContext parameter from Statement and PreparedStatement constructors? Instead, we can create http context inside the constructor. This is because HttpContext is tied to http, so when we add support for more protocols, we'll have to remove it anyway because we only need to pass a client(e.g. HttpClient, NativeClient or GrpcClient) for execution.

Let me know if you have different opinion and we can discuss.

src/main/java/ru/yandex/clickhouse/ClickHouseStatementImpl.java
enqueue: This is analogous to the client. This class does not know how to construct these, but gets ready to use "connection stuff".

src/test/java/ru/yandex/clickhouse/util/ClickHouseHttpClientBuilderTest.java
enqueue: Sure

src/main/java/ru/yandex/clickhouse/util/ClickHouseHttpClientBuilder.java
enqueue: Good point. I had not thought about this. Will try to find out how the `BalancedClickhouseDataSource` works.

enqueue: Added some tests for common scenarios to `BalancedClickhouseDataSourceTest`, do you have any others you can think of? The host is set in 'ClickhouseJdbcUrlParser.parseClickhouseUrl` from the URL, so we hopefully do not have to change the implementation of `BalancedClickhouseDataSource`

src/main/java/ru/yandex/clickhouse/ClickHouseStatementImpl.java
enqueue: Well, I would say that the statement should simply get everything it needs to do its work. It does not need to know how to create connection context, but this is my subjective judgement. I can understand that we want to try and keep the number of ties with Apache HTTP Client under control. I am going to follow your suggestion.

enqueue:
Hi @zhicwu  I updated the PR to take into account your suggestions. I need your help getting the build to green though. The maven output says something about `ClickHouseResponseSummary` missing, but there isn't any code referring to this class. Running mvn verify locally works. What am I doing wrong? Thank you very much in advance.

zhicwu:
Thank you @enqueue. ResponseSummary was merged from PR #498. Have you merged all changes from upstream/develop? I'm out for launch but I'll look into this when I get back to home.

src/main/java/ru/yandex/clickhouse/ClickHouseStatementImpl.java
zhicwu: I see your point. Perhaps it's too early to consider refactoring. I'm fine to keep HttpContext argument now, but I think we may use a new interface like ClickHouseClient to replace both CloseableHtppClient and HttpContext, in order to hide protocol implementation details.

src/test/java/ru/yandex/clickhouse/util/ClickHouseHttpClientBuilderTest.java
zhicwu: Thought host is null when using BalancedDataSource but I'll double check :p

src/main/java/ru/yandex/clickhouse/util/ClickHouseHttpClientBuilder.java
zhicwu: I see. Thanks for the detailed explanation.  I don't have anything to add then.

PR #571, opened 2021-02-17
tests for ORC / Parquet

den-crane:
Tests. Load data from ORC / Parquet streams.

closes: https://github.com/ClickHouse/clickhouse-jdbc/issues/460 ,  https://github.com/ClickHouse/clickhouse-jdbc/issues/474


den-crane:
ORC does not work in Clickhouse docker without  /etc/localtime

type: orc::TimezoneError, e.what() = Can't open /etc/localtime

src/test/java/ru/yandex/clickhouse/ClickHouseContainerForTest.java
```diff
@@ -33,7 +34,24 @@
...
+        *               ))
+        */
+
+        final String fullImageName = "yandex/clickhouse-server" + imageTag;
+        final String runInstallTZ = "RUN apt-get update && apt-get install tzdata";
+
+        ImageFromDockerfile i = new ImageFromDockerfile();
+        i.withFileFromString("Dockerfile", String.format("FROM %s \n %s", fullImageName, runInstallTZ));
+
+        clickhouseContainer = new GenericContainer<>(i)
```
zhicwu, line 54: Why not put an executable shellscript(e.g. `install-prerequisities.sh`) under `src/test/resources/initdb.d` folder, and mount it into container like below?
```java
.withClasspathResourceMapping("initdb.d", "/docker-entrypoint-initdb.d", BindMode.READ_ONLY)
```

You may need to update `.gitignore` in order to add shellscript, which is something looks like below:
```bash
#!/bin/bash

set -e

if [ ! -f /etc/localtime  ]; then
        apt-get update && DEBIAN_FRONTEND="noninteractive" apt-get -y install tzdata
fi
```

zhicwu (changes requested):
Also you can update `jdk.version` to `1.8` in pom.xml as we're going to drop jdk 7 support starting from 0.3.0 :)

src/test/java/ru/yandex/clickhouse/ClickHouseContainerForTest.java
den-crane: My way allows local docker to cache TZdata in docker image cache. 
Your way will download TZdata from Ubuntu repo in runtime always.

zhicwu: Good point. But maybe you should check if `/etc/localtime` exists first? Since `tzdata` will be added into official docker image, we no longer need to run apt-get update/install later.

zhicwu (changes requested):

src/test/java/ru/yandex/clickhouse/ClickHouseContainerForTest.java
den-crane: > But maybe you should check if /etc/localtime exists first

It is impossible. It wont be a problem / overhead when it will be added into official docker image.
Eventually we will remove this code (apt-get) because CH server will switch to Alpine.

zhicwu (approved):

PR #574, opened 2021-02-22
join_algorithm query property

michail-nikolaev:


zhicwu:
Added query parameter in PR #633.

PR #579, opened 2021-02-26
test for large external data

den-crane:
test for large external data (100 rows) 
// some user complained in a chat that jdbc issue have such issue. But I could not reproduce, and made this test


zhicwu:
Looks good in [manually triggered build](../actions/runs/605812067).

Issue #569, opened 2021-02-15
Not able to query array of UUIDs


zhicwu:
Got below exception when query against array of UUIDs.

<details>
  <summary>java.lang.IllegalStateException...</summary>

```
java.lang.RuntimeException: Parse exception: ByteFragment{[['5ff22319-793d-4e6c-bdc1-916095a5a496']<TAB>1], start=0, len=40}
	at ru.yandex.clickhouse.response.ClickHouseResultSet.getObject(ClickHouseResultSet.java:541)
	at org.jkiss.dbeaver.model.impl.jdbc.exec.JDBCResultSetImpl.getObject(JDBCResultSetImpl.java:627)
	at org.jkiss.dbeaver.model.impl.jdbc.data.handlers.JDBCComplexValueHandler.fetchColumnValue(JDBCComplexValueHandler.java:49)
	at org.jkiss.dbeaver.model.impl.jdbc.data.handlers.JDBCAbstractValueHandler.fetchValueObject(JDBCAbstractValueHandler.java:49)
	at org.jkiss.dbeaver.ui.controls.resultset.ResultSetDataReceiver.fetchRow(ResultSetDataReceiver.java:125)
	at org.jkiss.dbeaver.ui.editors.sql.execute.SQLQueryJob.fetchQueryData(SQLQueryJob.java:717)
	at org.jkiss.dbeaver.ui.editors.sql.execute.SQLQueryJob.executeStatement(SQLQueryJob.java:540)
	at org.jkiss.dbeaver.ui.editors.sql.execute.SQLQueryJob.lambda$0(SQLQueryJob.java:440)
	at org.jkiss.dbeaver.model.exec.DBExecUtils.tryExecuteRecover(DBExecUtils.java:168)
	at org.jkiss.dbeaver.ui.editors.sql.execute.SQLQueryJob.executeSingleQuery(SQLQueryJob.java:427)
	at org.jkiss.dbeaver.ui.editors.sql.execute.SQLQueryJob.extractData(SQLQueryJob.java:812)
	at org.jkiss.dbeaver.ui.editors.sql.SQLEditor$QueryResultsContainer.readData(SQLEditor.java:3226)
	at org.jkiss.dbeaver.ui.controls.resultset.ResultSetJobDataRead.lambda$0(ResultSetJobDataRead.java:118)
	at org.jkiss.dbeaver.model.exec.DBExecUtils.tryExecuteRecover(DBExecUtils.java:168)
[...]

PR #588, opened 2021-03-04
Test UUID Array

enqueue:
Test for #569


zhicwu:
Please excuse the ugly workarounds I added for dealing with timezone failures. The build failure here is just one of them.  I reserved a week for fixing those issues :) A rough idea is to add more test cases using mysql/mariadb for comparison, and then make changes accordingly.

Will get back to this pull request once I commit the changes for adding more data types later today.

PR #596, opened 2021-03-21
RoaringBitmap support

zhicwu:
Object | Read | Write
-- | -- | --
RoaringBitmap  | Y | Y
ImmutableRoaringBitmap | as RoaringBitmap | Y
MutableRoaringBitmap | as RoaringBitmap | Y
Roaring64Bitmap | as Roaring64NavigableMap and only when cardinality <= 32 | as Roaring64NavigableMap
Roaring64NavigableMap | only when cardinality <= 32 | Y

**Usage**
```java
// use JDBC interface - NOT recommended before 0.3.1
try (PreparedStatement statement = connection.prepareStatement("insert into my_bitmap_table values(..., ?, ...)")) {
    ...
    // RoaringBitmap bitmap = RoaringBitmap.bitmapOf(1,2,3,...);
    s.setObject(index++, ClickHouseBitmap.wrap(bitmap, ClickHouseDataType.UInt32));
    ...
    // the actual SQL in 0.3.0 will be something like below, which is also why batch insertion does not work...
    // insert into my_bitmap_table values(..., bitmapBuild([toUInt32(1),toUInt32(3),toUInt32(3),...]) ...)
    s.execute();
}

// use extended API - recommended in 0.3.0
try (ClickHouseStatement statement = connection.createStatement()) {
    statement.sendRowBinaryStream("insert into my_bitmap_table", new ClickHouseStreamCallback() {
        public void writeTo(ClickHouseRowBinaryStream stream) throws IOException {
            ...
            // RoaringBitmap bitmap = RoaringBitmap.bitmapOf(1,2,3,...);
            // In addition to RoaringBitmap, you can pass:
            // ImmutableRoaringBitmap, MutableRoaringBitmap and even Roaring64NavigableMap
            stream.writeBitmap(ClickHouseBitmap.wrap(bitmap, ClickHouseDataType.UInt32));
            ...
        }
    });
}
```

PR #598, opened 2021-03-22
fix NegativeArraySizeException error

zzsmdfj:
It will throw NegativeArraySizeException when result is very large by execute sql with `connection.prepareStatement(sql).executeQuery()` 

The error stack：
[error stack](https://github.com/zzsmdfj/picture/blob/master/clickhouse/1616382893968-00c4c88f-991f-43b9-91c4-622c8c5d6652.png)

### DIFF
--- a/src/main/java/ru/yandex/clickhouse/response/StreamSplitter.java
+++ b/src/main/java/ru/yandex/clickhouse/response/StreamSplitter.java
@@ -108,7 +108,11 @@ public class StreamSplitter {
             posNext = 0;
         } else {
             byte[] oldBuf = buf;
-            buf = new byte[buf.length*2];
+            if (buf.length > (Integer.MAX_VALUE - 2) / 2) {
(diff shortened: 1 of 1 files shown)


zhicwu:
Thanks @zzsmdfj for pointing this out. It seems safer to go with `Integer.MAX_VALUE - 8` according to [this](https://github.com/openjdk/jdk16u/blob/master/src/java.base/share/classes/jdk/internal/util/ArraysSupport.java#L583).

Anyway, I got your idea and just merged the change into #596, which will be released in the coming days.

zzsmdfj:
> Thanks @zzsmdfj for pointing this out. It seems safer to go with `Integer.MAX_VALUE - 8` according to [this](https://github.com/openjdk/jdk16u/blob/master/src/java.base/share/classes/jdk/internal/util/ArraysSupport.java#L583).
> 
> Anyway, I got your idea and just merged the change into #596, which will be released in the coming days.

ok。

PR #609, opened 2021-03-30
Add suggestions for use (JitPack)

aiyanbo:


zhicwu:
Thanks @aiyanbo, I merged the change into develop branch. I'm going to make RoaringBitmap an optional dependency so that no one will have to suffer again.

PR #618, opened 2021-04-10
Fix error when use_new_parser is set to false

zhicwu:


zhicwu:
When `use_new_parser` is set to `false`, the driver should append `FORMAT TabSeparatedWithNamesAndTypes` to the given query before passing the query to Clickhouse.

PR #626, opened 2021-04-15
Shared cookie store

Cheppali-Reddy:
Allows to share same cookie store among http clients of same database. This feature allows users to use [AWS ALB ](https://docs.aws.amazon.com/elasticloadbalancing/latest/application/sticky-sessions.html) sticky sessions.

With the help of sticky sessions, we can direct all the queries from a given client to same CH node. We can achieve `read your writes` consistency at each client level (Client will always see the writes it made as we are hitting the same CH node - assuming other CH properties are appropriately configured).


zhicwu:
Thanks for the pull request @Cheppali-Reddy. Both `client_name` and `session_id` options can be used for routing, but it seems cookie is more user-friendly. Overall I'm good for the PR, and I'll merge over the weekend(hopefully after fixing the build break).

PR #633, opened 2021-04-20
Add more query parameters

zhicwu:
Add 3 more query parameters: `allow_experimental_bigint_types`, `allow_experimental_map_type`, and `join_algorithm`.


lphu:
> Add 3 more query parameters: `allow_experimental_bigint_types`, `allow_experimental_map_type`, and `join_algorithm`.

I met the scene that needs skip some error rows when inserting batch. I found `input_format_allow_errors_num` and `input_format_allow_errors_num` parameters in docs.  But clickhouse-client doesn't have these params.

Would you pls add `input_format_allow_errors_num` and `input_format_allow_error_ratio` query parameters, or if it can open api to let users dynamicly add query parameters?

Issue #648, opened 2021-04-30
After update to 0.3.1 datasource ignores database from url


sneakythr0ws:
I use `new ClickHouseDataSource(url, clickHouseProperties());` for datasource creation
After update to 0.3.1 it ignores database from url and uses `default`
Downgrading to 0.3.0 fixes the problem

zhicwu:
Thanks for the reporting issue and this has been fixed on develop branch.

sneakythr0ws:
Is it fixed in v0.3.1-patch?

zhicwu:
> @zhicwu Is it fixed in v0.3.1-patch?

Yes.

PR #651, opened 2021-05-06
Fix issue #648 introduced in 0.3.1

zhicwu:
* only ignore database parameter when given DDL contains `DATABASE` keyword
* change current database of Statement/PreparedStatement as needed
* add unit test to address regression issue


martenlindblad:
Thanks for fixing this issue. I think this should be a hotfix since it breaks our current setup. We can't use 0.3.1 and it seems like 0.3.2 still is far away from release.

zhicwu:
> Thanks for fixing this issue. I think this should be a hotfix since it breaks our current setup. We can't use 0.3.1 and it seems like 0.3.2 still is far away from release.

Sorry I've been lazy lately :p Just released `0.3.1-patch` for this and I'll expedite 0.3.2 release(for real ;).

PR #658, opened 2021-05-10
Return correct value from getAutoCommit

findepi:
Fixes https://github.com/ClickHouse/clickhouse-jdbc/issues/657


zhicwu:
Thank you @findepi for fixing the issue.

findepi:
Thanks for a quick merge!

PR #677, opened 2021-06-25
Fixed the content type in requests with external data

MikhailMel:
Added an explicit indication of the text/plain content type with UTF-8 encoding for an SQL query in MultipartEntityBuilder, otherwise the text/plain content type with ISO_8859_1 encoding is used and Latin letters turn into'??????'


zhicwu:
Thanks for your contribution @MikhailMel. The change has been merged.

PR #682, opened 2021-07-08
Fix timezone NPE when initializing the connection.

phoesh:
The timezone variable will be used in ClickHouseStatementImpl.updateResult method.


zhicwu:
Close as this has been fixed in #685.

PR #693, opened 2021-07-14
Replace rbm dependency to fix package problems

dengziming:
`com.github.RoaringBitmap:RoaringBitmap` and `com.github.RoaringBitmap.RoaringBitmap:RoaringBitmap` will generate the same jar file name "RoaringBitmap-0.9.9.jar" as below, and the latter is an empty jar.
image

Only one jar will remain in `target/lib` after `mvn compile` since their name are identical, sadly the empty jar replaced the other, so I got a `java.lang.NoClassDefFoundError: org/roaringbitmap/RoaringBitmap`.
image

I think we should use `com.github.RoaringBitmap.RoaringBitmap:RoaringBitmap` rather than `com.github.RoaringBitmap:RoaringBitmap` to avoid this problem.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -149,7 +149,7 @@
                 <version>${lz4.version}</version>
             </dependency>
             <dependency>
-                <groupId>org.roaringbitmap</groupId>
+                <groupId>com.github.RoaringBitmap.RoaringBitmap</groupId>
                 <artifactId>RoaringBitmap</artifactId>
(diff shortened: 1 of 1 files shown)


zhicwu:
The artifact was moved back to maven central under `org.roaringbitmap` group since 0.9.10(latest is 0.9.15). It seems you're not using latest code on master/develop branch.

dengziming:
@zhicwu Thank you, I will check that soon.

PR #697, opened 2021-07-27
Avoid timezone query on create connection when not needed

nicolasblaye:
This aims to prevent this query to append if we use the timezone specified in configuration.


den-crane:
But it skips to query version()
```
 if (serverVersion == null) {
            serverVersion = "";
        }
```

AFAIK it needs to implement lazy evaluation of the version?

nicolasblaye:
Looking at the code, it does need the version for one call. 
I will add the possibility to specify the version too.

`if (connection.getServerVersion().compareTo("1.1.54237") > 0) {
            query = new StringBuilder(
                "SELECT database, table, name, type, default_kind as default_type, default_expression, comment ");
        } else {
            query = new StringBuilder(
                "SELECT database, table, name, type, default_type, default_expression, NULL AS comment ");
        }`

nicolasblaye:
@den-crane Sorry I don't see which test fails. `mvn test` returns an unrelated NPE for me. How can I run them?

den-crane:
@nicolasblaye https://github.com/ClickHouse/clickhouse-jdbc/blob/master/clickhouse-jdbc/src/test/java/ru/yandex/clickhouse/integration/StreamSQLTest.java#L265

Some tests use getServerVersion() to make a decision does it need to run or not. Now they just skipped.

nicolasblaye:
@den-crane As far as I know, my last two commits should make so version is still fetched by default via a select. It shouldn't change the test no?

den-crane:
OK then.  I did not check the last commit.

zhicwu:
Thanks for the pull request @nicolasblaye. It's one-time query for connection initialization, with trivial overhead IMO. Would you mind to share the problem you're trying to resolve here? From what I saw, disabling the query may help for unit testing, plus it will slightly increase throughput.

By the way, I changed target branch to `develop`. Could you check the build break as well as merge conflict so that we can merge?

nicolasblaye:
It's true that it's a small query, but we misconfigured our connection creation so we spam this query way too much (over a million time a day). Obviously, I am also fixing our connection creation issue but I figured I could also add that to the driver to avoid unnecessary (albeit very small) requests.

zhicwu:
Won't merge this pull request as this has been addressed in new JDBC driver. When `serverTimeZone` and `serverVersion` are specified, the driver will no longer issue the query on creating connection.

PR #699, opened 2021-07-30
Change log level for driver registration

buremba:
It looks like the log level is `INFO` for the driver registration. If you look at most JDBC implementations, they use DEBUG in this exact case since the registration is expected when we load ClickhouseDriver in the runtime.

### DIFF
--- a/clickhouse-jdbc/src/main/java/ru/yandex/clickhouse/ClickHouseDriver.java
+++ b/clickhouse-jdbc/src/main/java/ru/yandex/clickhouse/ClickHouseDriver.java
@@ -46,7 +46,7 @@ public class ClickHouseDriver implements Driver {
         } catch (SQLException e) {
             throw new RuntimeException(e);
         }
-        logger.info("Driver registered");
+        logger.debug("Driver registered");
     }
 
     @Override


buremba:
@zhicwu any ideas on why the tests are failing?

zhicwu:
> @zhicwu any ideas on why the tests are failing?

I think it's unrelated. I have a large PR on hand, let me merge this one over the weekend.

buremba:
Thanks @zhicwu !

buremba:
Hey @zhicwu, I see that latest `patch` version doesn't include this PR. Is there any timeline on when can we publish this?

zhicwu:
> Hey @zhicwu, I see that latest `patch` version doesn't include this PR. Is there any timeline on when can we publish this?

It will be released together with 0.3.2. Sorry it's a complete rewrite (unfortunately), so I don't have an exact ETA - maybe a few more weeks?

As to this specific issue, perhaps you can update logging configuration by changing log level for `ru.yandex.clickhouse` to `ERROR`? FYI, starting from 0.3.2, transaction APIs like commit(), rollback() and setSavepoint() etc. will be supported, and there'll be annoying warnings keep saying it's fake transaction. Faking transaction support can be turned off for sure, but still I think changing log level to ERROR is a better solution.

buremba:
OK, thanks for letting me know @zhicwu !

PR #712, opened 2021-09-03
ConfigurableApi.addExternalData and withExternalData

bgranvea:
I added `addExternal/withExternalData` methods in `ConfigurableApi`, similar to `addDbParam `and `options`. I also made `ClickHouseConnection.getProperties()` public.

These two changes simplify the way I'm using Clickhouse JDBC driver in my project.

### DIFF
--- a/clickhouse-jdbc/src/main/java/ru/yandex/clickhouse/ClickHouseConnection.java
+++ b/clickhouse-jdbc/src/main/java/ru/yandex/clickhouse/ClickHouseConnection.java
@@ -1,5 +1,7 @@
 package ru.yandex.clickhouse;
 
+import ru.yandex.clickhouse.settings.ClickHouseProperties;
+
 import java.sql.Connection;
 import java.sql.SQLException;
 import java.util.TimeZone;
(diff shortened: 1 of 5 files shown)


bgranvea:
> Thanks for your contribution and sorry for getting back late to this pull request. I think we probably should stop adding more functions into extended API, as it will be replaced by `clickhouse-client`.
> 
> For this specific case, it's something like below in new client:
> 
> ```java
> try (ClickHouseClient client = ClickHouseClient.newInstance(ClickHouseProtocol.GRPC)) {
>     ClickHouseRequest<?> request = client.connect(server)
>         // :ids is a named parameter
>         .query("insert into table1 select * from myTable2 where id in :ids")
>         // no add method at this point
>         .external(myTable1, myTable2, myTable3)
>         .removeExternal(myTable1)
>         .removeExternal("myTable3");
>     try (ClickHouseResponse resp = request.params(Arrays.asList(1,2,3,4,5)).execute().get()) {
>         // ...
>     }
> }
> ```

this builder pattern for this new API seems nice, but my code is mostly JDBC. How is it going to work? 

will JDBC implementation use internally a ClickHouseRequest? in this case will it be possible to access it so I can do something like this:

`preparedStatement.unwrap(ClickHousePreparement.class).request().external(myTable1) `

Is this API planned for 0.4?

Thanks.

zhicwu:
The API is available in 0.3.2 and package ru.yandex.clickhouse will be droped in 0.4.0.

Issue #713, opened 2021-09-07
[Feature Request] Supporting JPMS


arjenzhou:
Since Java 9 introduces JPMS, libs without `module-info` or Automatic module declaration cannot be used after Java 9, is there any plan to support it?

arjenzhou:
related discussions: 
https://github.com/brettwooldridge/HikariCP/issues/1045
https://github.com/oshi/oshi/pull/1247

zhicwu:
Thank you @arjenzhou. Yes, we definitely need this.

zhicwu:
Looks like JPMS is still not widely adopted. `Automatic-Module-Name` is merely a workaround to make things work the old way. I tried multiple profiles(jre8 & jre11) in Maven, but it's difficult to release multiple jars in one go, not to mention we had to exclude different files in different profiles. As a result, I switched to multi-release jar ([JEP#238](https://openjdk.java.net/jeps/238), which is less flexible but a better way to manage source code and deployment. However, we still have to repackage some third party libraries like RoaringBitmap, gRPC and maybe more :<

zhicwu:
This has been available since 0.3.2-test1.

In addition to explicit declaration in each module, `Automatic-Module-Name` was used for shaded jar in case it's needed somewhere.

PR #714, opened 2021-09-07
Supporting JPMS

arjenzhou:
#713 ClickHouse JDBC support JPMS Automatic Module

### DIFF
--- a/clickhouse-jdbc/pom.xml
+++ b/clickhouse-jdbc/pom.xml
@@ -51,6 +51,7 @@
         <shade.base>ru.yandex.clickhouse.jdbc.internal</shade.base>
         <spec.title>JDBC</spec.title>
         <spec.version>4.2</spec.version>
+        <automatic.module.name>ru.yandex.clickhouse.jdbc</automatic.module.name>
     </properties>
 
     <dependencies>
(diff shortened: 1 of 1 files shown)


zhicwu:
Thanks @arjenzhou. Since now the project has multiple modules, we'd better provide `module-info.java` for each of them, with consideration of which exact packages should be exported.

In the upcoming `clickhouse-client`, I added two more profiles: one for jre8 and the other for jre11. We can add `module-info.java` in each module and exclude them only in jre8 profile.

arjenzhou:
> Thanks @arjenzhou. Since now the project has multiple modules, we'd better provide `module-info.java` for each of them, with consideration of which exact packages should be exported.
> 
> In the upcoming `clickhouse-client`, I added two more profiles: one for jre8 and the other for jre11. We can add `module-info.java` in each module and exclude them only in jre8 profile.

Excited to head that, when will `clickhouse-client` be released?

zhicwu:
> Excited to head that, when will `clickhouse-client` be released?

Should be 4 months ago :p I thought this could be a quick release, but it turned out many details to deal with. Now I'm still testing the grpc client and document known issues :< If I can submit the initial pull request this weekend, I guess 0.3.2 can be release in 2 weeks.

PR #732, opened 2021-10-06
Update build config for repackaging 3rd party libs

zhicwu:
* use consistent version for all repackaged 3rd party libs
* update group id to `com.clickhouse` instead of `tech.clickhouse`
* remove useless build steps
* add new job to release 3rd party libs


zhicwu:
Follow up PR of #731, merge with known issues.

PR #736, opened 2021-10-06
New Java client and gRPC support

zhicwu:
This pull request is mainly for adding experimental new Java client as well as gRPC support. Please be aware that APIs are subject to change without any notice, until 0.4.0.

__Key Changes__
- adaptive integration test and categorized test cases(under `unit` or `integration` group)
- JPMS support(incomplete) along with multi-release jar file and JDK 17 support
- use `com.clickhouse` as package name and maven group id
  Note: `clickhouse-jdbc` still uses `ru.yandex` as before, but it will be changed starting from 0.4.0.
- more features being added into `clickhouse-client` module:
  - async execution(by default)
  - better support of RowBinary format, and more data types like Date32, Geo types, and nested data types etc.
  - fluent API with builder classes
  - named parameter support
  - stream API via `ClickHouseResponse.recordStream()`
- experimental gRPC support in `clickhouse-grpc-client`
- renamed `clickhouse-native-client` to `clickhouse-tcp-client` (still empty there :p)

__Known Issues__ about the gRPC client
* query may hang if you're using an old version of ClickHouse - see ClickHouse/ClickHouse#27024
* compression is not supported well - see ClickHouse/ClickHouse#28671
  Note: you may want to disable compression for better performance.
* server reports "session is locked" when executing multiple queries in one go - might be a bug at client side

__TODOs__ - will address in separate PRs
- clickhouse-client
  - re-write `ClickHouseCluster`
  - sort out exceptions for better error handling
  - primitive array support
- clickhouse-grpc-client
  - custom registry for compression/decompression
  - ssl/tls support
  - support extremes
  - shaded jar with only netty(for building native image later)
- clickhouse-jdbc
  - change driver to `com.clickhouse.jdbc.ClickHouseDriver` but keep the old one as sub-class
  - refactor to build on top of `clickhouse-client`
  - switch to RowBinary format

### DIFF
--- a/clickhouse-grpc-client/src/test/java/com/clickhouse/client/grpc/ClickHouseGrpcClientTest.java
+++ b/clickhouse-grpc-client/src/test/java/com/clickhouse/client/grpc/ClickHouseGrpcClientTest.java
@@ -14,11 +14,10 @@ import java.nio.charset.StandardCharsets;
 import java.nio.file.Files;
 import java.nio.file.Path;
 import java.util.ArrayList;
-import java.util.Arrays;
 import java.util.Collections;
-import java.util.LinkedList;
 import java.util.List;
 import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.ExecutionException;
 import java.util.stream.Collectors;
 import org.testng.Assert;
 import org.testng.annotations.DataProvider;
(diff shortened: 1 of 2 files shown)


zhicwu:
JDBC driver stays the same(except changes for test cases) so it's not risky to merge this large PR. Besides, this will unblock future development work for 0.3.2 release.

Issue #531, opened 2020-12-27
ru.yandex.clickhouse.except.ClickHouseUnknownException: ClickHouse exception, code: 1002, host: xxx.xxx.xxx.xxx, port: 8123; xxx.xxx.xxx.xxx:8123 failed to respond


liujiehaobang:
ru.yandex.clickhouse.except.ClickHouseUnknownException: ClickHouse exception, code: 1002, host: xxx.xxx.xxx.xxx, port: 8123; xxx.xxx.xxx.xxx:8123 failed to respond
	at ru.yandex.clickhouse.except.ClickHouseExceptionSpecifier.getException(ClickHouseExceptionSpecifier.java:91)
	at ru.yandex.clickhouse.except.ClickHouseExceptionSpecifier.specify(ClickHouseExceptionSpecifier.java:55)
	at ru.yandex.clickhouse.except.ClickHouseExceptionSpecifier.specify(ClickHouseExceptionSpecifier.java:24)
	at ru.yandex.clickhouse.ClickHouseStatementImpl.getInputStream(ClickHouseStatementImpl.java:633)
	at ru.yandex.clickhouse.ClickHouseStatementImpl.executeQuery(ClickHouseStatementImpl.java:117)
	at ru.yandex.clickhouse.ClickHouseStatementImpl.executeQuery(ClickHouseStatementImpl.java:100)
	at ru.yandex.clickhouse.ClickHouseStatementImpl.executeQuery(ClickHouseStatementImpl.java:95)
	at ru.yandex.clickhouse.ClickHouseStatementImpl.executeQuery(ClickHouseStatementImpl.java:90)
Caused by: org.apache.http.NoHttpResponseException: xxx.xxx.xxx.xxx:8123 failed to respond
	at org.apache.http.impl.conn.DefaultHttpResponseParser.parseHead(DefaultHttpResponseParser.java:143)
	at org.apache.http.impl.conn.DefaultHttpResponseParser.parseHead(DefaultHttpResponseParser.java:57)
	at org.apache.http.impl.io.AbstractMessageParser.parse(AbstractMessageParser.java:261)
	at org.apache.http.impl.DefaultBHttpClientConnection.receiveResponseHeader(DefaultBHttpClientConnection.java:165)
	at org.apache.http.impl.conn.CPoolProxy.receiveResponseHeader(CPoolProxy.java:167)
[...]

Issue #290, opened 2019-01-14
HTTP keep alive management


bgranvea:
In my application I have random connection errors like this:

```
ClickHouseStatementImpl - Error during connection to ru.yandex.clickhouse.settings.ClickHouseProperties@785d9e37, reporting failure to data source, message: localhost:8123 failed to respond
org.apache.http.NoHttpResponseException: localhost:8123 failed to respond
        at org.apache.http.impl.conn.DefaultHttpResponseParser.parseHead(DefaultHttpResponseParser.java:143)
        at org.apache.http.impl.conn.DefaultHttpResponseParser.parseHead(DefaultHttpResponseParser.java:57)

```
(I've modified the code to print the stacktrace in ClickHouseStatementImpl)

I think the root cause is that in some rare cases, Apache Http client doesn't detect a connection closed by the server (half-closed socket). I have a Wireshark capture which confirms this.

I've seen that the server returns how long it will keep the connection open but it doesn't seem to be used client side:

```
POST /?compress=1&extremes=0&database=default HTTP/1.1
Content-Length: 55
Content-Type: text/plain; charset=UTF-8
Host: xxxxxx:8123
Connection: Keep-Alive
User-Agent: Apache-HttpClient/4.5.2 (Java/1.8.0_152)

select timezone() FORMAT TabSeparatedWithNamesAndTypes;HTTP/1.1 200 OK
Date: Mon, 14 Jan 2019 08:04:41 GMT
Connection: Keep-Alive
Content-Type: text/tab-separated-values; charset=UTF-8
X-ClickHouse-Server-Display-Name: xxxxxxxxx
Transfer-Encoding: chunked
Keep-Alive: timeout=3
(...)
```

and

```
private ConnectionKeepAliveStrategy createKeepAliveStrategy() {
        return new ConnectionKeepAliveStrategy() {
            @Override
            public long getKeepAliveDuration(HttpResponse httpResponse, HttpContext httpContext) {
                // in case of errors keep-alive not always works. close connection just in case
                if (httpResponse.getStatusLine().getStatusCode() != HttpURLConnection.HTTP_OK) {
                    return -1;
                }
                HeaderElementIterator it = new BasicHeaderElementIterator(
                        httpResponse.headerIterator(HTTP.CONN_DIRECTIVE));
                while (it.hasNext()) {
                    HeaderElement he = it.nextElement();
                    String param = he.getName();
                    //String value = he.getValue();
[...]

PR #760, opened 2021-11-24
Validate stale connection to fix the bug: failed to respond

ryan-tu:
# Problem Statement:
A lot of work has been done to fix the "failed to respond" problem.

#290, #462, #531 report the problem,  
#515 use the server side Keep-Alive timeout to replace the client side parameter,  to reduce the possibility of "stale connection", 
#505 fix the wrong connection reuse strategy, wrong Keep-Alive duration to make sure connection is closed immediately when a request fails,
#540  Add retry logic for idempotent SQL such as SELECT, and use a client side parser to decide whether an sql is idempotent or not. But,  for non-idempotent SQL, such as INSERT, this problem still exists. And parsing each SQL in JDBC is a heavy cost, and it's hard to keep consistent with server side SQL syntax.

# Proposal:
The root cause to "failed to respond" is half-closed tcp connections(closed by server-side when idle timeout reaches) are still kept in underlying connection pool of HttpClient. 
The problem is a common issue to all connection pools like Druid, C3P0, DBCP, HikariCP, and each connection pool MUST provide a validation mechanism to evict stale connection.

The validation mechanism provided by Apache HttpClient before v4.4 requires check per request, thus is too heavy  to use.
Since v4.4, a new validation mechanism is imported by: ```PoolingHttpClientConnectionManager.setValidateAfterInactivity()```, which is a middle ground between check per request and no check. 

This PR add a new connection setting `validateAfterInactivityMillis` to enable it.

The best solution should be that the client honors server-side Keep-Alive timeout to do best effort to reuse connections, and set `validateAfterInactivityMillis` to a slightly shorter value than server-side Keep-Alive timeout to evict stale ones.


zhicwu:
Thanks @ryan-tu - excellent work and great summary!

The default `validateAfterInactivity` is 2000ms, which is short enough right? See https://github.com/apache/httpcomponents-client/blob/a0b85d44c26430ff8354a0af5a1b7d58e2e5bce5/httpclient/src/main/java/org/apache/http/impl/conn/PoolingHttpClientConnectionManager.java#L183. Also, you may want to add a test for this - I was able to reproduce the issue by forcibly closing the connection from server-side(using Wiremock).

FYI, I rewrote JDBC driver in PR #747. Apache http client is now replaced by HttpURLConnection(similar as clickhouse4j), and it works just fine. JDK9+ HttpClient is apparently a better choice, but my point is that we probably don't need a 3rd party library like Apache http client and OKhttp anymore.

ryan-tu:
@zhicwu  

UT test cases are added. It takes some time to reproduce the bug and prove the bug has been fixed.

> FYI, I rewrote JDBC driver in PR #747. Apache http client is now replaced by HttpURLConnection(similar as clickhouse4j), and it works just fine. JDK9+ HttpClient is apparently a better choice, but my point is that we probably don't need a 3rd party library like Apache http client and OKhttp anymore.

I'm no expert of HTTP, but i am aware that Apache HttpClient provides a rich advanced feature set,  like proxy, pool configuration, async API, HTTP2 etc..  I'm not sure if HttpURLConnection can fulfill all the requirements.
I do agree JDK9+ HttpClient is a good choice.

zhicwu:
Thanks again @ryan-tu for spending extra time adding the tests. We probably need `CountDownLatch` to avoid flaky test. I will definitely play with the tests during weekend and update accordingly.

Did you notice `validateAfterInactivity` is set to 2000ms by default in `PoolingHttpClientConnectionManager`? I think it's short enough as default keep-alive timeout from server is 3000ms. This means we're not supposed to run into failed-to-response issue without any change in code, so it might also be a timing issue(e.g. thread pool Apache http client uses for eviction, and how long the clean up thread wait before execution etc.). Anyway, I have no problem of adding the new option, since it gives people more flexibility in terms of customization.
```java
public PoolingHttpClientConnectionManager(
    final HttpClientConnectionOperator httpClientConnectionOperator,
    final HttpConnectionFactory<HttpRoute, ManagedHttpClientConnection> connFactory,
    final long timeToLive, final TimeUnit timeUnit) {
    super();
    this.configData = new ConfigData();
    this.pool = new CPool(new InternalConnectionFactory(
            this.configData, connFactory), 2, 20, timeToLive, timeUnit);
    this.pool.setValidateAfterInactivity(2000);
    this.connectionOperator = Args.notNull(httpClientConnectionOperator, "HttpClientConnectionOperator");
    this.isShutDown = new AtomicBoolean(false);
}
```

Regarding http client, although HttpURLConnection does not support advanced feature like http trailer, it supports basic ones including chunking, making it suitable for a simple JDBC driver. Apache http client on the other hand is too heavy and it contains many features that we don't need. Having said that, because we now have JDBC driver and protocol implementation separated in different modules(e.g. `clickhouse-http-client` for handling http/https), we can easily add Apache http client back when needed for sure.

PR #769, opened 2021-11-30
Enhance new JDBC driver

zhicwu:
This is derived from PR #747 focusing on below items:
- [x] add more connection properties: `autoCommit`, `fetchSize`, `namedParameter`, and `wrapperObject`
- [x] basic JDBC escape syntax support
  * `{d '2021-11-01'}` -> `date'2021-11-01'`
  * `{t '12:34:56'}` -> `timestamp'1970-01-01 12:34:56'`
  * `{ts '2021-11-01 12:34:56'}` -> `timestamp'2021-11-01 12:34:56'`
  * `{ts '2021-11-01 12:34:56.78'}` -> `toDateTime64('2021-11-01 12:34:56.78', 2)`
  * `{tt 'my_temp\'_''table'}` -> ``` `my_temp'_'table` ```
      Note: `tt` is short for **t**emporary **t**able. It's a hint for using external table or creating a session-level temporary table.
- [ ] enhance prepared statement
  * multiple values for batch insertion
  * leverage JDBC escape syntax and external table for non-insert query(in a similar way as input function based approach)
- [ ] share test cases for http and grpc clients
- [ ] better support of AggregateFunction(and further optimization to bitmap)
- [x] implement unwrap methods to expose Java client API for JDBC users
- [x] use HttpClient instead of HttpURLConnection in JDK11 or above
  Note: noticed error `HTTP/1.1 header parser received no bytes` occasianly during stress test.
- [x] treat ASCII and Unicode string differently
- [ ] enhance multi-format support(grpc is inconsistent with http)
- [ ] fix reported issues in new JDBC driver


zhicwu:
Merge to create test build - will have another PR to follow up.

PR #770, opened 2021-11-30
3.1 jdbc driver broke float handling for nan/inf

grantatspothero:
When upgrading from the jdbc driver from 2.x to 3.x I noticed this following error occurring in tests (for the [trinodb ](https://github.com/trinodb/trino)project)

```
Caused by: ru.yandex.clickhouse.except.ClickHouseUnknownException: ClickHouse exception, message: Error parsing 'inf' as java.lang.Float; For input string: "inf"
	at ru.yandex.clickhouse.response.parser.ClickHouseValueParser$ClickHouseValueParserFunctionWrapper.parse(ClickHouseValueParser.java:243)
	at ru.yandex.clickhouse.response.ClickHouseResultSet.getObject(ClickHouseResultSet.java:749)
```

The root cause of the problem is `ResultSet.getFloat` and `ResultSet.getObject(Float.class)` use different parsing functions (a custom double parser for getFloat and Float::valueOf for getObject): https://github.com/ClickHouse/clickhouse-jdbc/blob/v0.3.1-patch/clickhouse-jdbc/src/main/java/ru/yandex/clickhouse/response/parser/ClickHouseValueParser.java#L43-L45

Clickhouse sends `[+|-]?inf` over the wire (not `[+|-]?Infinity`) so if you use `ResultSet.getObject(Float.class)` the `Float::valueOf` function errors out. 

The fix is just creating a custom float parser (much like the double parser) such that both float and double parsing perform the same logic.


clickhouse-jdbc/src/test/java/ru/yandex/clickhouse/response/ClickHouseResultSetTest.java
```diff
@@ -167,6 +167,7 @@ public void withTotals() throws Exception {
         assertEquals(70511139L, rs.getLong(2));
     }
 
+    @Test
```
grantatspothero, line 170: Nothing changed about this testcase, intellij just warned me it was missing a "@Test" annotation.

zhicwu:
Thank you @grantatspothero for the contribution. We're working on 0.3.2 which is a complete rewrite and we'll move to a RowBinary-based new JDBC driver, which supports nan/inf([examples](../blob/8a4aee2081607c032bdef671778e2e10e07e8026/clickhouse-client/src/test/java/com/clickhouse/client/data/BinaryStreamUtilsTest.java#L1604-L1661)).

Anyway, I changed target branch from `master` to `develop`, could you resolve the conflict there so that I can merge?

grantatspothero:
@zhicwu Just a bump

zhicwu:
> @zhicwu Just a bump

Sorry for the late response. Will merge this pull request after I'm done with another.

PR #774, opened 2021-12-09
Treat Nothing as NULL

zhicwu:
And add more tests for LZ4 input stream


yurifal:
Got it working! Thank you.

PR #776, opened 2021-12-09
add query params: max_ast_depth & max_expanded_ast_elements

f1yegor:
add query params: max_ast_depth & max_expanded_ast_elements

### DIFF
--- a/clickhouse-jdbc/src/main/java/ru/yandex/clickhouse/settings/ClickHouseProperties.java
+++ b/clickhouse-jdbc/src/main/java/ru/yandex/clickhouse/settings/ClickHouseProperties.java
@@ -83,6 +83,8 @@ public class ClickHouseProperties {
     private Long    preferredBlockSizeBytes;
     private Long    maxQuerySize;
     private Long    maxAstElements;
+    private Long    maxExpandedAstElements;
+    private Long    maxAstDepth;
     private boolean sessionCheck;
     private String  sessionId;
     private Long    sessionTimeout;
@@ -154,6 +156,8 @@ public class ClickHouseProperties {
         this.preferredBlockSizeBytes = getSetting(info, ClickHouseQueryParam.PREFERRED_BLOCK_SIZE_BYTES);
         this.maxQuerySize = getSetting(info, ClickHouseQueryParam.MAX_QUERY_SIZE);
         this.maxAstElements = getSetting(info, ClickHouseQueryParam.MAX_AST_ELEMENTS);
+        this.maxExpandedAstElements = getSetting(info, ClickHouseQueryParam.MAX_EXPANDED_AST_ELEMENTS);
+        this.maxAstDepth = getSetting(info, ClickHouseQueryParam.MAX_AST_DEPTH);
         this.sessionCheck = (Boolean) getSetting(info, ClickHouseQueryParam.SESSION_CHECK);
         this.sessionId = getSetting(info, ClickHouseQueryParam.SESSION_ID);
         this.sessionTimeout = getSetting(info, ClickHouseQueryParam.SESSION_TIMEOUT);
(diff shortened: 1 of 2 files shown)


zhicwu:
Hi @f1yegor, thanks for the pull request. The driver has been rewritten and we no longer have to do this. Instead, query parameters and headers can be customized via two new properties `custom_http_params` and `custom_http_headers`. For your case, the connection URL can be something like `jdbc:ch://localhost?custom_http_params=max_ast_depth%3D15%26max_expanded_ast_elements%3D500`.

f1yegor:
Thank you, @zhicwu, that's what I need.

zhicwu:
> Thank you, @zhicwu, that's what I need.

Please feel free to try the [new test build](../releases/tag/v0.3.2-test2) and see if it will work for you.

anouarhassine:
@zhicwu the connection url you provided doesn't work. I see 2 issues with it:
- I see a warning showing `- don't know how to handle parameter pair: custom_http_params=max_ast_depth=15`
- the parameter `max_expanded_ast_elements` is never taken into account.

What I noticed though is that the 1st parameter is always discarded and displayed in a warning. For example: using the following connection url `jdbc:ch://localhost?custom_http_params=foo%3D15%26bar%3D500` you will see a warning message saying `- don't know how to handle parameter pair: custom_http_params=foo=15`.

Am using _clickhouse-jdbc_ version **0.3.1**

Now how can the `max_expanded_ast_elements` be set ?

zhicwu:
> Am using _clickhouse-jdbc_ version **0.3.1**
> Now how can the `max_expanded_ast_elements` be set ?

It's not supported before 0.3.2, which is under public testing and will be released soon. You may try the test build as I mentioned above. The example I provided earlier was not entirely accurate, please refer to below:
`jdbc:ch://localhost/default?custom_http_params=max_ast_depth%3D15,max_expanded_ast_elements%3D500,session_id%3D123`

Two things you need to be aware of:
1. 0.3.2 is NOT a drop-in replacement since it's a complete rewrite - should be easy to migrate if you only use JDBC standard API
2. starting from 0.3.2, both http and grpc interfaces are supported, but above option only works for http, as its name implied

anouarhassine:
Thanks @zhicwu for the clarifications. I tried your new example with version **0.3.2-test2** and I get the following error:
`- don't know how to handle parameter pair: custom_http_params=max_ast_depth=15,max_expanded_ast_elements=500,session_id=123`

Looking at your code [here](https://github.com/ClickHouse/clickhouse-jdbc/blob/094ed0b9d2dd8a18ae0c7b3f8f22c35e595822a6/clickhouse-jdbc/src/main/java/ru/yandex/clickhouse/ClickhouseJdbcUrlParser.java#L81) I don't see how it can work. because the value has many `=` characters. I also don't see any unit test for such case.

In my opinion, `custom_http_params` does not work as passing the parameters like you describe is broken.

zhicwu:
> In my opinion, `custom_http_params` does not work as passing the parameters like you describe is broken.

The value should be URL encoded, like in the example.

anouarhassine:
Yes. I used the string you gave in the example and it gave that warning message. No parameter was taken into consideration.

zhicwu:
> Yes. I used the string you gave in the example and it gave that warning message. No parameter was taken into consideration.

I see what's going on there. Sorry this wasn't clear before - in 0.3.2, there're two JDBC drivers. Could you use `com.clickhouse.jdbc.ClickHouseDriver` instead? Starting from 0.4.0, everything under `ru.yandex` package will be removed.

anouarhassine:
Thanks! I can confirm that `com.clickhouse.jdbc.ClickHouseDriver` is the one that works.
When will the version 0.3.2 be released ?

PR #777, opened 2021-12-13
Prepare 0.3.2

zhicwu:
- [x] more tests for `ClickHouseInputStream`
- [x] use separate thread pool for I/O(instead of commonPool)
- [x] remove log4j dependency 
- [x] new options
  * clickhouse-http-client: `http_connection_provider` defaults to `HTTP_URL_CONNECTION`, can be changed to `HTTP_CLIENT` in JDK 11 or above
  * clickhouse-jdbc: `customTypeMappings` to define mapping between ClickHouse data type and Java class
- [x] correct behaviour of handling named parameters
  * `select :p1(DateTime64(3)) p1, :p2 p2, :p1 p3` has only two parameters: `p1` and `p2`
  * `select ? p1, ? p2, ? p3` has 3 parameters
- [x] optimize LZ4 input and output stream for memory effciency
- [x] add new data type `Bool`(instead of treating it as alias to `Int8`)
- [ ] enhance prepared statement for below two use cases
  * `insert into ... values`
  * `insert into ... format ...`
- [ ] support more AggregateFunction types


enqueue:
Bear in mind that a JDBC connection pool might create some dozens of connections ahead of time for reuse. If each of these connections spawns another threadpool, we might run into issues.

zhicwu:
> Bear in mind that a JDBC connection pool might create some dozens of connections ahead of time for reuse. If each of these connections spawns another threadpool, we might run into issues.

`max_threads_per_client` defaults to zero, but when it's configured with a number greater than zero, each client will have its own thread pool, instead of reusing `ClickHouseClient.defaultExecutor`. Maybe keep alive timeout should be set for core pool as well to avoid the situation of having too many idle threads.

PR #783, opened 2021-12-21
Prepare 0.3.2

zhicwu:
Finish open items in #777 and release `0.3.2`.
- [x] add warnings
  * legacy driver will be removed in 0.4.0
  * new driver does not support version older than 21.3
- [x] update docs(supported data types, compatibility matrix, and examples)
- [x] fix timezone issue(`use_time_zone` and `use_server_time_zone_for_date` option work now)
- [x] fix most parser issues(not case when and performance) - allow disable parsing in JDBC driver in 0.3.3
- [x] enhance prepared statement for below two use cases
  * `insert into ... values`
  * `insert into ... format ...`
- [x] throw exception for AggregateFunction types - needs to implement type system to deserialize/serialize different type of states in 0.3.3
- [x] update `clickhouse-benchmark` to test insertion


zhicwu:
The build failure is irrelevant, merge for 0.3.2 release.

PR #785, opened 2021-12-29
Merge 0.3.2

zhicwu:


zhicwu:
`clickhouse-native-client` is now `clickhouse-tcp-client`, because `native` is a reserved keyword in module-info.java :<

PR #797, opened 2022-01-06
Fix NullPointerException

linchen059:
NullPointerException happens when retrieve null DateTime from clickhouse.

clickhouse-jdbc version is :

<dependency>
    <groupId>com.clickhouse</groupId>
    <artifactId>clickhouse-jdbc</artifactId>
    <version>0.3.2</version>
</dependency>


zhicwu:
Thanks for fixing the issue. Actually `value == null` is not needed as getValue will always return a non-null garbage free value object. I'll update the code, add test and release a patch over on Sunday if nothing else shows up in these two days.

PR #804, opened 2022-01-13
Fix version check for system.tables.comment column in DatabaseMetaData.getTables

ebyhr:
`comment` column doesn't exist in `system.tables` when version < 21.6 as far as I confirmed.


zhicwu:
Thank you @ebyhr for fixing the issue. I'll look into this tonight and add unit tests to validate.

PR #805, opened 2022-01-13
add two params

OodounaikingoO:
add _input_format_allow_errors_num_ and _input_formati_allow_errors_ratio_ params.

When we insert some data using batch mode, we may lose all the data in a batch if some parse errors(like the data type is inconsistent) in some rows. Setting this two params we can skip the wrong rows and insert other correct data in the batch.


zhicwu:
Thanks @OodounaikingoO. I'd suggest you upgrading driver to `com.clickhouse.jdbc.ClickHouseDriver`, because `ru.yandex` package will be dropped in 0.4.0. Actually starting from 0.3.2, you can set custom query parameters by setting connection property `custom_http_params` to something like `input_format_allow_errors_num=1,input_formati_allow_errors_ratio=0`, and similar `custom_http_headers` for headers. Alternatively, you no longer has to use JDBC API and you can use `clickhouse-http-client` or `clickhouse-grpc-client` instead.

Issue #807, opened 2022-01-16
New Driver returns incorrect precision


tangjiangling:
- JDBC Standard(`java.sql.DatabaseMetaData#getColumns`):

> The COLUMN_SIZE column specifies the column size for the given column. For numeric data, this is the maximum precision. For character data, this is the length in characters. For datetime datatypes, this is the length in characters of the String representation (assuming the maximum allowed precision of the fractional seconds component). For binary data, this is the length in bytes. For the ROWID datatype, this is the length in bytes. Null is returned for data types where the column size is not applicable.

- To be deprecated Driver(`ru.yandex.clickhouse.ClickHouseDatabaseMetadata#getColumns`):
```
// column size / precision
row.add(String.valueOf(columnInfo.getPrecision()));
```
- New Driver(`com.clickhouse.jdbc.ClickHouseDatabaseMetaData#getColumns`):
```
r.getValue("COLUMN_SIZE").update(column.getDataType().getByteLength());
```

We can see that the new Driver goes against the JDBC standard and the Driver that was marked as deprecated is the right one.

PR #808, opened 2022-01-16
Fix wrong precision

tangjiangling:
Fixes: #807 

`com.clickhouse.jdbc.ClickHouseDatabaseMetaData#getColumns`
returns the wrong precision, which is against the JDBC standard.


tangjiangling:
A related issue(https://github.com/trinodb/trino/issues/10541)

tangjiangling:
cc: @ebyhr @hashhar

tangjiangling:
The second commit describes:
> In method `com.clickhouse.jdbc.ClickHouseDatabaseMetaData#getColumns`, 
we should map `DECIMAL_DIGITS` and `CHAR_OCTET_LENGTH` to `int32(0)`, 
otherwise it will lead to incorrect results.

Why?

The original implementation maps `DECIMAL_DIGITS` and `CHAR_OCTET_LENGTH` to `int32(null)`, which causes the corresponding field type to become `com.clickhouse.client.data.ClickHouseEmptyValue`, which is unable to update the new value, so it will end up with the wrong result.

zhicwu:
Thanks @tangjiangling for the pull request :) I don't mind to create another patch release, but I'll need to run integration test against trino to see if there's more issues.

tangjiangling:
> would you mind to sign CLA and rebase?

signed.

tangjiangling:
> would you mind to sign CLA and rebase?

Sorry, I forgot to set the email, I will do that soon.

tangjiangling:
> Sorry, I forgot to set the email, I will do that soon.

done.

zhicwu:
> Is it a combination of all the "Merge pull request ClickHouse#..." into a single commit?

Yes, squash may work too.

>  incorrect precision

I think we should consider both cases. When the column is numeric, we use `column.getPrecision()`, otherwise fall back to `column.getDataType().getByteLength()`. Make sense?

> changed `toInt32(null)` to `toInt32(0)`

I'll need to look into this tonight and add some more tests accordingly.

tangjiangling:
> I think we should consider both cases. When the column is numeric, we use column.getPrecision(), otherwise fall back to column.getDataType().getByteLength(). Make sense?

+1

tangjiangling:
> +1, I will update the current PR later this evening.

I fixed it on the basis of this https://github.com/ClickHouse/clickhouse-jdbc/pull/815 .

tangjiangling:
Fixed by #815. Now we can close this.

PR #811, opened 2022-01-18
Add  formatTimestamp64

liumaojing:
Add  formatTimestamp64


zhicwu:
> Add formatTimestamp64

Thanks for the pull request. It seems you're trying to address your issue but the solution is not generic. For example: DateTime64 supports nano seconds but you're using `yyyy-MM-dd HH:mm:ss.SSS`. Moreover, it's probably better to merge the two PRs into.

Lastly, I hope you understand that `ru.yandex` is deprecated and will be removed in 0.4.0. You probably should try the new JDBC driver `com.clickhouse.jdbc.ClickHouseDriver` which supports both time zone and DateTime64.

liumaojing:
Got it. I'll go down and look into it

PR #812, opened 2022-01-18
setTimestamp format timestamp64

liumaojing:
setTimestamp format timestamp64


zhicwu:
Thanks again @liumaojing. As I suggested in the other pull request, please use new JDBC driver instead and it should work. Also I want to mention that `setTimestamp(Timestamp)` and `getTimestamp()` are not recommended in JDBC, it's better to use `setObject(LocalDateTime)` and `getObject()` instead, not to mention the former one is slow in current implementation.
image

PR #844, opened 2022-02-22
List parser to [] when convert to sql expression

fandylua:
List parser to [] when convert to sql expression


zhicwu:
Thanks for the pull request. What's the case you're trying to resolve? Collection(including List) has been taken for Tuple - why not convert your list to Object[] so that it will be mapped to array in ClickHouse?

fandylua:
tks, I try new Driver with, today. I find `list` not parser to `[]` but `()`

PR #858, opened 2022-03-06
fix java codes_that fails compile in README.md s

XenoAmess:


zhicwu:
Thanks again @XenoAmess :)

PR #887, opened 2022-04-01
Fix key type when deserializing map objects

jnd77:
Fix typo about key type


zhicwu:
Good catch @jnd77 :)  Merge as the build failures are irrelevant. I'll add tests for regression and fix build break in a separate PR.

Issue #906, opened 2022-04-25
Compile failed by testInsertDefaultValue failed


ruanwenjun:
When I follow with the [documentation](https://github.com/ClickHouse/clickhouse-jdbc#build-with-maven) and execute `mvn clean verify`  to compile the project.
I get a error
```java
2022-04-25 20:06:24:309 +0800 [main] DEBUG ClickHouseConnectionImpl - Target node: ClickHouseNode(addr=http:localhost:54851, db=test_statement)@1698261139
[ERROR] Tests run: 280, Failures: 2, Errors: 0, Skipped: 0, Time elapsed: 26.137 s <<< FAILURE! - in TestSuite
[ERROR] ru.yandex.clickhouse.BalancedClickhouseDataSourceTest.testIPv6  Time elapsed: 0.009 s  <<< FAILURE!
java.lang.AssertionError: expected [true] but found [false]
        at ru.yandex.clickhouse.BalancedClickhouseDataSourceTest.testIPv6(BalancedClickhouseDataSourceTest.java:328)

[ERROR] com.clickhouse.jdbc.ClickHousePreparedStatementTest.testInsertDefaultValue[DateTime64(3, 'UTC'), '1969-12-31 23:59:59.123', 1969-12-31 23:59:59.123](5)  Time elapsed: 0.051 s  <<< FAILURE!
java.lang.AssertionError: expected [1969-12-31 23:59:59.123] but found [1970-01-01 00:00:00.877]
        at com.clickhouse.jdbc.ClickHousePreparedStatementTest.testInsertDefaultValue(ClickHousePreparedStatementTest.java:1041)

[INFO] 
[INFO] Results:
[INFO] 
[ERROR] Failures: 
[ERROR]   ClickHousePreparedStatementTest.testInsertDefaultValue:1041 expected [1969-12-31 23:59:59.123] but found [1970-01-01 00:00:00.877]
[ERROR]   BalancedClickhouseDataSourceTest.testIPv6:328 expected [true] but found [false]
[INFO] 
[ERROR] Tests run: 280, Failures: 2, Errors: 0, Skipped: 0

```
My jdk version is 1.8.

It seems we insert the time is before 1970.

BTW, how to import the latest master branch into IDEA, I find we have the grpc code in repo, and when I execute the test case, it will throw below exception.
```java
[...]

PR #917, opened 2022-05-05
Tsv totals extremes 2

radmirnovii:
Fix for ClickHouseTabSeparatedProcessor with extremes and/or totals.

### DIFF
--- a/clickhouse-client/src/main/java/com/clickhouse/client/ClickHouseDataProcessor.java
+++ b/clickhouse-client/src/main/java/com/clickhouse/client/ClickHouseDataProcessor.java
@@ -105,6 +105,9 @@ public abstract class ClickHouseDataProcessor {
     protected final Iterator<ClickHouseValue> values;
     // protected final Object writer;
 
+    protected static ClickHouseRecord totals;
+    protected static ClickHouseRecord[] extremes;
+
     /**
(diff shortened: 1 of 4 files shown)


zhicwu:
Thanks for your contribution @radmirnovii. That was indeed a missing feature. Will you also consider RowBinary processor?

By the way, the build break is irrelevant. It's caused by ClickHouse server, which was reported in #906. I'll enhance corresponding integration test as a workaround.

zhicwu:
> Is it possible without parsing query?

Maybe not ;) Would be nice if we can support both formats.

Anyway, do you have time for additional changes? If not, I'll follow up to add test along with a few more changes to get it merged.

radmirnovii:
> Anyway, do you have time for additional changes? 
There is no empty spaces or any special values. Cannot find solution without query parsing. 
But i can add support for grpc version.

zhicwu (changes requested):

zhicwu:
Sorry I think it's better to revisit this in 0.3.3 release, mainly for 3 reasons:
1. it's a bit tricky for user to call the new methods to get totals and extremes - they have to know how many rows returned from the server first, and then call them in sequence right after reading the last row
2. we'd better leverage progress stats for a more generic implementation for all protocols and formats
3. unit tests and integration tests to prove the function works well

PR #921, opened 2022-05-07
Avoid losing precision when converting Float and Double to BigDecimal

### LINKED ISSUE #913: The 'getBigDecimal' method has a precision of 0 by default
When using a 'getBigDecimal' query, the 'ClickHouse DoubleValue' is returned, but when converting to 'BigDecimal', use 'scale=0' directly, which will result in a loss of precision. 
Although there is such a way as 'getBigDecimal(v, scale)', the current JDBC specification should be deprecated, and the method 'getBigDecimal(v)' is normally used. I think it would be better to use 'BigDecimal.valueOf(v)' to maintain the correct precision.
image

zhicwu:
Closes #913

### DIFF
--- a/clickhouse-client/src/main/java/com/clickhouse/client/ClickHouseEnum.java
+++ b/clickhouse-client/src/main/java/com/clickhouse/client/ClickHouseEnum.java
@@ -1,10 +1,13 @@
 package com.clickhouse.client;
 
+import java.io.Serializable;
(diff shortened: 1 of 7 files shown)


zhicwu:
Build failure was caused by ClickHouse/ClickHouse#36994, which has nothing to do with the change.

PR #935, opened 2022-05-11
Add null check for batch insert

### LINKED ISSUE #931: Treating the  Empty String as NULL since  0.3.2-patch6
I can reproduce this misbehaviour since version 0.3.2-patch6.

Up to version 0.3.2-patch5 everything worked fine.

I poked around an bit and found out that the jdbc-driver has problems with empty strings.
It seems that it handles empty strings like null values.

Before patch6 everything worked fine. This behavior still exists in patch9

This  relates to #896
`Unexpected error inserting/updating row in part [insertRow add batch]`
...
`com.clickhouse.client.data.BinaryStreamUtils.writeString (BinaryStreamUtils.java:1667)`

If is replace the empty string value with any character the error is gone.

zhicwu:
Closes #931

### DIFF
--- a/.github/workflows/third_party_libs.yml
+++ b/.github/workflows/third_party_libs.yml
@@ -6,7 +6,7 @@ on:
       version:
         description: "Release version"
         required: true
-        default: "1.0.0"
+        default: "1.2.0"
 
 jobs:
   release:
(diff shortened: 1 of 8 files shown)


zhicwu:
Will be addressed together with #956.

PR #936, opened 2022-05-12
update grpc to 1.45.1

n-oden:


zhicwu:
Thanks @n-oden. I was waiting for Okhttp 3/4 support, so that we can add another implementation in `clickhouse-http-client` without outdated dependencies :) Could you also update [third-party-libraries](https://github.com/ClickHouse/clickhouse-jdbc/tree/develop/third-party-libraries) since grpc-java still does not support JPMS(even 1.46)?

By the way, is your work depending on the new version? FYI, it's probably not worthy of using gRPC with ClickHouse at this point due to the performance issue mentioned at [here](https://github.com/ClickHouse/ClickHouse/issues/28671#issuecomment-1087049993).

zhicwu:
I see. But why not use shaded jar instead?

Anyway, protobuf should be updated as well or the code generated by protoc won't work. I don't have privilege to update your branch so I'll fix that in a separate PR.

PR #948, opened 2022-06-02
Support for custom url path

radmirnovii:
Motivation:
Useful with http proxy. Previous version in `ru.yandex.clickhouse` has this feature with option USE_PATH_AS_DB=false.

current driver has WEB_CONTEXT option but it used only with trailing slash

### DIFF
--- a/clickhouse-http-client/src/main/java/com/clickhouse/client/http/ClickHouseHttpConnection.java
+++ b/clickhouse-http-client/src/main/java/com/clickhouse/client/http/ClickHouseHttpConnection.java
@@ -125,21 +125,8 @@ public abstract class ClickHouseHttpConnection implements AutoCloseable {
         StringBuilder builder = new StringBuilder();
         builder.append(config.isSsl() ? "https" : "http").append("://").append(server.getHost()).append(':')
                 .append(server.getPort()).append('/');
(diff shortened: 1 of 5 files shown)


zhicwu:
Thank you @radmirnovii. Can we reuse `web_context`? I'm testing connection string with multiple hosts/URIs. Perhaps we can enhance the option to support both cases.

zhicwu:
Thank you @radmirnovii. Can we reuse `web_context`? I'm testing connection string with multiple hosts/URIs. Perhaps we can enhance the option to support both cases.

radmirnovii:
> Can we reuse `web_context`? 
Do you want to just remove adding trailing slash in path?

zhicwu:
> Do you want to just remove adding trailing slash in path?

Yes. Web context should be also considered for health check like `ping`, because the url will be changed to `http://host:port/<web_context>/ping`. What's your scenario? I didn't see unit test there, so my guess was just arbitrary path without query or fragment, right?

Apologize for holding your PR for so long, I'll merge the 2 PRs with minor changes maybe, once I got load balancing performance issue fixed :<

zhicwu:
Fixed in #956 along with some more tests.

PR #952, opened 2022-06-14
Update StreamSplitter.java

liumaojing:
The result is truncated.


zhicwu:
Thanks @liumaojing. What's the issue you're trying to resolve? Can you add a test? Also looking at the code, I'm not sure when `marking` will be set to true?

liumaojing:
> 谢谢@liumaojing. 您要解决的问题是什么？可以加个测试吗？也在看代码，我不确定什么时候`marking`会设置为true？

The column in the query has a long column of data around 65535 bytes,
public ByteFragment[] split(byte sep) {
        StreamSplitter ss = new StreamSplitter(this, sep);  debug to this is wrong, shard data error

mzitnik:
@liumaojing thanks for contribution as @zhicwu asked can you provide a test for this issue and run locally to verify that all tests are passing and than we could merge

Issue #920, opened 2022-05-06
Multiple values insert prepared statement


illyaSlobozhanin:
[QUESTION] about multiple values in prepared statement. 
Probably, after this [COMMIT](https://github.com/ClickHouse/clickhouse-jdbc/commit/09cd66e564966d76d72f996754dbf8311bf1b49d)
between **patch5** and **patch6** and later versions.

I can't anymore use multiple values in prepared statement, like
`PreparedStatement stmt = Connection#prepareStatement("insert into table values(?,?), (?,?)");`
only 
`PreparedStatement stmt = Connection#prepareStatement("insert into table values(?,?)");`
and then insert values for each row separated by 
`stmt.addBatch()`

Is it final change or, maybe some options are needed to be turned on? 

Under the hood detail.
`values (?,?), (?,?)` is not properly parsed and it doesn't see `ClickHouseSqlStatement.KEYWORD_VALUES_START` (values start position index in sql).
Here in the [CODE](https://github.com/ClickHouse/clickhouse-jdbc/blob/1d6257c51a1d5db0a4732441fc3c1c7dba7b5316/clickhouse-jdbc/src/main/java/com/clickhouse/jdbc/internal/ClickHouseConnectionImpl.java#L601) returns null which cause NPE later in code.
For `values (?,?)` all worked correctly.

zhicwu:
Hi @illyaSlobozhanin, this sounds like a bug. However, why use multiple values expression instead of batch update? In general, it is recommended to use batch update as it performs better.

illyaSlobozhanin:
Hi, @zhicwu thanks for you reply. Unfortunately, we use mybatis and using its API we can only generate sql queries with all insert values It doesn\`t support proper batch insert using `addBatch()`. But probably we'll rewrite code into plain jdbc.

zhicwu:
Sorry for the inconvenience. I'll fix the issue in next release.

PR #957, opened 2022-06-20
New driver fails on insert with multiple values

reimai:
Hi, I've been updating my project to the latest jdbc driver and found that it crashes with an npe on an insert with [multiple values](https://clickhouse.com/docs/en/sql-reference/statements/insert-into/). The old one works fine. Could you please look into this? Here is a test for both drivers.


zhicwu:
Thank you @reimai. This was reported in #920 and it's been fixed in #940.

PR #985, opened 2022-07-06
Fixes number cast exception for ClickHouseArrayValue with Boolean[]

subkanthi:
Fixes number cast exception for ClickHouseArrayValue with Boolean[]


zhicwu:
Hi @subkanthi, thanks for your contribution and sorry for getting back late on this. I made a few more changes:
* change target branch to `develop`
* update `update(Object[])` method by moving the type check outside of the loop for performance reason, and it also means we no longer support corner case like object array with mixed types
* update `asArray(Class)` to convert byte[] to Boolean[]
* some more tests

Let me know if you're good with above changes and/or have more to add. I'll create a separate PR to enhance `asObject(Class)` method to support converting byte[] to boolean[] etc.

P.S. the build failure is irrelevant to this change.

Issue #991, opened 2022-07-11
addBatch write timed out


imyoubo:
![image](https://user-images.githubusercontent.com/41062250/178268375-9c817536-1e90-49ce-b9b2-620d3e5eaf1a.png)
![image](https://user-images.githubusercontent.com/41062250/178269793-eb9a19e8-a506-46c6-922f-d6e96e6ed3db.png)

I batch write clickhouse through jdbc client.
But I ran into a problem while writing it.
When addBatch is executed about 2000 times, it is blocked, stuck there and cannot continue, and after exceeding the set socket_timeout, an exception of Write timed out is reported.
It's weird that the same code ru.yandex.clickhouse.ClickHouseDriver doesn't have this problem

The version is 0.3.2-patch10
Please help with this issue, thanks！

我通过jdbc client 批量写clickhouse。
但是在写的时候遇到了一个问题。
当addBatch 执行了大约2000次左右时，就阻塞住了，卡在那里无法继续，然后超过设置的socket_timeout后，报Write timed out 异常了。
很奇怪，同样的代码 ru.yandex.clickhouse.ClickHouseDriver 没有出现这个问题

版本是 0.3.2-patch10
请帮忙看一下这个问题，谢谢！

imyoubo:
My program is executeBatch every 25w records
But every time after executing addBatch about 2000 times, it blocks, let alone executeBatch

我的程序是每25w条记录执行一次executeBatch
但是每次都是执行了2000次左右addBatch后，它就阻塞了，更没有执行过 executeBatch

zhicwu:
Hi @imyoubo, could you change `max_queued_buffers` to zero(unlimited) and try again? For example: `jdbc:ch://localhost?max_queued_buffers=0`.

zhicwu:
I'll remove the size limit in JDBC driver, so that nobody has to set the parameter explicitly.

imyoubo:
> Hi @imyoubo, could you change `max_queued_buffers` to zero(unlimited) and try again? For example: `jdbc:ch://localhost?max_queued_buffers=0`.

oh I tried it just now, and it really solved it. It seems to be the configuration problem of max_queued_buffers, thank you very much

imyoubo:
Hi @zhicwu  There seems to be a new problem, When executing executeBatch, an error is reported, "Please call addBatch method at least once before batch execution"
[...]

PR #996, opened 2022-07-14
Fix wrong setting usage for creating output stream

exploder86:
Value PERFORMANCE of settings "request_buffering" is ignored during output stream creation


exploder86:
I found this bug when debugging clickhouse-jdbc, it hangs  when executing big batch updates with spring jdbcTemplate. Batch size consists from 100 000 items and if buffer is small then main thread will loop inside NonBlockingPipedOutputStream#updateBuffer because CapacityPolicy also exceeded it's limits and return false.

zhicwu:
> I found this bug when debugging clickhouse-jdbc, it hangs when executing big batch updates with spring jdbcTemplate. Batch size consists from 100 000 items and if buffer is small then main thread will loop inside NonBlockingPipedOutputStream#updateBuffer because CapacityPolicy also exceeded it's limits and return false.

Thank you @exploder86, I think it's similar as #991.

PR #1009, opened 2022-07-25
Use BND for generating of Use bnd-jar-plugin to generate a proper

laeubi:
This is a PR to address the issues mentioned in https://github.com/ClickHouse/clickhouse-jdbc/issues/905

1. Use bnd-jar-plugin to generate a proper Manifest.MF file that is needed for OSGI-Bundles(jars) (**from** @stbischof)
2. Implement the DatasourceFactory (**tbd**)


laeubi:
@zhicwu Is master not the correct branch to use here? It seems the develop has some conflicts now can you explain what master vs develop is? master for lats release?

zhicwu:
> @zhicwu Is master not the correct branch to use here? It seems the develop has some conflicts now can you explain what master vs develop is? master for lats release?

Hi @laeubi, sorry this is a bit confusing. As of now, `master` is for released code, while `develop` is for new and experimental stuff. In the near future, once we fully retire the legacy driver, `develop` will be dropped and `master` will be renamed to `main`.

Could you rebase your code to latest on develop? And I'm sorry that there'll be more changes in these days for transaction support, which may impact your work.

zhicwu:
@laeubi, can we keep multi-release jar and JDK 8 support? As I mentioned in the original thread, multi-release jar is required for more features like ByteUtils using Vector API(requiring JDK 16+). Have you considered to start with a new module just for OSGi bundles?

laeubi:
@zhicwu I'll check this, I just picked up @stbischof work and got the feeling that multi-release *might* be dropped, but will check if we can keep this.

stbischof:
Wouldn't it be a better idea to use multiple service-jars that provide a service via ServiceLoader that can be consumes instead of your multi release approach?

zhicwu:
> Wouldn't it be a better idea to use multiple service-jars that provide a service via ServiceLoader that can be consumed instead of your multi release approach?

Thanks for the suggestion. Actually I thought about that in the beginning but we're talking about multiple JDKs here. They're actually same service but using different version of JDK. We can create multiple packages one for each JDK but given the fact that we have many packages already, are we going to multiply that by 2 or 3?

Is it impossible to generate OSGi bundles together with multi-release jar? Does it work if we create a separate module just for OSGi?

laeubi:
OSGi support multi-release: https://blog.osgi.org/2018/02/osgi-r7-highlights-java-9-support.html
Having an extra module sounds a bit superfluous, as the idea is to *reuse* whats there instead of adding much more on-top.

It is more that BND (a tool to automatically generate the header data) do not yet support multi-release jars but I'm trying to get this running atm.

PR #1010, opened 2022-07-25
Add column name to exception message in addBatch() 

subkanthi:
Add column name to exception message in addBatch()


zhicwu:
Thanks @subkanthi for the pull request. The build failure was probably due to code out of sync. Any I got your point and I can fix it in a separate PR.

PR #1011, opened 2022-07-26
Fix Roaring64NavigableMap reading mismatches

wzqiang1332:
https://github.com/ClickHouse/clickhouse-jdbc/issues/918#issue-1227574374

PR #1031, opened 2022-08-03
https://github.com/ClickHouse/clickhouse-jdbc/issues/874

panda0120:
the sort of the byte array of the Integer number is the difference between java and c


zhicwu:
Thanks for your contribution @panda0120. So PR #1011 didn't resolve all the issues. Do we need to bump dependency as well?

panda0120:
> Thanks for your contribution @panda0120. So PR #1011 didn't resolve all the issues. Do we need to bump dependency as well?

I'm ready to create an another pr to https://github.com/RoaringBitmap/RoaringBitmap.git,.if they received, this pr can be canceled.

PR #1040, opened 2022-08-13
Fix overflow error in ClickHouseLongValue.update method for BigDecimal and BigInteger

Blackmorse:
Calling method `java.sql.PreparedStatement#setBigDecimal` for column with type `UInt64` with the value greater than `Long.MAX_VALUE` results in _Overflow_ error


zhicwu:
Thanks for fixing the issue. Actually both signed and unsigned cases should be considered. Will revisit this when introducing byte code engineering. With that, we can add more checks without slowing things down.

PR #1055, opened 2022-08-29
Enable nightly build

zhicwu:


zhicwu:
Merge build script regardless build status.

PR #1070, opened 2022-09-07
Respect custom Authorization header in clickhouse-http-client

### LINKED ISSUE #1069: BasicAuthentication Support
Is basic authentication supported in the most recent version of the clickhouse JDBC drivers? (`clickhouse-jdbc-0.3.2-patch11` is my version).

I found an older PR saying that everyone got switched over https://github.com/ClickHouse/clickhouse-jdbc/issues/522, but looks like it may not be true in the most revent version.

I tried using the `custom_http_headers` property to set my own auth header, but the client complains about double auth because https://github.com/ClickHouse/clickhouse-jdbc/blob/f285b73d45597e29cf36dff4daf4cb4f9f53453a/clickhouse-http-client/src/main/java/com/clickhouse/client/http/ClickHouseHttpConnection.java#L166 this method is still always attempting to add the `X-Clickhouse` user and password methods. I can't find anywhere in the codebase that ever attempts to set an accessToken based method.

Am I missing a way to properly configure the client to use a `BasicAuth` header?

zhicwu:
Fixes #1069

### DIFF
--- a/clickhouse-client/src/test/java/com/clickhouse/client/BaseIntegrationTest.java
+++ b/clickhouse-client/src/test/java/com/clickhouse/client/BaseIntegrationTest.java
@@ -19,6 +19,10 @@ public abstract class BaseIntegrationTest {
         return ClickHouseServerForTest.getClickHouseNode(protocol, true, ClickHouseNode.builder().build());
     }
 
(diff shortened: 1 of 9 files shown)


zhicwu:
Merge as the build failure is irrelevant.

PR #1074, opened 2022-09-09
Offload ID generation to ClickHouseRequestManager

zhicwu:


zhicwu:
Will create a separate PR to optimize clickhouse-cli-client

PR #1087, opened 2022-09-14
Optimize reads for array values

zhicwu:
Deserialization of primitive arrays is now twice faster than before, and it also benefits the upcoming Native data format.

```
Benchmark                              (client) (connection)  (statement)  (type)   Mode  Cnt  Score   Error  Units
Query.selectArrayOfUInt16  v0.3.3-SNAPSHOT             reuse     prepared  object  thrpt   20  3.066 ? 0.208  ops/s
Query.selectArrayOfUInt16  v0.3.2-patch11              reuse     prepared  object  thrpt   20  1.380 ? 0.079  ops/s
Query.selectArrayOfUInt16  clickhouse-native-jdbc      reuse     prepared  object  thrpt   20  0.720 ? 0.102  ops/s
```

```
Benchmark                          (client)    (connection)  (statement)  (type)   Mode  Cnt    Score   Error  Units
Query.selectInt8    v0.3.3-SNAPSHOT(Native)           reuse     prepared  object  thrpt   20  130.568 ? 7.936  ops/s
Query.selectUInt64  v0.3.3-SNAPSHOT(Native)           reuse     prepared  object  thrpt   20   66.351 ? 4.128  ops/s
Query.selectInt8    v0.3.3-SNAPSHOT(RowBinary)        reuse     prepared  object  thrpt   20   85.267 ? 3.679  ops/s
Query.selectUInt64  v0.3.3-SNAPSHOT(RowBinary)        reuse     prepared  object  thrpt   20   46.181 ? 3.380  ops/s
Query.selectInt8    v0.3.2-patch11(RowBinary)         reuse     prepared  object  thrpt   20   85.092 ? 6.157  ops/s
Query.selectUInt64  v0.3.2-patch11(RowBinary)         reuse     prepared  object  thrpt   20   49.179 ? 3.010  ops/s
```

PR #1104, opened 2022-10-09
Bug fix: inserting Array of nullable primitive

Blackmorse:
Description:
Unable to insert the data into array of nullables

E.g.
```sql
CREATE TABLE default.t
(
    `a` Int8,
    `arr` Array(Nullable(Int8))
)
ENGINE = MergeTree
ORDER BY a
```

Trying to push an array(Scala code):
```scala
    val driver = new ClickHouseDriver()
    val connection = driver.connect("jdbc:clickhouse://localhost:8123", new Properties())
    val statement = connection.prepareStatement("insert into default.t (a, arr) values (?, ?)")

    val a = Array(1.toByte, 2.toByte, null).map(_.asInstanceOf[AnyRef])
    val arr = connection.createArrayOf("Array(Nullable(Int8))", a)

    statement.setByte(1, 1.toByte)
    statement.setArray(2, arr)
    statement.addBatch()
    statement.executeBatch()

    connection.close()
```
Error:
```
class [Ljava.lang.Object; cannot be cast to class [B ([Ljava.lang.Object; and [B are in module java.base of loader 'bootstrap')
java.lang.ClassCastException: class [Ljava.lang.Object; cannot be cast to class [B ([Ljava.lang.Object; and [B are in module java.base of loader 'bootstrap')
	at com.clickhouse.client.data.ClickHouseRowBinaryProcessor$MappedFunctions.writeArray(ClickHouseRowBinaryProcessor.java:57)
	at com.clickhouse.client.data.ClickHouseRowBinaryProcessor$MappedFunctions.serialize(ClickHouseRowBinaryProcessor.java:486)
	at com.clickhouse.jdbc.internal.InputBasedPreparedStatement.addBatch(InputBasedPreparedStatement.java:333)
	at com.blackmorse.spark.clickhouse.types.Int8Tests.$anonfun$new$4(Int8Tests.scala:37)
...
```

Reason:
 when using (e.g) `Array(Nullable(Int8))` type, `ClickHouseArrayValue` but `ClickHouseRowBinaryProcessor` still expects `ClickHouseByteArrayValue`.

### DIFF
--- a/clickhouse-client/src/main/java/com/clickhouse/client/data/ClickHouseRowBinaryProcessor.java
+++ b/clickhouse-client/src/main/java/com/clickhouse/client/data/ClickHouseRowBinaryProcessor.java
@@ -43,7 +43,7 @@ public class ClickHouseRowBinaryProcessor extends ClickHouseDataProcessor {
             ClickHouseColumn baseColumn = column.getArrayBaseColumn();
             int level = column.getArrayNestedLevel();
             Class<?> javaClass = baseColumn.getPrimitiveClass();
-            if (level > 1 || !javaClass.isPrimitive()) {
+            if (level > 1 || !javaClass.isPrimitive() || baseColumn.isNullable()) {
                 Object[] array = value.asArray();
                 ClickHouseValue v = ClickHouseValues.newValue(config, nestedColumn);
                 int length = array.length;
--- a/clickhouse-client/src/test/java/com/clickhouse/client/data/ClickHouseRowBinaryProcessorTest.java
+++ b/clickhouse-client/src/test/java/com/clickhouse/client/data/ClickHouseRowBinaryProcessorTest.java
@@ -129,6 +129,14 @@ public class ClickHouseRowBinaryProcessorTest {
         out.flush();
         Assert.assertEquals(bas.toByteArray(), BinaryStreamUtilsTest.generateBytes(2, 1, 2));
 
+        value = ClickHouseArrayValue.of(new Short[]{1, 2, null});
+        bas = new ByteArrayOutputStream();
+        out = ClickHouseOutputStream.of(bas);
+        ClickHouseRowBinaryProcessor.getMappedFunctions().serialize(value, config,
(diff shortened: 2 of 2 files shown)


zhicwu:
Thank you @Blackmorse! Actually this has been covered in my local changes.

Basically, widen unsigned types is probably not the best way, because it consumes more memory and has side effects in type conversion. I added `UnsignedLong`, `UnsignedInteger`, `UnsignedShort`, and `UnsignedByte` classes to better handle unsigned primitive types along with some changes in *Value classes, and more tests to ensure they work as expected. A few options were added too in case people prefer the old way.

I've completed the read part last night, and I should be able to complete the write part in a day or two.

Blackmorse:
Hi, @zhicwu 
Do you have any idea when you can apply you local code changes mentioned before? This issue is affecting my software, so maybe it's worth merging it now and not to wait for huge changes from your side?
Thanks

zhicwu:
> Hi, @zhicwu Do you have any idea when you can apply you local code changes mentioned before? This issue is affecting my software, so maybe it's worth merging it now and not to wait for huge changes from your side? Thanks

Hi @Blackmorse, I'm sorry for blocking your work. I must confess that I didn't spend any time on this project in the past weeks due to work on hand.

Are you expecting a patch release in maven central? As you probably noticed, there are some more changes in `develop` branch and the legacy driver got removed. So if all you need is the latest release plus this PR, I'd suggest you to build and push new artifact to local maven repository.

If it's fine to use nightly build(v0.3.3-SNAPSHOT) and wait for a few more days, I can commit some changes in local in a day or two, which will surely fix the issue along with performance improvement(in addition to PR #1087). However, this involves breaking changes to reduce serialization cost - if you're using interfaces ClickHouseDeserializer, ClickHouseSerializer and ClickHouseDataProcessor etc., your code will be impacted.

Lastly, I'm sure there's always good reason to use nullable types, but another alternative approach I can think of is to use Array(Int8), which avoids boxed type and null flag in serialization.

zhicwu:
Fixed in #1124.

PR #1110, opened 2022-10-17
Use logback for logging

n-oden:
- add logback jars to all components
- pin slf4j version to what's bundled w/ logback-classic

### DIFF
--- a/README.md
+++ b/README.md
@@ -177,6 +177,14 @@ Use `mvn -DskipITs clean verify` to compile and generate packages if you're usin
 </toolchains>
 ```
 
+## Logging
+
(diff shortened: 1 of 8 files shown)


n-oden:
See https://github.com/ClickHouse/clickhouse-jdbc-bridge/pull/159 for backstory. :)

This is my first time working with a multi-component repo like this so if there's a more concise way to do this I'm all ears.

n-oden:
I'm pretty sure that the test failure here is unrelated to my code change.

zhicwu:
Sorry for the late response @n-oden and thanks for the pull request. Yes, the build failure is completely irrelevant to the change.

The proposed change looks good but I would not merge. The major concern is that we prefer least dependency, but the change introduced many mandatory dependencies even we know they're optional. Since we support slf4j, I think it's better to leave that to downstream applications to choose, agree?

Having said that, I think it's probably a good idea to add an example for using logback in case it's needed somewhere.

PR #1127, opened 2022-11-16
[Bug] Write time field data loss precision.

gaaraG:
https://github.com/ClickHouse/clickhouse-jdbc/issues/1126
The Calendar class's setTime preserves only the millisecond bit and should retain its original precision after the time zone is converted.


zhicwu:
Thanks for your contribution! @gaaraG, could you sign CLA and rebase to develop branch?

gaaraG:
Of course.

PR #1129, opened 2022-11-20
Fix security warning mentioned in #1128 and discard two sonar issues

zhicwu:


zhicwu:
The build failure is irrelevant to the changes.

PR #1137, opened 2022-11-25
Add the verb "be" in error messages

fabiencelier:


zhicwu:
Thank you @fabiencelier!

PR #1146, opened 2022-12-04
Add apache http client

JackyWoo:
This pr will close #1139 .

### change log
1. Add apache http client
2. who support configuring socket option.
3. Make `ClickHouseHttpClientTest` parameterized

PR #1171, opened 2023-01-03
Enhance Apache HTTP client

zhicwu:
This is mainly a follow-up PR for #1146 with below changes:
* added more socket options and renamed `socket_op_ip_tos` to `socket_ip_tos` for consistency @JackyWoo 
* removed `ByteArraysInputStream` because it's been covered by ClickHouseInputStream
* enabled more tests for Apache HTTP Client and got failures fixed accordingly


zhicwu:
@JackyWoo, you may use the most recent snapshot at [here](https://s01.oss.sonatype.org/content/repositories/snapshots/com/clickhouse/clickhouse-jdbc/0.4.0-SNAPSHOT/) to validate the changes at your end. I still have some more changes to add, but I believe I can release 0.4.0 in a few days.

JackyWoo:
@zhicwu I will test it tomorrow

clickhouse-client/src/main/java/com/clickhouse/client/config/ClickHouseClientOption.java
```diff
@@ -252,15 +252,59 @@ public enum ClickHouseClientOption implements ClickHouseOption {
      * Socket timeout in milliseconds.
      */
     SOCKET_TIMEOUT("socket_timeout", 30 * 1000, "Socket timeout in milliseconds."),
+    /**
+     * Whether allows for the reuse of local addresses and ports. See
+     * {@link java.net.StandardSocketOptions#SO_REUSEADDR}.
+     */
+    SOCKET_REUSEADDR("socket_reuseaddr", false,
+            "Whether allows for the reuse of local addresses and ports. "
+                    + "Only works for client using custom Socket(e.g. TCP client or HTTP provider with custom SocketFactory etc.)."),
```
mayankvadariya, line 261: @zhicwu does custom SocketFactory refers to `ApacheHttpConnectionImpl.SocketFactory` and `ApacheHttpConnectionImpl.SSLSocketFactory` ?

I am wondering if it is possible to inject a custom socket factory.

zhicwu: > Does custom SocketFactory refers to ApacheHttpConnectionImpl.SocketFactory and ApacheHttpConnectionImpl.SSLSocketFactory?

Yes, for now. Ideally we should have a custom socket factory for not only Apache HTTP Client but also others(i.e. NIO-based implementation).

> I am wondering if it is possible to inject a custom socket factory.

Sorry it's not supported at this point. What's the purpose of injecting custom socket factory? Is it to use consistent socket options across datasources?

mayankvadariya: Thanks for confirming. Yes, I would like to create a custom socket.

mayankvadariya: Are you planning to include that feature in the future?

mayankvadariya: Thanks for confirming. Yes, I would like to create a custom socket.

Are you planning to include that feature in the future? Other JDBC drivers(eg. mysql) offer such capabilities.

zhicwu: Thanks for the suggestion. Just created a separate issue for this, should not be hard to add the feature.

Neyzoter:
Only works for client using custom Socket.

---

I tcpdump, and found that socket_keepalive=true cannot set socket to keepalive. Should I use costom SocketFactory to set socket option?

PR #1174, opened 2023-01-04
support more compression algorithms

zhicwu:
Added support for more compression algorithms(see table shown below), and enabled client options `compress_alogrithm`, `decompress_alogrithm`, `decompress_level`.

| Algorithm | Compress Client Request | Decompress Server Response |
| ---------- | ------------------------- | ------------------------------ |
| brotli | N | Y |
| bz2 | N | Y |
| deflate | N | Y |
| gzip | N | Y |
| lz4 | Y | Y |
| snappy | Y | N (see ClickHouse/ClickHouse#44885) |
| xz | N | Y |
| zstd | Y | Y |

Closes #1101

PR #1175, opened 2023-01-05
`isReusable` should return a constant value

JackyWoo:
`isReusable` should return a constant value.
Or  `ClickHouseHttpClient` may close on flying connection when creating new connection.

[        if (connection != null && connection.isReusable()) {
            closeConnection(connection, false);
        }](https://github.com/ClickHouse/clickhouse-jdbc/blob/60b8ca8646ca981986c7263d5b19a668dcb94c5c/clickhouse-http-client/src/main/java/com/clickhouse/client/http/ClickHouseHttpClient.java#L51-L53)

### DIFF
--- a/clickhouse-http-client/src/main/java/com/clickhouse/client/http/ApacheHttpConnectionImpl.java
+++ b/clickhouse-http-client/src/main/java/com/clickhouse/client/http/ApacheHttpConnectionImpl.java
@@ -22,7 +22,7 @@ import org.apache.hc.client5.http.config.ConnectionConfig;
 import org.apache.hc.client5.http.impl.classic.CloseableHttpClient;
 import org.apache.hc.client5.http.impl.classic.CloseableHttpResponse;
 import org.apache.hc.client5.http.impl.classic.HttpClientBuilder;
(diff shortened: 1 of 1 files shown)


JackyWoo:
When connection is on fly by a request  the others wait it released?

zhicwu:
Thanks for testing and raising the PR. `ClickHouseHttpConnection.isReusable()` suggests if the connection instance can be reused among requests. Each time when there's a new request made to the same server, we have two cases to deal with:
1. reuse the same connection if it's not busy on something else
2. create a new connection so that we can process requests in parallel

The change I made was for the latter one, which also fixed the test failures. Yes, it may cause connection leak but I thought the connections are all managed by Apache connection manager. Did you run into memory issue using latest code on develop branch? Will run stress test tonight to see how it goes.

JackyWoo:
@zhicwu We can use `PoolingHttpClientConnectionManager` and keep `isReusable` return true in which way connections will be managed by Apache connection manager.

zhicwu:
Thank you @JackyWoo, the changes look good. I'm closing this PR as the changes are now merged into #1174, where I cleaned up code for consistency and improved compression/decompression support.

PR #1194, opened 2023-01-17
fixes issue with shiftOrResize when a mark occurs before and a reset …

PHaroZ:
In `ru.yandex.clickhouse.response.StreamSplitter` if `mark()` is called (eg from `ru.yandex.clickhouse.response.ClickHouseResultSet#onTheSeparatorRow`) and then `shiftOrResize()` occurs, then `reset()` leaves the stream in an invalid result.

`posNext` and `posRead` are reset but `buf` still contain the next chunk of data.

In such a case `ru.yandex.clickhouse.response.ClickHouseResultSet#getTotals` fails if the data of the total line are split across two chunks of data.

I know this class should be removed but, while waiting for a new stable version, this issue cause unavoidable error. (and, as far as I know, there is no equivalent feature in the new jdbc driver)

### DIFF
--- a/clickhouse-jdbc/src/main/java/ru/yandex/clickhouse/response/StreamSplitter.java
+++ b/clickhouse-jdbc/src/main/java/ru/yandex/clickhouse/response/StreamSplitter.java
@@ -22,7 +22,7 @@ public class StreamSplitter {
     // position until which the values from buf already passed out through next()
     private int posNext;
 
-    private int markedRead;
+    private boolean marked;
     private int markedNext;
 
     private boolean readOnce;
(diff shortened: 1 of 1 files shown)


zhicwu:
Hi @PHaroZ, thanks for sharing. As of now, in 0.3.2+, the last row in result set is the `total`, which might be weird. Perhaps we can fix this by adding multi-resultset support, which returns two resultsets for a `with total` query.

wudstrand:
Do you think this [bug](https://github.com/ClickHouse/clickhouse-java/issues/1299) is related?

PR #1219, opened 2023-02-02
add changelog.md

mshustov:
## Summary
We want to use CHANGELOG.md as a source of truth across different communication channels (the documentation website, ClickHouse UI, etc.)
To make it portable, CHANGELOG must follow a pre-defined structure and format,  outlined in the current PR.
We still can keep https://github.com/ClickHouse/clickhouse-java/releases section, but it's better to link the CHANGELOG to a release instead of copying the content in several places.


zhicwu:
There was a change log under clickhouse-jdbc. Do you want to restore that?

PR #1247, opened 2023-02-19
make jdbc driver executable without any dependency

zhicwu:
## Summary
* clean up pom files by removing unnecessary dependencies
* make all dependencies optional for clickhouse-jdbc
* add main class to clickhouse-jdbc for testing purpose
* add GraalVM configuration for building native image

## Checklist


zhicwu:
| Query | Java Client | Java Client(native) | JDBC Driver | JDBC Driver(native) |
| ------ | ----------- | ------------------- | ------------ | -------------------- |
| select 500000000 | 200ms | 10ms | 330ms | 28ms |
| select number::Int8 v from numbers(500000000) | 10s | 29s | 15s | 48s |

GraalVM CE does not support PGO(profile-guided optimization), so the native binary only reduced startup time and memory footprints without C1/C2 compilation as in JIT. Would be interesting if someone with GraalVM EE to fine-tune the configuration and share the test results.
Note: upx reduced the binary file from 32.7MB to 10.8MB, without performance degradation.

PR #1263, opened 2023-03-01
remove redundant socket options

JackyWoo:
Remove redundant socket options in `ApacheHttpConnectionImpl`.


zhicwu:
Thanks @JackyWoo. Can we remove the duplicated lines at bottom instead? It is intended to enforce the same socket timeout and buffer size whether or not they have been provided.

PR #1265, opened 2023-03-02
Add Credentials Overload for Client

rickysaltzer:
- Overloaded the `ClickHouseClient.newInstance` method to allow supplying default `ClickHouseCredentials` during creation.

## Summary

Given how convenient calling `ClickHouseClient.newInstance(..)` is, it makes sense to overload and allow the user to supply default credentials.


zhicwu:
Thanks for your contribution @rickysaltzer!

PR #1271, opened 2023-03-09
More intuitive client info

JackyWoo:
## Summary
Now when getting JVM version info we use property in this order ` java.vendor.version`, `java.vm.version`,  `java.version`.

And Oracal JDK 1.8 we get 
```
Java HotSpot(TM) 64-Bit Server VM/25.221-b11
```

Maybe `1.8` is more intuitive.

## Change log

when getting JVM version info use `java.version`.

### DIFF
--- a/clickhouse-client/src/main/java/com/clickhouse/client/config/ClickHouseClientOption.java
+++ b/clickhouse-client/src/main/java/com/clickhouse/client/config/ClickHouseClientOption.java
@@ -437,10 +437,7 @@ public enum ClickHouseClientOption implements ClickHouseOption {
         }
         CLIENT_OS_INFO = new StringBuilder().append(getSystemConfig("os.name", "O/S")).append('/')
                 .append(getSystemConfig("os.version", UNKNOWN)).toString();
-        String javaVersion = System.getProperty("java.vendor.version");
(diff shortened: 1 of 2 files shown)


zhicwu:
Thank you @JackyWoo. I agree that `java.version` is shorter and cleaner, but it was changed to `java.vendor.version` so that we can figure out who should pay Oracle for using their JDK ;)

Perhaps we should add an option for simplified user agent?

JackyWoo:
@zhicwu sorry for late reply.  

For commercial use of Java version 8 (JDK and JRE), update 201 is the latest free available update. 

And `java.version` result is in the following format:

```
major_version.minor_version.micro_version[_update_version][-milestone]
```

So maybe `java.version` can also tell us whether we should pay Oracle for using their JDK. 
Please correct me if I am wrong.

zhicwu:
Thanks @JackyWoo. But Oracle, the company, may change its idea time from time, which is not predictable. I do agree the user agent is too long and should be simplified. Adding an option on client side might not be useful - I'm checking if there's any server setting that I can leverage for the change.

JackyWoo:
@zhicwu Thanks, I will close the PR temporarily.

PR #1336, opened 2023-04-27
fix too much socket fds for apache http connection

JackyWoo:
## Summary

I found apache http connection created to many socket fds(ESTABLISHED and TIMED_WAIT) and it will cause connection timeout error. The following is some tests.

1.apache http connection(before)
```
netstat -na | grep 8123 | grep ESTABLISHED | wc -l && netstat -na | grep 8123 | wc -l
     626
   26523
netstat -na | grep 8123 | grep ESTABLISHED | wc -l && netstat -na | grep 8123 | wc -l

     671
   12215
```

2.apache http connection(after)
```
netstat -na | grep 8123 | grep ESTABLISHED | wc -l && netstat -na | grep 8123 | wc -l
      16
      19
netstat -na | grep 8123 | grep ESTABLISHED | wc -l && netstat -na | grep 8123 | wc -l
      22
```

3.http url connection
```
netstat -na | grep 8123 | grep ESTABLISHED | wc -l && netstat -na | grep 8123 | wc -l
      24
     146
netstat -na | grep 8123 | grep ESTABLISHED | wc -l && netstat -na | grep 8123 | wc -l
      24
     157
```

PS: the performance testing results have shown a slight decrease
```
Benchmark                                   (client)  (connection)  (statement)   (type)   Mode  Cnt     Score     Error  Units
Query.selectInt8  clickhouse-apache-http-client-jdbc         reuse       normal  default  thrpt   20  1861.249 ± 219.318  ops/s
Query.selectInt8  clickhouse-apache-http-client-jdbc         reuse     prepared  default  thrpt   20  1868.108 ± 199.007  ops/s
```
## Change log
1. disable keep alive

### DIFF
--- a/clickhouse-http-client/src/main/java/com/clickhouse/client/http/ApacheHttpConnectionImpl.java
+++ b/clickhouse-http-client/src/main/java/com/clickhouse/client/http/ApacheHttpConnectionImpl.java
@@ -25,13 +25,13 @@ import org.apache.hc.client5.http.impl.classic.CloseableHttpClient;
 import org.apache.hc.client5.http.impl.classic.CloseableHttpResponse;
 import org.apache.hc.client5.http.impl.classic.HttpClientBuilder;
 import org.apache.hc.client5.http.impl.io.BasicHttpClientConnectionManager;
-import org.apache.hc.client5.http.impl.io.PoolingHttpClientConnectionManager;
 import org.apache.hc.client5.http.socket.ConnectionSocketFactory;
 import org.apache.hc.client5.http.socket.PlainConnectionSocketFactory;
 import org.apache.hc.client5.http.ssl.DefaultHostnameVerifier;
 import org.apache.hc.client5.http.ssl.SSLConnectionSocketFactory;
 import org.apache.hc.core5.http.Header;
 import org.apache.hc.core5.http.HttpRequest;
+import org.apache.hc.core5.http.HttpResponse;
 import org.apache.hc.core5.http.config.Registry;
 import org.apache.hc.core5.http.config.RegistryBuilder;
 import org.apache.hc.core5.http.io.SocketConfig;
@@ -84,6 +84,10 @@ public class ApacheHttpConnectionImpl extends ClickHouseHttpConnection {
             r.register("https", SSLSocketFactory.create(c));
         }
         return HttpClientBuilder.create().setConnectionManager(new HttpConnectionManager(r.build(), c))
+                // disable keep alive for it may cause too many socket fd.
+                .setKeepAliveStrategy((HttpResponse response, HttpContext context) -> null)
+                .setConnectionReuseStrategy((HttpRequest request, HttpResponse response, HttpContext context)
+                        -> Boolean.parseBoolean(null))
                 .disableContentCompression().build();
     }
 


zhicwu:
Hi @JackyWoo, as I recalled, JDK HttpClient has similar issue. We probably should consider reverting to PoolingHttpClientConnectionManager, incorporating options to manage the max open connections and concurrent connections per route, among other features. However, this may not be a big concern, as the default keep-alive duration set by the ClickHouse server is 3 seconds, which is unlikely to be altered by many users.

Back to the initial concern, adding Apache http client was to support socket options, right? Have you considered NIO based solution? A full-featured HTTP client may not be necessary; a basic one that functions with ClickHouse would suffice. To begin, I introduced `AbstractSocketClient` as an alternative to the Apache HTTP client(and upcoming TCP client), hopefully with improved performance and eventually retire HttpURLConnection too. Would you be interested in pursuing this option instead?

JackyWoo:
@zhicwu Thanks for your reply.
1. I will try `PoolingHttpClientConnectionManager` with  re-usable = true.
2. `ApacheHttpConnection` got more fds than `HttpURLConnection ` as I tested and it lead to exceptions, but `HttpURLConnection ` had no exceptions.
3. May I ask when `AbstractSocketClient ` will be available?

zhicwu:
>  May I ask when `AbstractSocketClient` will be available?

There's an incomplete version at [here](https://github.com/ClickHouse/clickhouse-java/blob/794683247ecc4b99a6c2d4253247927e42930aeb/clickhouse-client/src/main/java/com/clickhouse/client/AbstractSocketClient.java#L33), which still requires extensive work for http support.

JackyWoo:
@zhicwu I revert connection manatger to `PoolingHttpClientConnectionManager ` and make it reusable.

benchmark shows throughput promotes about 9% than `http-url-client`
```
Benchmark                                   (client)  (connection)  (statement)   (type)   Mode  Cnt     Score     Error  Units
Query.selectInt8  clickhouse-apache-http-client-jdbc         reuse       normal  default  thrpt   20  2539.465 ± 158.646  ops/s
Query.selectInt8  clickhouse-apache-http-client-jdbc         reuse     prepared  default  thrpt   20  2570.479 ± 205.385  ops/s
```
And fds

```
netstat -na | grep 8123 | grep ESTABLISHED | wc -l && netstat -na | grep 8123 | wc -l
      24
      38
```

clickhouse-client/src/main/java/com/clickhouse/client/config/ClickHouseClientOption.java
```diff
@@ -378,7 +378,13 @@ public enum ClickHouseClientOption implements ClickHouseOption {
      * false.
      */
     USE_TIME_ZONE("use_time_zone", "", "Time zone of all DateTime* values. "
-            + "Only used when use_server_time_zone is false. Empty value means client time zone.");
+            + "Only used when use_server_time_zone is false. Empty value means client time zone."),
+
+    /**
+     * Max connections of ApacheHttpConnectionImpl
+     */
+    APACHE_HTTP_CLIENT_MAX_CONNECTIONS("apache_http_client_max_connections", 10,
```
zhicwu, line 386: Could you please move this option to `ClickHouseHttpClientOption`? We may promote it to a more generic one like `MAX_OPEN_CONNECTIONS` at a later stage.

zhicwu (changes requested):

zhicwu (approved):

PR #1350, opened 2023-05-04
add methods to write nullable fields

centralhardware:
## Summary

This changes allow to rewrite this code 

```java
BinaryStreamUtils.writeNonNull(stream);
BinaryStreamUtils.writeUuid(stream,<some variable>);
```
to 
```java
BinaryStreamUtils.writeNullableUuid(stream, <some variable>)
```

## Checklist
- [*] A human-readable description of the changes was provided to include in CHANGELOG

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+## 0.5.1
+### Usability improvement
+* BinaryStreamUtils: Add methods to write nullable fields
+
 ## 0.5.0
 ### Bug Fixes
 * Java client threw confusing error when query is invalid.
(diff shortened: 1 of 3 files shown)


zhicwu:
Thank you @centralhardware for the contribution. It's definitely great idea to improve usability by providing convenient APIs. However, as you probably noticed, package `com.clickhouse.data.format` is invisible to other modules.
https://github.com/ClickHouse/clickhouse-java/blob/46914be6ef13c574a48b1691ca805692bb72f2cc/clickhouse-data/src/main/java9/module-info.java#L6-L11

Understood it can still be accessed in JDK 8, but the APIs are inmature and they'll be refactored soon. As to `BinaryStreamUtils`, which was created for dealing with RowBinary data format, it's probably not good idea to use it in your code, because it will make your application tied to specific data format, meaning you have to change the code to use  Native, which is a more effcient data format in some cases.

I'm not sure the problem you're trying to resolve, but for now, maybe it's better to stay with JDBC or the serde APIs as shown at [here](/ClickHouse/clickhouse-java/tree/main/clickhouse-data#quick-start).

PR #1357, opened 2023-05-09
use Vector API to read/write primitive arrays

zhicwu:
## Summary
Use Vector API to read/write primitive arrays.

## Checklist


zhicwu:
Tried VarHandle, Foreign Memory & Vector APIs, but the performance improvement was less than 10% when using JDK 21-ea.26. Will revisit this after JDK 21 is officially released.

PR #1382, opened 2023-06-14
Disable Object Reuse By Default

rickysaltzer:
## Summary
- Enabling object reuse by default has the potential to cause unforeseen bugs in user code.

- When a user decides to pass a `Stream<ClickHouseRecord>` back as a `List<ClickHouseRecord>`, all objects by default will be pointing back to the same object. This can be extremely jarring for users who aren't aware objects are being reused.

- Added a `reuseObjects()` method to quickly enable object reuse when appropriate. This allows the user to decide when memory efficiency is a goal.

## Checklist

### DIFF
--- a/clickhouse-client/src/main/java/com/clickhouse/client/ClickHouseRequest.java
+++ b/clickhouse-client/src/main/java/com/clickhouse/client/ClickHouseRequest.java
@@ -1768,6 +1768,18 @@ public class ClickHouseRequest<SelfT extends ClickHouseRequest<SelfT>> implement
     }
 
     /**
+     * Enable object reuse during result collection, resulting in better memmory efficiency.
+     * WARNING: If you plan to pass the underlying {@link com.clickhouse.data.ClickHouseRecord} back without
+     * intermediate extraction, this object will potentially be overwritten with the last seen value of the stream.
+     * @return the request with object reuse enabled
+     */
+    @SuppressWarnings("unchecked")
+    public SelfT reuseObjects() {
+        option(ClickHouseClientOption.REUSE_VALUE_WRAPPER, "true");
+        return (SelfT) this;
+    }
+
+    /**
      * Sets a setting. See
      * https://clickhouse.tech/docs/en/operations/settings/settings/ for more
      * information.
--- a/clickhouse-client/src/test/java/com/clickhouse/client/ClientIntegrationTest.java
+++ b/clickhouse-client/src/test/java/com/clickhouse/client/ClientIntegrationTest.java
@@ -75,7 +75,6 @@ import java.util.ArrayList;
 import java.util.Collections;
 import java.util.List;
 import java.util.Map;
-import java.util.Set;
 import java.util.TimeZone;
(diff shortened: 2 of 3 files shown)


rickysaltzer:
I use this library within Kotlin code, and so, Kotlin Collections makes it **very** easy to go from `Iterable<ClickHouseRecord>` to `Sequence` or `List`. This is what caused me to spend quite a long time figuring out why my tests were failing (expecting data from clickhouse to be correct).

zhicwu:
Thanks for your contribution @rickysaltzer! Besides memory efficiency, object creation also slows things down - the CI failure was because all tests couldn't finish in 15 minutes.

I think the proposed change will greatly impact to all reads, so I'd not suggest to do that. Have you tried `ClickHouseRecord.copy()` if you don't need to process many records? Alternatively, we may add a new method in the same class to extract all values? For example:
```java
public Object[] extractValues() {
  int size = size();
  Object[] arr = new Object[size];
  for (int i = 0; i < size; i++) {
      arr[i] = getValue(i).asObject();
  }
  return arr;
}
```

mshustov:
@zhicwu  It would be worthwhile to brainstorm alternative changes to API with @rickysaltzer and @mzitnik. Data integrity is a top priority. Let's try to find a balance here.

zhicwu:
Thanks again @rickysaltzer for the inputs! Your points are indeed valid and well-reasoned. However, I would like to emphasize that it's important to consider the differences in memory efficiency and performance between a small library optimized for a single JVM and a distributed middleware like Flink.

As you know, we have multiple APIs to choose from, each with its own characteristics. JDBC is a well-known and mature option, while R2DBC is asynchronous and gaining popularity, although the driver has not thoroughly tested yet. On the other hand, the Java client provides better performance and lower memory usage compared to others. If performance and memory usage are not a concern, why not stick with JDBC?

Anyway, I think what we're trying to resolve here is to improve Java client API to minimize unintended side effects. `ClickHouseResponse.records(Class<T>)` can be think of an attempt, which has no such issue and it will be faster than `records()` once ASM is integrated. Apart from that, I hope we can eventually drop `ClickHouse*Value` and potentially `ClickHouseRecord`. These additions were initially made to support a middleware called JDBC bridge, which is no longer a part of the goals. Instead, a potential approach could be to utilize `Object[]` along with precise APIs to retrieve each value, without relying on implicit type conversions, which can sometimes be problematic.

@mzitnik & @mshustov, anything to add?

rickysaltzer:
Thanks for your response, I do very much appreciate the efficiency we're trying to maintain, especially when it comes to higher-level APIs leveraging this one. 

That being said, I think coming up with an elegant solution to this issue is warranted.

mzitnik:
A few comments here 
I think API should be self-descriptive. I would also not expect Data integrity issues @zhicwu I understand that currently, it blocks our CI from completing in 15 min. Do you have any expectations for when ASM will be integrated?
The main argument here is on the correct default behavior. 
Thanks @rickysaltzer for rising this issue.

Issue #1389, opened 2023-07-04
Nested column in combination with AggregateFunction isn't working for JDBC driver


mdonkers:
### Describe the bug

When creating an `AggregateFunction` with `Nested` column, the JDBC driver cannot process the column.
For example 

```sql
set flatten_nested=0;
create table test_simple_aggregate_nested(id Int8, n0 SimpleAggregateFunction(anyLast, Nested(a String,b String))) ENGINE = AggregatingMergeTree() ORDER BY (id)
```
will throw an exception that `bString` column cannot be processed, because spaces are removed while processing the column description.

### Steps to reproduce
1. Create an `AggregatingMergeTree` table with `AggregateFunction` with `Nested` column type. And `flatten_nested=0` set to leave it as Nested type instead of flattened Arrays.
2. Try to query the data using the JDBC driver

### Error log

(exception is from the IntelliJ Database integration, which used JDBC as well)

```
java.sql.SQLException: Unknown data type: bString, 
server ClickHouseNode [uri=http://localhost:8123/, 
options={custom_http_params=session_id=DataGrip_46c66044-85bd-4257-b6dd-ea2cee5bd931}]@246729261 
at com.clickhouse.jdbc.SqlExceptionUtils.handle(SqlExceptionUtils.java:85) 
at com.clickhouse.jdbc.SqlExceptionUtils.create(SqlExceptionUtils.java:31) 
at com.clickhouse.jdbc.SqlExceptionUtils.handle(SqlExceptionUtils.java:90) 
at com.clickhouse.jdbc.internal.ClickHouseStatementImpl.getLastResponse(ClickHouseStatementImpl.java:120) 
at com.clickhouse.jdbc.internal.ClickHouseStatementImpl.executeQuery(ClickHouseStatementImpl.java:474) 
at com.clickhouse.jdbc.internal.ClickHouseStatementImpl.execute(ClickHouseStatementImpl.java:462) in RemoteStatementImpl.execute(RemoteStatementImpl.java:79) 
Caused by: java.lang.IllegalArgumentException: Unknown data type: ImplementationString 
at com.clickhouse.data.ClickHouseDataType.of(ClickHouseDataType.java:238) 
at com.clickhouse.data.ClickHouseColumn.readColumn(ClickHouseColumn.java:487) 
at com.clickhouse.data.ClickHouseColumn.parse(ClickHouseColumn.java:564) 
[...]

PR #1390, opened 2023-07-04
[bug #1389] Resolve processing Nested column type in combination with AggregateFunction

mdonkers:
## Summary
Resolves https://github.com/ClickHouse/clickhouse-java/issues/1389

Skipping of whitespace is prevented when processing a section within brackets.
I had to change some other tests that were failing now, but don't think it would break anything else. Not sure why whitespace is removed in the first place?

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@
 
 ### Bug Fixes
 * Java client threw confusing error when query is invalid.
+* JDBC Driver correctly processes `AggregateFunction(Nested(...))` columns
 
 ## 0.4.6, 2023-05-02
 ### New Features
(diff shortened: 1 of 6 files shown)


zhicwu:
Thank you @mdonkers for the fix, it looks good to me. The build failure is unrelated, but I'll need to fix that before merging this PR. I'll get back to this next week.

mdonkers:
Thanks!

One question from my side, why are whitespaces removed in the first place? I've seen e.g. `" a, b c"` in one of the tests that gets translated to the params `Arrays.asList("a", "bc")`
Would there be situations where the whitespace not being removed could now lead to issues? I guess if there is something like `( a, b c)` it would result in `Arrays.asList("a", "b c"). I contemplated making the `readParameters` smarter to look back at `Nested` or some other keywords to detect name-type combinations but that would be a much bigger change.

Just trying to understand if my PR could have unexpected consequences somewhere else.

mdonkers:
Hi @zhicwu, I expect this one slipped through the cracks.
Do you have an expectation on when this could be merged?

zhicwu:
Hi @mdonkers, sorry for taking so long. I need to play with the change by a few more tests. Will try to get it merged tonight.

zhicwu (approved):

PR #1394, opened 2023-07-08
optimize find column index

Okapist:
## Summary

ClickhouseResultSet look for column index on each row. It use expensive equalsIgnoreCase method.
I added index cache for it.


Okapist:
Collections.synchronizedMap(new HashMap<>()) may be replaced with ConcurrentHashMap if not need to support old jdk versions.

zhicwu:
Thank you @Okapist for the pull request and sorry for the late reply.

Can we create the HashMap only when needed(e.g. first time calling findColumnIndex)? Besides, when instantiate the cache, can we specify the load factor since we know the maximum number of columns?

Okapist:
> Can we create the HashMap only when needed(e.g. first time calling findColumnIndex)? Besides, when instantiate the cache, can we specify the load factor since we know the maximum number of columns?

Good idea. Pull request updated.

zhicwu:
Thanks @Okapist. Will merge this one after fixing build failure.

Issue #1391, opened 2023-07-04
Support custom socket factory


zhicwu:
As requested at https://github.com/ClickHouse/clickhouse-java/pull/1171#discussion_r1248335413

mayankvadariya:
:tada: Thanks much @zhicwu  for creating this issue. I will keep an eye to see when this gets closed.

mayankvadariya:
Thanks @zhicwu for working on this feature. Can we release this feature?

mayankvadariya:
@mzitnik, can you or @zhicwu help to release this feature?

mayankvadariya:
@zhicwu @mzitnik checking again here if this feature can be released?

mzitnik:
@mayankvadariya Sorry for the delay. We are working on that having some issues in our release process. You can bypass it for now by building a local version. Again sorry for the delay

zhicwu:
> @mayankvadariya Sorry for the delay. We are working on that having some issues in our release process. You can bypass it for now by building a local version. Again sorry for the delay

Hopefully I can fix the issues in `nightly` and `release` two workflows in the weekend.

mayankvadariya:
@zhicwu @mzitnik do you have any update on the release timeline?

PR #1400, opened 2023-07-12
Add support for custom socket factory

mayankvadariya:
## Summary
https://github.com/ClickHouse/clickhouse-java/issues/1391 - Attempt to inject a custom socket factory with `APACHE_HTTP_CLIENT` connection provider.
## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,6 @@
+### New Features
+* Add support for customer socket factory [#1391](https://github.com/ClickHouse/clickhouse-java/issues/1391)
+
 ## 0.5.0
 ### Breaking Changes
 * ClickHouseByteBuffer can no longer be extended
--- a/clickhouse-http-client/src/main/java/com/clickhouse/client/http/ApacheHttpConnectionImpl.java
+++ b/clickhouse-http-client/src/main/java/com/clickhouse/client/http/ApacheHttpConnectionImpl.java
@@ -1,7 +1,6 @@
 package com.clickhouse.client.http;
 
 import com.clickhouse.client.AbstractSocketClient;
-
 import com.clickhouse.client.ClickHouseClient;
 import com.clickhouse.client.ClickHouseConfig;
 import com.clickhouse.client.ClickHouseNode;
@@ -42,6 +41,9 @@ import org.apache.hc.core5.ssl.SSLContexts;
 import org.apache.hc.core5.util.Timeout;
 import org.apache.hc.core5.util.VersionInfo;
 
+import javax.net.ssl.SSLContext;
+import javax.net.ssl.SSLException;
+
 import java.io.BufferedReader;
 import java.io.ByteArrayInputStream;
 import java.io.ByteArrayOutputStream;
@@ -61,9 +63,6 @@ import java.util.TimeZone;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.TimeUnit;
 
-import javax.net.ssl.SSLContext;
-import javax.net.ssl.SSLException;
-
 /**
  * Created by wujianchao on 2022/12/1.
  */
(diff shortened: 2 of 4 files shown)


mayankvadariya:
@zhicwu I attempted this PR to add support for custom socket factory. Please review the approach and advise for any changes. Thank you!

zhicwu:
Thank you @mayankvadariya for the pull request! Just some immediate feedback I can think of:
1. Please consider to use more generic names for the new options, for examples: CUSTOM_SOCKET_FACTORY & CUSTOM_SECURE_SOCKET_FACTORY, so that later they can be used in `clickhouse-tcp-client` as well. Alternatively, we can move the options to `ClickHouseHttpOption` (better without the http* prefix), so that it's clear that they're just for `clickhouse-http-client`.
2. Since we're adding new options, in case user wants to use their own extension in an environment like DBeaver, perhaps it's better to try context class loader first, and then parent class loader for loading the custom factory class
3. Does it make sense to reuse existing socket options instead of additional ones for custom socket factory?

mayankvadariya:
1. Aptly said. Moved the options to `ClickHouseHttpOption`
2. Please elaborate more on this.
3. It's more about scoping input properties to `ClickHouseOption`, unknown/unparsed properties are being ignored through driver. Either we can add a way to allow accessing unknown/unparsed properties through `ClickHouseOption` or add these options. This way custom socket factory doesn't have to depend on the supported parameters by the Clickhouse drivers. Most of the drivers allows access to unknown/unparsed properties when creating a custom socket factory.
I am also up for getting rid of both options(`*_OPTIONS`) by making unknown/unparsed input properties somehow available through `ClickHouseConfig`.

zhicwu:
> 2. Please elaborate more on this.

Assume we're using DBeaver, when loading a class, I was thinking maybe it's better to start with the context class loader.
Anyway, let's not worry about this for now. I need to think it through and then probably come up with a separate PR later for the change.

image

> 3. It's more about scoping input properties to `ClickHouseOption`, unknown/unparsed properties are being ignored through driver. Either we can add a way to allow accessing unknown/unparsed properties through `ClickHouseOption` or add these options. This way custom socket factory doesn't have to depend on the supported parameters by the Clickhouse drivers. Most of the drivers allows access to unknown/unparsed properties when creating a custom socket factory.
>   I am also up for getting rid of both options(`*_OPTIONS`) by making unknown/unparsed input properties somehow available through `ClickHouseConfig`.

Good point. Perhaps we just need an extra field in ClickHouseConfig for unknown properties. Will make some changes this weekend.

mayankvadariya:
@zhicwu Did you get any chance to add extra config for getting unknown properties?

zhicwu:
> @zhicwu Did you get any chance to add extra config for getting unknown properties?

Sorry I need someone from ClickHouse team to merge #1403 first to unblock my work, and then I'll be able to make the change and merge other PRs. Let me ping them on Slack.

mayankvadariya:
@zhicwu we are waiting on this to get merged at the earliest. Do you have any timeline on when this PR might get merged?

anusudarsan:
@zhicwu any progress on the blocker here?

zhicwu:
> @zhicwu any progress on the blocker here?

Hi @anusudarsan, sorry for the delay. I had meeting with ClickHouse team on Monday - we'll gradually drop gRPC client but keep it for now. I'll fix CI issue and get back to this tonight.

mzitnik:
@zhicwu any progress on it?

mayankvadariya:
@zhicwu Checking for the progress on the blockers.

zhicwu:
Sorry for being super lazy - I was focusing on something else 😂

The initial CI failure has been addressed, but this PR is still failing compilation likely due to moving classes to the http client module. Regarding supporting unknown properties, I plan to complete that this weekend along with two other changes:
1. Stop printing properties in exceptions
2. Support properties with defaults (e.g. new Properties(defaultProps))

mayankvadariya:
Thanks @zhicwu for checking. I replaced string.formatted with string.format as the CI builds with java8. Please trigger the workflow again.

zhicwu:
Hi @mayankvadariya, over the weekend, I gave this some thought and believe it would be better to create a generic interface to define custom socket factories. This would allow support for not just Apache HTTP Client but other implementations as well. I've created a separate PR for this approach - see #1420. Please let me know if you have any concerns with this proposal.

zhicwu:
Close as this will be implemented in #1420.

PR #1434, opened 2023-08-28
Get ResultSet meta data before execution

zhicwu:
## Summary
Get ResultSet meta data before execution

## Checklist

### DIFF
--- a/clickhouse-jdbc/src/test/java/com/clickhouse/jdbc/ClickHousePreparedStatementTest.java
+++ b/clickhouse-jdbc/src/test/java/com/clickhouse/jdbc/ClickHousePreparedStatementTest.java
@@ -4,6 +4,7 @@ import java.io.ByteArrayInputStream;
 import java.io.File;
 import java.io.IOException;
 import java.math.BigDecimal;
+import java.math.BigInteger;
 import java.net.Inet4Address;
 import java.net.Inet6Address;
 import java.net.MalformedURLException;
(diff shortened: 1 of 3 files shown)


clickhouse-jdbc/src/test/java/com/clickhouse/jdbc/ClickHousePreparedStatementTest.java
```diff
@@ -2005,6 +2007,53 @@ public void testInsertWithSettings() throws SQLException {
         }
     }
 
+    @Test(groups = "integration")
+    public void testGetMetaData() throws SQLException {
+        try (Connection conn = newConnection(new Properties());
+                PreparedStatement ps = conn.prepareStatement("select ? a, ? b")) {
+            ResultSetMetaData md = ps.getMetaData();
```
ebyhr, line 2014: What happens if we pass DDL (e.g. CREATE TABLE) to `conn.prepareStatement`? It would be nice to add a test in my opinion.

zhicwu: Thanks for taking your time to review the PR!

> What happens if we pass DDL (e.g. CREATE TABLE) to conn.prepareStatement?

`null` will be returned at this point. Below is the check in `SqlBasedPreparedStatement.java`. However, in ClickHouse, create table on cluster will actually return ResultSet, but I guess we don't need that as it's not helping for push-down implementation.
```java
if (parsedStmt.isRecognized() && !parsedStmt.isQuery()) {
    return null;
}
```

> It would be nice to add a test in my opinion.

Yes, the implementation is incomplete and I'll add more tests for sure.

sylph-eu:
Hi, any plans to complete this PR?

sylph-eu:
JFYI #1700

PR #1460, opened 2023-09-28
including back slf4j api for apache http

BentsiLeviav:
The new default HTTP library (Apache HTTP) requires a slf4j dependency. Failing to provide this dependency will result in exceptions, leading the system to fall back to an alternative HTTP library.

### DIFF
--- a/clickhouse-http-client/pom.xml
+++ b/clickhouse-http-client/pom.xml
@@ -45,10 +45,6 @@
                     <groupId>org.apache.httpcomponents.core5</groupId>
                     <artifactId>httpcore5</artifactId>
                 </exclusion>
-                <exclusion>
-                    <groupId>org.slf4j</groupId>
-                    <artifactId>slf4j-api</artifactId>
-                </exclusion>
             </exclusions>
             <optional>true</optional>
         </dependency>


mzitnik (approved):

rsim:
@mzitnik @BentsiLeviav I tried the latest `clickhouse-jdbc-0.5.0-shaded.jar` and got a lot of warnings in the console
```
WARNING: Error when creating APACHE_HTTP_CLIENT, fall back to HTTP_URL_CONNECTION
java.lang.NoClassDefFoundError: org/slf4j/LoggerFactory
  at com.clickhouse.client.internal.apache.hc.client5.http.impl.io.PoolingHttpClientConnectionManager.<clinit>(PoolingHttpClientConnectionManager.java:108)
  at com.clickhouse.client.http.ApacheHttpConnectionImpl.getDefaultUserAgent(ApacheHttpConnectionImpl.java:214)
  at com.clickhouse.client.http.ClickHouseHttpConnection.getUserAgent(ClickHouseHttpConnection.java:405)
  at com.clickhouse.client.http.ClickHouseHttpConnection.<init>(ClickHouseHttpConnection.java:367)
  at com.clickhouse.client.http.ApacheHttpConnectionImpl.<init>(ApacheHttpConnectionImpl.java:78)
  at com.clickhouse.client.http.ClickHouseHttpConnectionFactory.createConnection(ClickHouseHttpConnectionFactory.java:22)
  at com.clickhouse.client.http.ClickHouseHttpClient.newConnection(ClickHouseHttpClient.java:56)
  at com.clickhouse.client.http.ClickHouseHttpClient.newConnection(ClickHouseHttpClient.java:26)
```

I found out about this change and when I added to classpath `slf4j-api-2.0.9.jar` and `slf4j-nop-2.0.9.jar` then there were no warnings and `APACHE_HTTP_CLIENT` was used.

Previously the shaded jar contained all needed "shaded" dependencies. I'm wondering if slf4j should also be included as shaded dependencies in the shaded jar to avoid the need to add them explicitly? In our case we would prefer to have just a single shaded jar file which does not depend on any external logging facitilites.

Kind regards,
Raimonds

BentsiLeviav:
Hi @rsim,

Thank you for your comment.

Our `APACHE_HTTP_CLIENT` is based on `httpclient5`, which requires SLF4J. We can't shade it as it is a dependency of a dependency.

> We would prefer to have just a single shaded JAR file which does not depend on any external logging facilities.

This could be achieved by having a shaded Apache HTTP package that shades the use of SLF4J.

We'll reach out to the Apache HTTP Client team to get more information about this matter. In case they provide us with a shaded version, we'll update it.

PR #1483, opened 2023-11-09
DoubleValue update of inf str

Chen768959:
## Summary
When SQL is formatted using the `FORMAT TabSeparatedWithNamesAndTypes` in ClickHouse
the `ClickHouseTabSeparatedProcessor` will `update` the value as a `String`. In this case, if the double type is `inf, -inf, or nan`, it will not be able to be parsed.

for example
image

## Checklist


mzitnik:
@Chen768959 sorry for the delay 
Can you some tests to cover

Chen768959:
> @Chen768959 sorry for the delay Can you some tests to cover
Sure

chernser (approved):

PR #1485, opened 2023-11-10
ClickHouseDatabaseMetaData.getTables dont ignoreError

jianchengwang:
because getTables method is very commonly used. Just for me, I don't import org.lz4:lz4-java:1.8.0 but because ClickHouseDatabaseMetaData.getTables ignore error, I debug for a long time.


mzitnik (commented):
Tests are failing can you check

Issue #1532, opened 2024-01-25
Integration tests are not run because clickhouse-server does not start in testcontainers


mbaksheev:
### Describe the bug
Actually it is not a bug, it is an issue with integration tests.

According to [CONTIBUTING.md](https://github.com/ClickHouse/clickhouse-java/blob/main/CONTRIBUTING.md) the integration tests use testcontainers to run clickhouse-server.

When I try to run the integration tests they are failed with error
```
2024-01-25 16:10:14:046 +0100 [main] ERROR clickhouse/clickhouse-server:latest - Could not start container java.lang.IllegalStateException: Wait strategy failed. Container exited with code 70
```
The actual problem is that clickhouse-server does not start inside docker and fails with error:
```
2024.01.25 14:15:40.021419 [ 20 ] {} <Error> RaftInstance: got exception: open: Address family not supported by protocol [system:97] on port 9234
2024.01.25 14:15:40.021583 [ 20 ] {} <Error> void DB::KeeperDispatcher::initialize(const Poco::Util::AbstractConfiguration &, bool, bool, const MultiVersion<Macros>::Version &): Code: 568. DB::Exception: Cannot create interserver listener on port 9234. (RAFT_ERROR), Stack trace (when copying this message, always include the lines below):
```
This happens because clickhouse-keeper starts in ipv6 mode by default, but in docker ipv6 is disabled by default.
Also it's not obvious from the contribution guide that ipv6 support is required to run integration tests.

According to this [issue ](https://github.com/ClickHouse/ClickHouse/issues/51474) ipv6 can be disabled for clickhouse keeper with option `keeper_server.enable_ipv6=false`

My question:
_Does it make sense to change clickhouse-keeper configuration for integration tests and disable ipv6?   I can prepare pull request for it if there are no reasons to keep the current behavior._

### Steps to reproduce
1. Run integration tests: `mvn clean verify` - ipv6 should be disabled in docker (by default)

### Expected behavior
Integration tests are run

mzitnik:
@mbaksheev 
Can you provide the the machine specs to reproduce?

mbaksheev:
@mzitnik 

```
HP HP EliteBook 850 G8 Notebook PC
CPU: 11th Gen Intel® Core™ i7-1165G7 @ 2.80GHz × 8
--------------------------------------------------------------
OS: Ubuntu 22.04.3 LTS 6.5.0-14-generic
--------------------------------------------------------------
[...]

PR #1533, opened 2024-01-26
Disable ipv6 mode for keeper-server for testcontainers environment.

mbaksheev:
## Summary
Addreses #1532

Disable ipv6 mode for keeper-server for testcontainers environment because otherwise ClickHouse does not start in docker and integration tests are failed to run


chernser (approved):
Verified.

PR #1569, opened 2024-03-20
Feature: Persist effect of "SET ROLE" queries.  

chernser:
## Summary
Implement setting role per java client. The main goal is to persist set of roles between different HTTP calls because there is no persistent session for a user in that case. 

## Checklist


mzitnik (approved):
LGTM

slvrtrn:
@chernser, let's retest with the head version of the ClickHouse docker image and merge it.

PR #1573, opened 2024-03-21
Add client address to http referer by default

JackyWoo:
## Summary

- Add client address to http referer by default

Send a request to ClickHouse with default settings and We will get client address in ClickHouse `system.query_log`

```
Row 1:
──────
query:           select * from system.numbers limit 53
client_hostname:
client_name:
http_referer:    10.211.42.53
http_user_agent: ClickHouse-JdbcDriver/0.6.0-SNAPSHOT (Linux/5.4.0-53-generic; OpenJDK 64-Bit Server VM/17.0.10+7-Ubuntu-120.04.1; HttpURLConnection; rv:be06439)
```

## Checklist

### DIFF
--- a/clickhouse-http-client/src/main/java/com/clickhouse/client/http/ClickHouseHttpClient.java
+++ b/clickhouse-http-client/src/main/java/com/clickhouse/client/http/ClickHouseHttpClient.java
@@ -2,11 +2,11 @@ package com.clickhouse.client.http;
 
 import java.io.IOException;
 import java.io.UncheckedIOException;
+import java.net.InetAddress;
+import java.net.NetworkInterface;
+import java.net.SocketException;
 import java.nio.charset.StandardCharsets;
-import java.util.Collection;
-import java.util.Collections;
-import java.util.List;
-import java.util.Map;
+import java.util.*;
 import java.util.concurrent.CompletionException;
 
 import com.clickhouse.client.AbstractClient;
@@ -18,8 +18,10 @@ import com.clickhouse.client.ClickHouseRequest;
 import com.clickhouse.client.ClickHouseResponse;
 import com.clickhouse.client.ClickHouseTransaction;
 import com.clickhouse.client.ClickHouseStreamResponse;
+import com.clickhouse.client.config.ClickHouseClientOption;
 import com.clickhouse.client.http.config.ClickHouseHttpOption;
 import com.clickhouse.config.ClickHouseOption;
+import com.clickhouse.data.ClickHouseChecker;
 import com.clickhouse.logging.Logger;
 import com.clickhouse.logging.LoggerFactory;
 
@@ -27,6 +29,57 @@ public class ClickHouseHttpClient extends AbstractClient<ClickHouseHttpConnectio
     private static final Logger log = LoggerFactory.getLogger(ClickHouseHttpClient.class);
 
     static final List<ClickHouseProtocol> SUPPORTED = Collections.singletonList(ClickHouseProtocol.HTTP);
+
+    public static class HostNameAndAddress{
+        public String hostName;
+        public String address;
+    }
+
+    private static HostNameAndAddress getLocalHost() {
+        // get local address but not localhost
+        HostNameAndAddress hostNameAndAddress = new HostNameAndAddress();
+        try {
+            Enumeration<NetworkInterface> networkInterfaces = NetworkInterface.getNetworkInterfaces();
+
+            for (NetworkInterface ni : Collections.list(networkInterfaces)) {
+                Enumeration<InetAddress> inetAddresses = ni.getInetAddresses();
+                for (InetAddress ia : Collections.list(inetAddresses)) {
+                    // We just use the first non-loopback address
+                    if (!ia.isLoopbackAddress()) {
+                        hostNameAndAddress.address = ia.getHostAddress();
+                        hostNameAndAddress.hostName = ia.getCanonicalHostName();
+                        break;
+                    }
+                }
+            }
+        } catch (SocketException e) {
+            // ignore
+        }
+        return hostNameAndAddress;
+    }
+
+    public static String getReferer(ClickHouseConfig config) {
+        String referer = null;
+        if (!ClickHouseChecker.isNullOrEmpty(config.getStrOption(ClickHouseHttpOption.SEND_HTTP_CLIENT_ID)))
+        {
+            if (ClickHouseChecker.isNullOrEmpty(ClickHouseOption.toKeyValuePairs(config.getStrOption(ClickHouseHttpOption.CUSTOM_HEADERS)).get("referer"))) {
+                if (config.getStrOption(ClickHouseHttpOption.SEND_HTTP_CLIENT_ID).equals("HOST_NAME"))
+                    referer = LOCAL_HOST.hostName;
+                else
+                    referer = LOCAL_HOST.address;
+            }
+        }
+        return referer;
+    }
+
+    public static HostNameAndAddress LOCAL_HOST = null;
+
+    public ClickHouseHttpClient() {
+        synchronized (this) {
+            if (LOCAL_HOST == null)
+                LOCAL_HOST = getLocalHost();
+        }
+    }
 
     @Override
     protected boolean checkConnection(ClickHouseHttpConnection connection, ClickHouseNode requestServer,
@@ -118,7 +171,7 @@ public class ClickHouseHttpClient extends AbstractClient<ClickHouseHttpConnectio
             httpResponse = conn.post(config, sql, sealedRequest.getInputStream().orElse(null),
                     sealedRequest.getExternalTables(), sealedRequest.getOutputStream().orElse(null),
                     ClickHouseHttpConnection.buildUrl(server.getBaseUri(), sealedRequest),
-                    ClickHouseHttpConnection.createDefaultHeaders(config, server, conn.getUserAgent()),
+                    ClickHouseHttpConnection.createDefaultHeaders(config, server, conn.getUserAgent(), getReferer(config)),
                     postAction);
         } else {
             httpResponse = conn.post(config, sql, sealedRequest.getInputStream().orElse(null),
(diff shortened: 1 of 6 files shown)


clickhouse-http-client/src/main/java/com/clickhouse/client/http/ClickHouseHttpConnection.java
```diff
@@ -59,6 +52,30 @@ public abstract class ClickHouseHttpConnection implements AutoCloseable {
     private static final byte[] SUFFIX_STRUCTURE = "_structure\"\r\n\r\n".getBytes(StandardCharsets.US_ASCII);
     private static final byte[] SUFFIX_FILENAME = "\"; filename=\"".getBytes(StandardCharsets.US_ASCII);
 
+    private static final String LOCAL_ADDRESS;
+
+    static {
+        // get local address but not localhost
+        String address = "";
+        try {
+            Enumeration<NetworkInterface> networkInterfaces = NetworkInterface.getNetworkInterfaces();
```
chernser, line 61: Good day, @JackyWoo!
Working with OS in the static block would lead to problem on application startup and we need to avoid it. 
Another issue with static blocks is they may not be bypassed when not needed. 
Current change overrides default behavior for all client what may break someone's application if they rely on it.

chernser:
@JackyWoo 
 Feature looks reasonable. Would it be fine to pass `referer` as some sort of setting? In this case we would let client to identify itself more predictably.

clickhouse-http-client/src/main/java/com/clickhouse/client/http/ClickHouseHttpConnection.java
```diff
@@ -206,6 +223,11 @@ protected static Map<String, String> createDefaultHeaders(ClickHouseConfig confi
             map.put(name, value);
         }
 
+        if (!LOCAL_ADDRESS.isEmpty() && !map.containsKey("referer"))
+        {
+            map.put("referer", LOCAL_ADDRESS);
```
chernser, line 228: In some cases `referer` would be an application name. For example, in the cloud with dynamic IP addresses.

clickhouse-http-client/src/main/java/com/clickhouse/client/http/ClickHouseHttpConnection.java
JackyWoo: Thanks

clickhouse-http-client/src/main/java/com/clickhouse/client/http/ClickHouseHttpConnection.java
chernser: I mean that some clients want to use a site name or application. As an example, when application run in k8s IP address would be hard to track back - so pod ID would be more handy.

chernser:
@JackyWoo thank you for the update! 
I need some time to review.

clickhouse-http-client/src/main/java/com/clickhouse/client/http/ClickHouseHttpConnection.java
```diff
@@ -358,6 +388,11 @@ protected ClickHouseHttpConnection(ClickHouseNode server, ClickHouseRequest<?> r
             throw new IllegalArgumentException("Non-null server and request are required");
         }
 
+        synchronized (ClickHouseHttpConnection.class) {
```
chernser, line 391: To avoid extra work please move local address logic into `com.clickhouse.client.http.ClickHouseHttpClient`
1. do initialization in the `ClickHouseHttpClient` constructor and instance fields to keep address and host name. 
2. add extra parameter `referer` to the `com.clickhouse.client.http.ClickHouseHttpConnection#createDefaultHeaders` (we will refactor it later) and path it from client where used. If referer == null or empty - no Referer header should be sent.

chernser (commented):
Please add a few tests. For example, test different option values and extra long custom ID

See `com.clickhouse.client.http.ClickHouseHttpConnectionTest` as reference. 

Thanks!

JackyWoo:
@chernser please review when you are free.

clickhouse-http-client/src/main/java/com/clickhouse/client/http/ClickHouseHttpClient.java
```diff
@@ -28,6 +30,57 @@ public class ClickHouseHttpClient extends AbstractClient<ClickHouseHttpConnectio
...
+    public static class HostNameAndAddress{
+        public String hostName;
+        public String address;
+    }
+
+    private static HostNameAndAddress getLocalHost() {
+        // get local address but not localhost
+        HostNameAndAddress hostNameAndAddress = new HostNameAndAddress();
+        try {
+            Enumeration<NetworkInterface> networkInterfaces = NetworkInterface.getNetworkInterfaces();
```
chernser, line 42: This is the only place where SocketException may be throw and only in case when host has no interface. 
It is a good practice to minimize scope of a try-catch to deal only with a few exceptions at a time.
Please, rethrow SocketException inside a RuntimeException to indicate a critical error.

clickhouse-http-client/src/main/java/com/clickhouse/client/http/ClickHouseHttpClient.java
```diff
@@ -28,6 +30,57 @@ public class ClickHouseHttpClient extends AbstractClient<ClickHouseHttpConnectio
 
     static final List<ClickHouseProtocol> SUPPORTED = Collections.singletonList(ClickHouseProtocol.HTTP);
 
+    public static class HostNameAndAddress{
+        public String hostName;
+        public String address;
+    }
+
+    private static HostNameAndAddress getLocalHost() {
```
chernser, line 38: please make it a package private non-static method to make test easier:
```
HostNameAndAddress getLocalHost();
```

clickhouse-http-client/src/main/java/com/clickhouse/client/http/ClickHouseHttpClient.java
```diff
@@ -28,6 +30,57 @@ public class ClickHouseHttpClient extends AbstractClient<ClickHouseHttpConnectio
 
     static final List<ClickHouseProtocol> SUPPORTED = Collections.singletonList(ClickHouseProtocol.HTTP);
 
+    public static class HostNameAndAddress{
```
chernser, line 33: Lest just use `InetAddress` because it has all information we need.

clickhouse-http-client/src/main/java/com/clickhouse/client/http/ClickHouseHttpClient.java
```diff
@@ -28,6 +30,57 @@ public class ClickHouseHttpClient extends AbstractClient<ClickHouseHttpConnectio
...
+                    }
+                }
+            }
+        } catch (SocketException e) {
+            // ignore
+        }
+        return hostNameAndAddress;
+    }
+
+    public static String getReferer(ClickHouseConfig config) {
```
chernser, line 61: please make a package private.

clickhouse-http-client/src/main/java/com/clickhouse/client/http/ClickHouseHttpClient.java
```diff
@@ -28,6 +30,57 @@ public class ClickHouseHttpClient extends AbstractClient<ClickHouseHttpConnectio
...
+                if (config.getStrOption(ClickHouseHttpOption.SEND_HTTP_CLIENT_ID).equals("HOST_NAME"))
+                    referer = LOCAL_HOST.hostName;
+                else
+                    referer = LOCAL_HOST.address;
+            }
+        }
+        return referer;
+    }
+
+    public static HostNameAndAddress LOCAL_HOST = null;
```
chernser, line 75: please make it an instance variable

clickhouse-http-client/src/main/java/com/clickhouse/client/http/ClickHouseHttpClient.java
```diff
@@ -28,6 +30,57 @@ public class ClickHouseHttpClient extends AbstractClient<ClickHouseHttpConnectio
...
+
+    public ClickHouseHttpClient() {
+        synchronized (this) {
```
chernser, line 78: please remove synchronization as LOCAL_HOST is an instance field.

clickhouse-http-client/src/main/java/com/clickhouse/client/http/ClickHouseHttpClient.java
```diff
@@ -28,6 +30,57 @@ public class ClickHouseHttpClient extends AbstractClient<ClickHouseHttpConnectio
...
+            // ignore
+        }
+        return hostNameAndAddress;
+    }
+
+    public static String getReferer(ClickHouseConfig config) {
+        String referer = null;
+        if (!ClickHouseChecker.isNullOrEmpty(config.getStrOption(ClickHouseHttpOption.SEND_HTTP_CLIENT_ID)))
+        {
+            if (ClickHouseChecker.isNullOrEmpty(ClickHouseOption.toKeyValuePairs(config.getStrOption(ClickHouseHttpOption.CUSTOM_HEADERS)).get("referer"))) {
```
chernser, line 65: There is no guarantee that custom header "referer" will be always in lower case, because this map is filed by user and not checked in this way. 
I suggest to move this logic to `createDefaultHeaders` method where it unpacks values and checks if "authorization" header is set (203:196). If done so - the same logic as for "authorization" header can be implemented.

chernser:
@JackyWoo thank you very much for the update. I've reviewed. Overall looks good - just need a few changes. 
Thank you!

clickhouse-http-client/src/main/java/com/clickhouse/client/http/ClickHouseHttpClient.java
JackyWoo: `LOCAL_HOST` is needed in both in `ClickHouseHttpClient` and `ClickHouseHttpConnection`, because `ClickHouseHttpConnection.createDefaultHeaders` is invoked both there.

So I made` LOCAL_HOST` as a static field.

chernser: I think, I've missed something. 
LOCAL_HOST is accessed only in the `ClickHouseHttpClient.getReferer(config)`.  
 `ClickHouseHttpConnection.createDefaultHeaders` uses already baked value. 

Or do you mean it is used in tests also? ClickHouseHttpConnectionTest

Thanks!

chernser (approved):

PR #1582, opened 2024-04-03
Using a more elegant Pattern-Matcher method to extract error code from error message

zalj:
## Summary
Using a more elegant Pattern-Matcher method to extract error code from error message.

Remove and avoid from hard coding of "Poco::Exceptions" and similar format.


chernser (approved):

chernser:
@zalj Thank you for the contribution!

Issue #1470, opened 2023-10-12
Failover not working for multiple ck nodes using random load balance


happyjohnson123:
### Describe the bug
I want to use jdbc to connect the multiple clickhouse nodes, set the load balance with 'random', and also set the failover with '3'. But when I configure my two endpoints (one endpoint is valid, and another endpoint is invalid) in jdbc url, the query will pop up the connection timeout exception for that invalid endpoint, failover does not work. 

### Steps to reproduce
1. Using docker to setup a clickhouse instance with 'yandex/clickhouse-server:latest' image version
2. Configure the jdbc url with 'url:jdbc:clickhouse:http://192.168.2.122:8123,192.168.0.103:8123/default&load_balancing_policy=random&failover=3' (192.168.2.122 is invalid endpoint, 192.168.0.103 is valid endpoint which is setup in step 1)
3. Using jdbc code to query result

### Expected behaviour
failover mechanism should switch to the valid endpoint to query the result

### Code example
```
String url = "jdbc:clickhouse:http://192.168.2.122:8123,192.168.0.103:8123/default?load_balancing_policy=random&failover=3";
Properties p = new Properties();
p.setProperty("user", "default");
p.setProperty("password", "");
ClickHouseDataSource dataSource = new ClickHouseDataSource(url, p);
try (Connection connection = dataSource.getConnection()) {
        Statement stmt = connection.createStatement();
        ResultSet resultSet = stmt.executeQuery("show databases");
}
```
### Error log
java.sql.SQLException: Connect to http://192.168.2.122:8123 [/192.168.2.122] failed: connect timed out, server ClickHouseNode [uri=http://192.168.2.122:8123/default]@645498125
at com.clickhouse.jdbc.SqlExceptionUtils.handle(SqlExceptionUtils.java:85)
	at com.clickhouse.jdbc.SqlExceptionUtils.create(SqlExceptionUtils.java:31)
	at com.clickhouse.jdbc.SqlExceptionUtils.handle(SqlExceptionUtils.java:90)
	at com.clickhouse.jdbc.internal.ClickHouseConnectionImpl.getServerInfo(ClickHouseConnectionImpl.java:131)
	at com.clickhouse.jdbc.internal.ClickHouseConnectionImpl.<init>(ClickHouseConnectionImpl.java:335)
	at com.clickhouse.jdbc.ClickHouseDataSource.getConnection(ClickHouseDataSource.java:46)
[...]

PR #1609, opened 2024-04-11
Fix failover for Apache HTTP client. Connect timeout error is not detected correctly

mbaksheev:
## Summary
Fix for issue #1470 

### Issue
Connect time out errors are not detected correctly for Apache HTTP client because error message  for this client is in different format e.g: 
```
Connect to http://192.168.2.122:8123/ [/192.168.2.122] failed: connect timed out
```
As a result errors like
```
Caused by: com.clickhouse.client.internal.apache.hc.client5.http.ConnectTimeoutException: Connect to http://192.168.2.122:8123/ [/192.168.2.122] failed: connect timed out
```
are not detected as eligible errors to execute failover.

### Fix
`ClickHouseException.isConnectTimedOut` is changed to detect `connect timed out` substring in any part of error message

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ### Bug Fixes
 - Change RowBinaryWithDefaults settings. Output is changed from true to false
+- Fix failover for Apache HTTP client. Connect timout error is not detected correctly
 
 ## 0.6.0-patch3
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ### Bug Fixes
 - Change RowBinaryWithDefaults settings. Output is changed from true to false
-- Fix failover for Apache HTTP client. Connect timout error is not detected correctly
+- Fix failover for Apache HTTP client. Connect timeout error is not detected correctly
 
 ## 0.6.0-patch3
 
--- a/clickhouse-cli-client/src/test/java/com/clickhouse/client/cli/ClickHouseCommandLineClientTest.java
+++ b/clickhouse-cli-client/src/test/java/com/clickhouse/client/cli/ClickHouseCommandLineClientTest.java
@@ -3,22 +3,21 @@ package com.clickhouse.client.cli;
 import com.clickhouse.client.ClickHouseClient;
 import com.clickhouse.client.ClickHouseClientBuilder;
 import com.clickhouse.client.ClickHouseException;
+import com.clickhouse.client.ClickHouseLoadBalancingPolicy;
 import com.clickhouse.client.ClickHouseNode;
 import com.clickhouse.client.ClickHouseProtocol;
 import com.clickhouse.client.ClickHouseServerForTest;
 import com.clickhouse.client.ClientIntegrationTest;
 import com.clickhouse.client.cli.config.ClickHouseCommandLineOption;
 import com.clickhouse.data.ClickHouseCompression;
-
-import java.io.IOException;
-import java.util.concurrent.ExecutionException;
-
 import org.testcontainers.containers.GenericContainer;
-import org.testng.Assert;
 import org.testng.SkipException;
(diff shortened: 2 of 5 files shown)


clickhouse-cli-client/src/test/java/com/clickhouse/client/cli/ClickHouseCommandLineClientTest.java
```diff
@@ -190,4 +189,14 @@ public void testTransactionTimeout() throws ClickHouseException {
     public void testRowBinaryWithDefaults() throws ClickHouseException, IOException, ExecutionException, InterruptedException {
         throw new SkipException("Skip due to supported");
     }
+
+    @Test(dataProvider = "loadBalancingPolicies", groups = {"unit"})
```
mbaksheev, line 193: I'm not sure about failover for CLI client. Currently it does not work. So the tests are ignored. Please, let me know if failover should work for CLI client as well

clickhouse-grpc-client/src/test/java/com/clickhouse/client/grpc/ClickHouseGrpcClientTest.java
```diff
@@ -205,4 +205,13 @@ public void testRowBinaryWithDefaults() throws ClickHouseException, IOException,
         throw new SkipException("Skip due to supported");
     }
 
+    @Test(dataProvider = "loadBalancingPolicies", groups = {"unit"})
+    public void testLoadBalancingPolicyFailover(ClickHouseLoadBalancingPolicy loadBalancingPolicy) {
+        throw new SkipException("Skip due to failover is not supported");
```
mbaksheev, line 210: The same as for CLI client.
I'm not sure about failover  here. Currently it does not work, so the tests are ignored. Please, let me know if failover should work for GRPC client as well

clickhouse-client/src/main/java/com/clickhouse/client/ClickHouseException.java
```diff
@@ -119,10 +118,7 @@ public static ClickHouseException forCancellation(Exception e, ClickHouseNode se
     public static boolean isConnectTimedOut(Throwable t) {
         if (t instanceof SocketTimeoutException || t instanceof TimeoutException) {
             String msg = t.getMessage();
-            if (msg != null && msg.length() >= MSG_CONNECT_TIMED_OUT.length()) {
-                msg = msg.substring(0, MSG_CONNECT_TIMED_OUT.length()).toLowerCase(Locale.ROOT);
-            }
-            return MSG_CONNECT_TIMED_OUT.equals(msg);
+            return msg != null && msg.toLowerCase().contains(MSG_CONNECT_TIMED_OUT);
```
mbaksheev, line 121: Probably this logic for error detection should be more client-specific. I have some ideas how it can be implemented, but it requires more efforts for  implementation. 
As a fast fix this solution looks good from my point of view.

chernser:
@mbaksheev Thank you for the PR! we will review it shortly.

clickhouse-client/src/main/java/com/clickhouse/client/ClickHouseException.java
chernser: Parsing error message is too fragile. We should use error codes or exception classes. 
The old code misses some other cases. For example, if we were waiting for a task and failed with TimeoutException it will not be a connection timeout, but in some cases it means that network is disrupted and we need to switch server.

chernser (approved):

CHANGELOG.md
```diff
@@ -2,6 +2,7 @@
 
 ### Bug Fixes
 - Change RowBinaryWithDefaults settings. Output is changed from true to false
+- Fix failover for Apache HTTP client. Connect timeout error is not detected correctly
```
chernser, line 5: Special thanks for updating the CHANGELOG!

PR #1616, opened 2024-04-22
WIP - Insert V2

Paultagoras:
## Checklist

### DIFF
--- a/client-v2/src/main/java/com/clickhouse/client/api/Client.java
+++ b/client-v2/src/main/java/com/clickhouse/client/api/Client.java
@@ -32,6 +32,7 @@ import com.clickhouse.client.api.internal.TableSchemaParser;
 import com.clickhouse.client.api.query.QueryResponse;
 import com.clickhouse.client.api.query.QuerySettings;
 import com.clickhouse.data.format.BinaryStreamUtils;
+import org.apache.commons.lang3.time.StopWatch;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.slf4j.MDC;
@@ -247,7 +248,7 @@ public class Client {
         if (data == null || data.isEmpty()) {
             throw new IllegalArgumentException("Data cannot be empty");
         }
-        long s1 = System.currentTimeMillis();
+        StopWatch watch = StopWatch.createStarted();
 
         //Add format to the settings
         if (settings == null) {
@@ -271,7 +272,6 @@ public class Client {
             throw new IllegalArgumentException("No serializer found for the given class. Please register() before calling this method.");
         }
 
-        long s2 = System.currentTimeMillis();
         //Call the static .serialize method on the POJOSerializer for each object in the list
         for (Object obj : data) {
             for (POJOSerializer serializer : serializers) {
@@ -282,9 +282,9 @@ public class Client {
                 }
             }
         }
-        long s3 = System.currentTimeMillis();
 
-        LOG.debug("Total serialization time: {}", s3 - s1);
+        watch.stop();
+        LOG.debug("Total serialization time: {}", watch.getTime());
         return insert(tableName, new ByteArrayInputStream(stream.toByteArray()), settings);
     }
 
@@ -294,18 +294,18 @@ public class Client {
     public InsertResponse insert(String tableName,
                                      InputStream data,
                                      InsertSettings settings) throws IOException, ClientException {
-        long s1 = System.currentTimeMillis();
+        StopWatch watch = StopWatch.createStarted();
         InsertResponse response;
         try (ClickHouseClient client = createClient()) {
             ClickHouseRequest.Mutation request = createMutationRequest(client.write(getServerNode()), tableName, settings)
-                    .format((ClickHouseFormat) settings.getSetting(ClickHouseClientOption.FORMAT.getKey()));
+                    .format(settings.getFormat());
 
             Future<ClickHouseResponse> future;
             try(ClickHousePipedOutputStream stream = ClickHouseDataStreamFactory.getInstance().createPipedOutputStream(request.getConfig())) {
                 future = request.data(stream.getInputStream()).execute();
 
                 //Copy the data from the input stream to the output stream
-                byte[] buffer = new byte[8196];
+                byte[] buffer = new byte[settings.getInputStreamBatchSize()];
                 int bytesRead;
                 while ((bytesRead = data.read(buffer)) != -1) {
                     stream.write(buffer, 0, bytesRead);
@@ -318,8 +318,8 @@ public class Client {
             }
         }
 
-        long s2 = System.currentTimeMillis();
-        LOG.debug("Total insert (InputStream) time: {}", s2 - s1);
+        watch.stop();
+        LOG.debug("Total insert (InputStream) time: {}", watch.getTime());
         return response;
     }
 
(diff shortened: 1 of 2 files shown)


client-v2/src/main/java/com/clickhouse/client/api/Client.java
```diff
@@ -159,29 +181,137 @@ public boolean ping(int timeout) {
...
     }
 
     /**
      * Insert data into ClickHouse using a POJO
      */
-    public Future<InsertResponse> insert(String tableName,
-                                     List<Object> data,
-                                     InsertSettings settings,
-                                     List<ClickHouseColumn> columns) throws ClickHouseException, SocketException {
+    public InsertResponse insert(String tableName,
```
chernser, line 243: I think this list of exceptions is confusing because user will need to understand each case when exception is thrown to handle properly. For example, can we retry after ExecutionException? 
I suggest to wrap them all under fewer exceptions with simple meanings. For example, NetworkException, ValidationException, ClientException.

client-v2/src/main/java/com/clickhouse/client/api/Client.java
```diff
@@ -159,29 +181,137 @@ public boolean ping(int timeout) {
...
+        //Add format to the settings
+        if (settings == null) {
+            settings = new InsertSettings();
+        }
+
+        boolean hasDefaults = this.hasDefaults.get(data.get(0).getClass());
+        if (hasDefaults) {
+            settings.setFormat(ClickHouseFormat.RowBinaryWithDefaults);
+        } else {
+            settings.setFormat(ClickHouseFormat.RowBinary);
```
chernser, line 260: I think it should be RowBinaryWithNames because:
- RowBinary requires listing all columns and 
- We do not know if a POJO is fully mapped to a table

client-v2/src/main/java/com/clickhouse/client/api/Client.java
```diff
@@ -159,29 +181,137 @@ public boolean ping(int timeout) {
...
+            throw new IllegalArgumentException("Data cannot be empty");
+        }
+        long s1 = System.currentTimeMillis();
+
+        //Add format to the settings
+        if (settings == null) {
+            settings = new InsertSettings();
+        }
+
+        boolean hasDefaults = this.hasDefaults.get(data.get(0).getClass());
```
chernser, line 256: I've just realized that if POJO is not 1:1 mapping of the table then assuming that user what to use defaults may be wrong, so may be make it as settings too?

Problem may be when user mistakenly missed field in a POJO and table has defaults then we would write wrong data actually. 

It is a minor problem.

client-v2/src/main/java/com/clickhouse/client/api/Client.java
```diff
@@ -159,29 +181,137 @@ public boolean ping(int timeout) {
...
+            try(ClickHousePipedOutputStream stream = ClickHouseDataStreamFactory.getInstance().createPipedOutputStream(request.getConfig())) {
+                future = request.data(stream.getInputStream()).execute();
+
+                //Copy the data from the input stream to the output stream
+                byte[] buffer = new byte[8196];
+                int bytesRead;
+                while ((bytesRead = data.read(buffer)) != -1) {
+                    stream.write(buffer, 0, bytesRead);
+                }
+            }
```
chernser, line 308: Here we would catch IOExceptions and wrap them and log.

client-v2/src/main/java/com/clickhouse/client/api/Client.java
```diff
@@ -159,29 +181,137 @@ public boolean ping(int timeout) {
...
+            response = new InsertResponse(client, future.get());
```
chernser, line 309: this we may do under try catch to handle execution exception that report that operation timeouted to be user friendly, I think

client-v2/src/main/java/com/clickhouse/client/api/insert/InsertResponse.java
```diff
@@ -0,0 +1,25 @@
...
+        this.client = client;
+    }
+
+    public ClickHouseResponseSummary getSummary() {
+        return responseRef.getSummary();
+    }
+
+    @Override
+    public void close() {
+        responseRef.close();
```
chernser, line 22: I would wrap with try-catch so responseRef.close() exception would not cause client unclosed

chernser:
Overall looks good!

client-v2/src/main/java/com/clickhouse/client/api/Client.java
```diff
@@ -159,29 +182,145 @@ public boolean ping(int timeout) {
...
+        try (ClickHouseClient client = createClient()) {
+            ClickHouseRequest.Mutation request = createMutationRequest(client.write(getServerNode()), tableName, settings)
+                    .format((ClickHouseFormat) settings.getSetting(ClickHouseClientOption.FORMAT.getKey()));
+
+            Future<ClickHouseResponse> future;
+            try(ClickHousePipedOutputStream stream = ClickHouseDataStreamFactory.getInstance().createPipedOutputStream(request.getConfig())) {
+                future = request.data(stream.getInputStream()).execute();
+
+                //Copy the data from the input stream to the output stream
+                byte[] buffer = new byte[8196];
```
mzitnik, line 308: Do we need to make it configurable?

client-v2/src/main/java/com/clickhouse/client/api/Client.java
```diff
@@ -159,29 +182,145 @@ public boolean ping(int timeout) {
...
-                                     List<Object> data,
-                                     InsertSettings settings,
-                                     List<ClickHouseColumn> columns) throws ClickHouseException, SocketException {
+    public InsertResponse insert(String tableName,
+                                         List<Object> data,
+                                         InsertSettings settings) throws ClientException, IOException {
+        if (data == null || data.isEmpty()) {
+            throw new IllegalArgumentException("Data cannot be empty");
+        }
+        long s1 = System.currentTimeMillis();
```
mzitnik, line 250: lets start using [StopWatch](https://commons.apache.org/proper/commons-lang/apidocs/org/apache/commons/lang3/time/StopWatch.html)

client-v2/src/main/java/com/clickhouse/client/api/Client.java
```diff
@@ -159,29 +182,145 @@ public boolean ping(int timeout) {
...
+        //Create an output stream to write the data to
+        ByteArrayOutputStream stream = new ByteArrayOutputStream();
+
         //Lookup the Serializer for the POJO
+        List<POJOSerializer> serializers = this.serializers.get(data.get(0).getClass());
+        if (serializers == null || serializers.isEmpty()) {
+            throw new IllegalArgumentException("No serializer found for the given class. Please register() before calling this method.");
+        }
+
+        long s2 = System.currentTimeMillis();
```
mzitnik, line 274: lets start using [StopWatch](https://commons.apache.org/proper/commons-lang/apidocs/org/apache/commons/lang3/time/StopWatch.html)

client-v2/src/main/java/com/clickhouse/client/api/Client.java
chernser: The main problem is to be sure that we write all bytes and did not miss a column because it is missing in pojo.

chernser (approved):

PR #1624, opened 2024-04-30
Readers for binary formats

chernser:
## Summary
Implementation of readers for binary formats 

## Checklist


client-v2/src/main/java/com/clickhouse/client/api/data_formats/ClickHouseBinaryFormatReader.java
```diff
@@ -0,0 +1,150 @@
...
+     * Copies current record to a map.
+     *
+     * @param destination - map to which the record will be copied
+     * @throws IOException
+     */
+    void copyRecord(Map<String, Object> destination) throws IOException;
+
+    String getString(String colName);
+
+    Byte getByte(String colName);
```
mzitnik, line 65: why we are returning Byte instead of byte?

mzitnik, line 65: Should we have also 
byte getByte(int index)

client-v2/src/test/java/com/clickhouse/client/api/data_formats/internal/BinaryStreamReaderTests.java
```diff
@@ -0,0 +1,29 @@
...
+
+import java.io.ByteArrayInputStream;
+import java.io.ByteArrayOutputStream;
+import java.time.LocalDate;
+import java.util.TimeZone;
+
+public class BinaryStreamReaderTests {
+
+
+    @Test(groups = {"unit"}, enabled = false)
```
mzitnik, line 18: We need to extend testing to all our types

client-v2/src/test/java/com/clickhouse/client/query/QueryTests.java
```diff
@@ -65,65 +85,368 @@ public void testSimpleSelectTableFormat() {
...
+                DATASET_VALUE_GENERATORS, 10);
+        QuerySettings settings = new QuerySettings().setFormat(ClickHouseFormat.RowBinaryWithNamesAndTypes.name());
+        Future<QueryResponse> response = client.query("SELECT col1, col3, hostname() as host FROM " + table, null, settings);
+        QueryResponse queryResponse = response.get();
+
+        TableSchema schema = new TableSchema();
+        schema.addColumn("col1", "UInt32");
+        schema.addColumn("col3", "String");
+        schema.addColumn("host", "String");
+        ClickHouseBinaryFormatReader reader = createBinaryFormatReader(queryResponse, settings, schema);
```
mzitnik, line 202: I found this API a little bit overcomplicated. Didn't we agree to have a records method that you can pull the data from it ?

client-v2/src/test/java/com/clickhouse/client/query/QueryTests.java
```diff
@@ -65,65 +85,368 @@ public void testSimpleSelectTableFormat() {
...
+
+
+            // Create table
+            StringBuilder createStmtBuilder = new StringBuilder();
+            createStmtBuilder.append("CREATE TABLE IF NOT EXISTS default.").append(table).append(" (");
+            for (String column : columns) {
+                createStmtBuilder.append(column).append(", ");
+            }
+            createStmtBuilder.setLength(createStmtBuilder.length() - 2);
+            createStmtBuilder.append(") ENGINE = Memory");
```
mzitnik, line 374: Lets use mergetree and not memory mergetree is the main engine in ClickHouse i want things will be persist on disk and when we test cloud also want to make sure that we are testing mergetree

mzitnik (changes requested):

clickhouse-data/src/main/java/com/clickhouse/data/ClickHouseDataStreamFactory.java
```diff
@@ -136,8 +137,13 @@ public ClickHouseDataProcessor getProcessor(ClickHouseDataConfig config, ClickHo
             throws IOException {
         ClickHouseFormat format = ClickHouseChecker.nonNull(config, ClickHouseDataConfig.TYPE_NAME).getFormat();
         ClickHouseDataProcessor processor = null;
-        if (ClickHouseFormat.RowBinary == format || ClickHouseFormat.RowBinaryWithNamesAndTypes == format || ClickHouseFormat.RowBinaryWithDefaults == format) {
+        if (ClickHouseFormat.RowBinary == format ||
+                ClickHouseFormat.RowBinaryWithNamesAndTypes == format ||
+                ClickHouseFormat.RowBinaryWithDefaults == format ||
+                ClickHouseFormat.RowBinaryWithNames == format) {
             processor = new ClickHouseRowBinaryProcessor(config, input, output, columns, settings);
+        } else if (ClickHouseFormat.Native == format) {
```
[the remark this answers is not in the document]
chernser: I've realized it too late :-) Implementation is quite simple but I agree it is not our main focus.

client-v2/src/main/java/com/clickhouse/client/api/data_formats/ClickHouseBinaryFormatReader.java
chernser: we may have a missing value (but we may introduce separate method for that). Just in java it is more "natural" to work with objects and anyway it will need to be boxed when user want to work with collections.

chernser: My thought process was: 
- using index on its own requires sync with query all the time otherwise wrong column may be accessed
- so there should be a map between name and index 
- so I've just built-in this

If someone want to just read all columns in the loop - we may provide an iterator.

client-v2/src/test/java/com/clickhouse/client/api/data_formats/internal/BinaryStreamReaderTests.java
chernser: I will do it right after we are done with this PR.

client-v2/src/test/java/com/clickhouse/client/query/QueryTests.java
chernser: Using records means adding another layer and memory overhead, I think. So I've done 2 ways of getting data from result 
- Iterator style and getters (what I thought was the idea) 
- com.clickhouse.client.api.data_formats.ClickHouseBinaryFormatReader#copyRecord method to copy each record to a destination map to let user process it. 

What do you think is working not well?

client-v2/src/test/java/com/clickhouse/client/query/QueryTests.java
chernser: ok. Will change.

client-v2/src/main/java/com/clickhouse/client/api/data_formats/ClickHouseBinaryFormatReader.java
mzitnik: if we have the case that the client asks to return the data in CSV,TSV format we have indexes

mzitnik (approved):
Approving, but please one of my comments

PR #1633, opened 2024-05-14
Operation (Query) Metrics 

chernser:
## Summary
Adds class to represent metrics of an operation. 

## Checklist

### DIFF
--- a/client-v2/src/main/java/com/clickhouse/client/api/Client.java
+++ b/client-v2/src/main/java/com/clickhouse/client/api/Client.java
@@ -317,7 +317,8 @@ public class Client {
     public InsertResponse insert(String tableName,
                                      InputStream data,
                                      InsertSettings settings) throws IOException, ClientException {
-        StopWatch watch = StopWatch.createStarted();
+
+        long startNanoTime = System.nanoTime();
         InsertResponse response;
         try (ClickHouseClient client = createClient()) {
             ClickHouseRequest.Mutation request = createMutationRequest(client.write(getServerNode()), tableName, settings)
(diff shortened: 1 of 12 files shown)


client-v2/src/main/java/com/clickhouse/client/api/internal/StopWatch.java
```diff
@@ -0,0 +1,41 @@
...
+    public StopWatch() {
+        // do nothing
+    }
+
+    public StopWatch(long startNanoTime) {
+        this.startNanoTime = startNanoTime;
+    }
+
+    public void start() {
+        startNanoTime = System.nanoTime();
```
mzitnik, line 19: from what i remember nano time is more expensive https://www.tutorialspoint.com/java-system-nanotime-vs-system-currenttimemillis

client-v2/src/main/java/com/clickhouse/client/api/Client.java
```diff
@@ -460,4 +479,12 @@ private static Set<String> createFormatWhitelist(String shouldSupport) {
         }
         return Collections.unmodifiableSet(formats);
     }
+
+    private static final String INTERNAL_OPERATION_ID = "operationID";
+
+    private String startOperation() {
+        String operationId = UUID.randomUUID().toString();
```
mzitnik, line 486: Are we sending operationId also as **query_id** this can be a good connection between our code and the actual operation in ClickHouse

mzitnik (approved):

client-v2/src/main/java/com/clickhouse/client/api/Client.java
chernser: Good point.

client-v2/src/main/java/com/clickhouse/client/api/internal/StopWatch.java
chernser: I've found some good doc from JDK developer https://web.archive.org/web/20160308031939/https://blogs.oracle.com/dholmes/entry/inside_the_hotspot_vm_clocks 
It explains why nanoTime() *may* be slow and what should be used for precise measurement.  

Conclusion:
- nanoTime() may be slow in scale of microseconds
- nanoTime() may require hardware IO and that is why it may be slow
- nanoTime() is what recommended for elapsed time measurements. 
- currentTimeMillis() is not so precise and may have a lag of 10ms
- we will use nanoTime() until we see a noticeable drop in performance.

thalinda-oss (approved):

PR #1639, opened 2024-05-16
Fixed max buffer limitation

chernser:
## Summary
Max buffer size can now be overridden by user 

## Checklist


chernser:
@Paultagoras 
I see significant improvements when I set buffer almost equal to dataset size.  With smaller buffers transfer takes about 4 secs. 
With 750 mb buffer it is around 2 seconds. Whole dataset is generated in memory in 1600ms (1.6 s) - just for comparison. 

I think when buffer is big is may fit whole data fast but if more space is needed than it will be costly to allocate another buffer. Smaller buffer have another side effect - they require more IO calls to push data to OS level. 

So may be a good recommendation here is to use buffer size > estimated size of dataset to avoid expensive allocations. 
Note for us: we need to research a buffer allocation algorithm that would be most effective in sending big datasets.

chernser:
As I see it - if we have smaller buffers - we have to block more frequently to copy them to TCP layer. I'm currently talking about current implementation. But I have not tested it in this way.

mzitnik (approved):

PR #1648, opened 2024-05-25
[client-v2] Settings 

chernser:
## Summary
- [x] Make same names for the same settings across client API 
- [x] Document all settings
- [x] Remove unneeded or confusing setting

### DIFF
--- a/client-v2/src/main/java/com/clickhouse/client/api/Client.java
+++ b/client-v2/src/main/java/com/clickhouse/client/api/Client.java
@@ -16,6 +16,7 @@ import com.clickhouse.client.api.insert.SerializerNotFoundException;
 import com.clickhouse.client.api.internal.SerializerUtils;
 import com.clickhouse.client.api.internal.SettingsConverter;
 import com.clickhouse.client.api.internal.TableSchemaParser;
+import com.clickhouse.client.api.internal.ValidationUtils;
 import com.clickhouse.client.api.metadata.TableSchema;
 import com.clickhouse.client.api.query.QueryResponse;
 import com.clickhouse.client.api.query.QuerySettings;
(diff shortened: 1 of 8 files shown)


client-v2/src/main/java/com/clickhouse/client/api/Client.java
```diff
@@ -103,92 +144,239 @@ public Builder() {
...
+         *
+         * @param protocol - Endpoint protocol
+         * @param host - Endpoint host
+         * @param port - Endpoint port
+         */
         public Builder addEndpoint(Protocol protocol, String host, int port) {
+            if (host == null || host.isEmpty()) {
+                throw new IllegalArgumentException("Host is required");
+            }
+            if (port <= 0) {
```
mzitnik, line 188: We need to check the upper limit as well

client-v2/src/main/java/com/clickhouse/client/api/Client.java
```diff
@@ -103,92 +144,239 @@ public Builder() {
...
             return this;
         }
 
+        /**
+         * Access token for authentication with server. Required for all operations.
+         * Same access token will be used for all endpoints.
+         *
+         * @param accessToken - plain text access token
+         */
         public Builder addAccessToken(String accessToken) {
```
mzitnik, line 241: should we also change it to set instead of add

client-v2/src/main/java/com/clickhouse/client/api/Client.java
```diff
@@ -208,25 +400,29 @@ private ClickHouseNode getServerNode() {
...
     }
 
     /**
-     * Ping the server to check if it is alive
+     * Pings the server to check if it is alive
      * @param timeout timeout in milliseconds
      * @return true if the server is alive, false otherwise
      */
-    public boolean ping(int timeout) {
+    public boolean ping(long timeout) {
```
mzitnik, line 415: Should we check the value range?

mzitnik (changes requested):

mzitnik (approved):
Approve,  but please see a few comments

client-v2/src/main/java/com/clickhouse/client/api/Client.java
chernser: The Math.toIntExact(timeout) will throw exception if value overflows int. 
I would may be add max 10 minutes but is it too restrictive? May be someone will use ping to wakeup an instance - for how long would it wake up?

PR #1660, opened 2024-05-28
Save operation specific settings, and pass along server settings

Paultagoras:
Tweaking how the insert settings are handled, based on earlier discussions - changed the method signature slightly but I think it's actually clearer this way

### DIFF
--- a/client-v2/src/main/java/com/clickhouse/client/api/insert/InsertSettings.java
+++ b/client-v2/src/main/java/com/clickhouse/client/api/insert/InsertSettings.java
@@ -5,6 +5,8 @@ import java.util.HashMap;
 import java.util.Map;
 
 public class InsertSettings {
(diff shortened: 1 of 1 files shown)


client-v2/src/main/java/com/clickhouse/client/api/insert/InsertSettings.java
```diff
@@ -22,6 +23,7 @@ public InsertSettings(Map<String, Object> settings) {
 
     private void setDefaults() {// Default settings, for now a very small list
         this.setInputStreamBatchSize(8196);
```
mzitnik, line 25: Lets use const here not a number

mzitnik (changes requested):

mzitnik (approved):

Issue #1538, opened 2024-01-30
clickhouse-r2dbc program is blocked by ClickHouseResponse


wooEnrico:
### Describe the bug
The Code example is plan to execute 100 times insert, but only 10 times, and the program is blocked. Then got timeout

I am trying to solve this problem locally and seem to have found the cause. When executing`com.clickhouse.r2dbc.ClickHouseResult#getRowsUpdated` it is necessary to manually execute `com.clickhouse.client.ClickHouseResponse#close`.

### Expected behaviour
expected execute insert 100 times

### Code example
```java
        ConnectionFactory connectionFactory = ConnectionFactories
                .get("r2dbc:pool:clickhouse:http://root:123456@127.0.0.1:8123/test");

        Mono.from(connectionFactory.create()).flatMapMany(connection -> {
            return Flux.range(0, 100).flatMap(integer -> {
                return Mono.from(connection.createStatement("insert into uk(u, k) values(:u, :k)")
                        .bind("u", UUID.randomUUID().toString())
                        .bind("k", integer)
                        .execute()).flatMap(result -> Mono.from(result.getRowsUpdated()));
            }).doOnNext(System.out::println);
        }).blockLast();
```
### Error log
```
1
1
1
1
1
1
1
1
1
1
10:00:31.457 [ClickHouseWorker-16] DEBUG com.clickhouse.client.http.ClickHouseHttpClient - Query: insert into uk(u, k) values('f37b5c84-cf57-4dc0-a039-dbd8e55e175b', 24)
10:00:31.456 [ClickHouseWorker-14] DEBUG com.clickhouse.client.http.ClickHouseHttpClient - Query: insert into uk(u, k) values('db133869-ed7b-4035-ae2f-35ec69abeae3', 20)
10:00:31.458 [ClickHouseWorker-8] DEBUG com.clickhouse.client.http.ClickHouseHttpClient - Query: insert into uk(u, k) values('cce4aa17-e0d9-4826-bd0a-a24363ee6993', 21)
10:00:31.456 [ClickHouseWorker-12] DEBUG com.clickhouse.client.http.ClickHouseHttpClient - Query: insert into uk(u, k) values('49f8210c-eaeb-4e5d-95b6-8a03cbdc71a4', 22)
[...]

PR #1667, opened 2024-06-03
fix response close for getRowsUpdated(#1538)

wooEnrico:


wooEnrico:
The reason for the modification is ` com.clickhouse.r2dbc.ClickHouseResult ` Object Construction Parameters ` com.clickhouse.client.ClickHouseResponse ` did not call the close() method after use and did not release the database connection. Therefore, multiple calls can cause the database connection pool to have no available connections, resulting in timeout issues

Please refer to the problem description and code reproduction method for details #1538

chernser (approved):

PR #1671, opened 2024-06-04
Add simplified query API

chernser:
## Summary
Query API returns response object with bare InputStream. User have to choose what reader to use for reading the stream. 
Extended API makes things a bit simpler because it requests data in descriptive format and creates a reader upfront


client-v2/src/test/java/com/clickhouse/client/query/QueryTests.java
```diff
@@ -103,6 +104,26 @@ public void testSimpleQueryWithTSV() {
...
+    }
+
+    @Test(groups = {"integration"})
+    public void testReadQuery() {
+        prepareSimpleDataSet();
+
+        Future<QueryResponseReader> response = client.readQuery("SELECT * FROM " + DATASET_TABLE);
+
+        try {
+            ClickHouseBinaryFormatReader reader = response.get().getReader();
```
mzitnik, line 117: why not have a records method?

client-v2/src/test/java/com/clickhouse/client/query/QueryTests.java
```diff
@@ -103,6 +104,26 @@ public void testSimpleQueryWithTSV() {
...
+            Map<String, Object> record = new HashMap<>();
+            while (reader.next()) {
+                reader.copyRecord(record);
```
mzitnik, line 120: Do we have examples with getString / getLong / ... methods

mzitnik (approved):

client-v2/src/test/java/com/clickhouse/client/query/QueryTests.java
chernser: yes in the new example `com.clickhouse.examples.client_v2.SimpleReader`

client-v2/src/test/java/com/clickhouse/client/query/QueryTests.java
chernser: Agree. I will think and implement it more natural way as we discussed.

mzitnik (approved):

mzitnik (approved):

PR #1672, opened 2024-06-04
Update BinaryStreamUtils.java

Paultagoras:
This adds a method to convert a String to a binary array (using UTF-8)

### DIFF
--- a/clickhouse-data/src/main/java/com/clickhouse/data/format/BinaryStreamUtils.java
+++ b/clickhouse-data/src/main/java/com/clickhouse/data/format/BinaryStreamUtils.java
@@ -225,6 +225,18 @@ public final class BinaryStreamUtils {
     }
 
     /**
+     * Writes string bytes into given output stream.
+     *
+     * @param output non-null output stream
+     * @param string  non-null utf-8 string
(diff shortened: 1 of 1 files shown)


clickhouse-data/src/main/java/com/clickhouse/data/format/BinaryStreamUtils.java
```diff
@@ -224,6 +224,18 @@ public static void writeBytes(OutputStream output, byte[] bytes) throws IOExcept
...
+    /**
+     * Writes string bytes into given output stream.
+     *
+     * @param output non-null output stream
+     * @param string  non-null utf-8 string
+     * @throws IOException when failed to write value to output stream or reached
+     *                     end of the stream
+     */
+    public static void writeBytes(OutputStream output, String string) throws IOException {
+        output.write(string.getBytes(StandardCharsets.UTF_8));
```
chernser, line 236: Is it correct that method doesn't write a length before bytes? 
Also please take a look at `com.clickhouse.data.format.BinaryStreamUtils#writeString(java.io.OutputStream, java.lang.String, java.nio.charset.Charset)` - it does similar things. Will it work for your case?

PR #1674, opened 2024-06-05
Adding proxy support, adding some debug methods, tweaking styles based on Kafka Connect rework

Paultagoras:
## Checklist

### DIFF
--- a/client-v2/src/test/java/com/clickhouse/client/query/QueryTests.java
+++ b/client-v2/src/test/java/com/clickhouse/client/query/QueryTests.java
@@ -11,7 +11,6 @@ import com.clickhouse.client.ClickHouseRequest;
 import com.clickhouse.client.ClickHouseResponse;
 import com.clickhouse.client.api.Client;
 import com.clickhouse.client.api.DataTypeUtils;
-import com.clickhouse.client.api.OperationStatistics;
 import com.clickhouse.client.api.enums.Protocol;
 import com.clickhouse.client.api.data_formats.ClickHouseBinaryFormatReader;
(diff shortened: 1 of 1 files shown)


client-v2/src/main/java/com/clickhouse/client/api/enums/ProxyType.java
```diff
@@ -0,0 +1,5 @@
+package com.clickhouse.client.api.enums;
+
+public enum ProxyType {
+    IGNORE, DIRECT, HTTP, SOCKS;
```
chernser, line 4: I suggest to reuse `java.net.Proxy.Type`. 

Current class I think has some extra constants
- IGNORE - ignoring a proxy is the same as DIRECT connection 
- DIRECT - creates confusion in our case - because our API requires to add proxy and adding "direct" proxy has no effect.

client-v2/src/main/java/com/clickhouse/client/api/Client.java
```diff
@@ -376,6 +379,17 @@ public Builder setDefaultDatabase(String database) {
             return this;
         }
 
+        public Builder addProxy(ProxyType type, String host, int port) {
```
chernser, line 382: I think the main point of adding a proxy is setting credential for it. 
Another usecase is to let client understand that real server name differs from the target where request is sent. 
Lets discuss it on the standup.

mzitnik (commented):
Lets add a proxy test

mzitnik:
@Paultagoras do you know why the tests are not working?

mzitnik (approved):

mzitnik (commented):
Please make sure the tests are passing

PR #1675, opened 2024-06-06
Refactored metrics to make them more generic

chernser:
## Summary
Previously there was server stats as primitive field. It was done to reduce memory footprint. But as any early optimization it cause need to make some changes. So metrics became a map of different values and anyone may read them all from operation in a simple over-map-entries loop.


client-v2/src/test/java/com/clickhouse/client/query/QueryTests.java
```diff
@@ -920,18 +918,13 @@ public void testQueryMetrics() throws Exception {
...
-        Assert.assertEquals(serverStats.numRowsRead, rowsToInsert); // 10 rows in the table
-        Assert.assertEquals(serverStats.numRowsWritten, rowsToInsert); // 10 rows inserted
-        Assert.assertEquals(serverStats.totalRowsToRead, 0);
-        Assert.assertEquals(serverStats.resultRows, rowsToInsert);
-        Assert.assertTrue(stats.clientStatistics.getElapsedTime("query") > 0);
+        metrics = response.getMetrics();
+        System.out.println("Server read rows: " + metrics.getMetric(ServerMetrics.NUM_ROWS_READ).getLong());
+        System.out.println("Client stats: " + metrics.getMetric(ClientMetrics.OP_DURATION).getLong());
+
+        Assert.assertEquals(metrics.getMetric(ServerMetrics.NUM_ROWS_READ).getLong(), rowsToInsert); // 10 rows in the table
```
mzitnik, line 925: We discuss to have getter methods (explicit) in this case, i need `metrics.getMetric(ServerMetrics.NUM_ROWS_READ).getLong())` why not just metric.getNumRowsRead() return long

mzitnik (changes requested):

client-v2/src/test/java/com/clickhouse/client/query/QueryTests.java
chernser: There are going to be many metrics over time. Getting only few of them is not a major usecase. 
I think users will have list of metrics what they want to publish to some metrics storage - so such getter is more useful than getter for each metric.

mzitnik (approved):

PR #1676, opened 2024-06-06
Added insert that accepts byte[]

chernser:
## Summary
There are 3 major usecase for insert data:
- Data is read from some source and InputStream is created already. User just passes it to the method 
- There list of POJO's that is where serializer shines
- Data is transformed and it is written to some buffer. In most cases it is a `byte[]` 

PR add such handy method.


chernser:
This API should not  be present because we do want encourage using streaming style of data ingestion.

PR #1686, opened 2024-06-10
Fixing proxy support and added a test to cover

Paultagoras:
Turns out that the proxy_type and other properties are case sensitive!

### DIFF
--- a/client-v2/src/main/java/com/clickhouse/client/api/Client.java
+++ b/client-v2/src/main/java/com/clickhouse/client/api/Client.java
@@ -385,9 +385,9 @@ public class Client {
             ValidationUtils.checkNonBlank(host, "host");
             ValidationUtils.checkRange(port, 1, ValidationUtils.TCP_PORT_NUMBER_MAX, "port");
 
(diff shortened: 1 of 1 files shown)


client-v2/src/main/java/com/clickhouse/client/api/Client.java
```diff
@@ -385,9 +385,9 @@ public Builder addProxy(ProxyType type, String host, int port) {
             ValidationUtils.checkNonBlank(host, "host");
             ValidationUtils.checkRange(port, 1, ValidationUtils.TCP_PORT_NUMBER_MAX, "port");
 
-            this.configuration.put(String.valueOf(ClickHouseClientOption.PROXY_TYPE), type.toString());
-            this.configuration.put(String.valueOf(ClickHouseClientOption.PROXY_HOST), host);
-            this.configuration.put(String.valueOf(ClickHouseClientOption.PROXY_PORT), String.valueOf(port));
+            this.configuration.put(String.valueOf(ClickHouseClientOption.PROXY_TYPE).toLowerCase(), type.toString());
```
chernser, line 388: `com.clickhouse.client.config.ClickHouseClientOption#getKey` should be used instead. It works now, but some config options doesn't match with enum constant name

chernser (approved):

PR #1694, opened 2024-06-18
[client-v2] Fix proxy configuration globally for the client

chernser:
## Summary
Proxy configuration is copied to client used by operations. 

## Checklist


mzitnik (approved):
Approving but we still need tests for it @Paultagoras can you add

PR #1699, opened 2024-06-20
Adding query_id to V1 client

Paultagoras:
## Checklist


mzitnik (changes requested):
Can we have a real test that do actual work 
We can set the query_id and check if we get the same one on the response

mzitnik (approved):
Approving, but let also add it in our new API as separate PR

PR #1701, opened 2024-06-20
Adding query_id to V2 client

Paultagoras:
## Checklist


client-v2/src/main/java/com/clickhouse/client/api/internal/GenericMetric.java
```diff
@@ -0,0 +1,26 @@
+package com.clickhouse.client.api.internal;
+
+import com.clickhouse.client.api.metrics.Metric;
+
+public class GenericMetric implements Metric {
```
chernser, line 5: Metric is intended to be something measurable. String is a constant. It may act as a label.

client-v2/src/main/java/com/clickhouse/client/api/insert/InsertResponse.java
```diff
@@ -85,4 +85,12 @@ public long getServerTime() {
     public long getResultRows() {
         return operationMetrics.getMetric(ServerMetrics.RESULT_ROWS).getLong();
     }
+
+    /**
+     * Alias for {@link ServerMetrics#QUERY_ID}
+     * @return number of returned bytes
+     */
+    public String getQueryId() {
+        return operationMetrics.getMetric(ServerMetrics.QUERY_ID).getString();
```
chernser, line 94: It is a good idea to put query ID close to metrics, because eventually it will be a metric label. 
Would you please add `labels` to operations metrics and put query id there?

client-v2/src/main/java/com/clickhouse/client/api/internal/Gauge.java
```diff
@@ -18,4 +18,9 @@ public void set(long value) {
     public long getLong() {
         return value;
     }
+
+    @Override
+    public String getString() {
```
chernser, line 23: Just want to note: 
this is the first sign that something wrong with `getString()` in interface: we have to adopt existing classes to a new version. If we extending interface in the future we should keep it backward compatible as much as possible.

chernser (approved):

chernser:
@Paultagoras Thank you for the change!

PR #1702, opened 2024-06-21
[client-v2] fixed passing properties to old client

chernser:
## Summary
Because not every property is passed to the config of underlying client, it was not possible to connect over SSL.


mzitnik (approved):

mzitnik:
@chernser, it seems like the tests are not running due to the first-step failure

chernser:
> @chernser, it seems like the tests are not running due to the first-step failure

Thank you! I've fixed it.

Issue #1529, opened 2024-01-23
Query with id is already running. (QUERY_WITH_SAME_ID_IS_ALREADY_RUNNING)


somemistake:
### This is my code
```java
IntStream
                    .range(0, config.getPartitionCount())
                    .mapToObj(PartitionUtils::getPartition)
                    .forEach( partition ->
                            namedParameterJdbcTemplate.execute(
                               "OPTIMIZE TABLE transaction_operation PARTITION :partition FINAL DEDUPLICATE",
                                new MapSqlParameterSource("partition", Integer.parseInt(partition)),
                                PreparedStatement::execute
                            );
                    );
```
### 
```java
Code: 216. DB::Exception: Query with id = 2b82eae6-e0f3-47dc-a857-2bb02dff7bbd is already running. (QUERY_WITH_SAME_ID_IS_ALREADY_RUNNING) (version 22.2.2.1)
```

driver version: com.clickhouse:clickhouse-jdbc:0.4.6

chernser:
I was able to reproduce the problem with multiple threads. Fixing this would need some refactoring.

chernser:
If statements are called from different threads then it is clear what happening: 
```
pool-1-thread-1: Executing select 1 queryID Optional[b48cef26-a612-4196-b779-df87e6a5a174] queryId: b48cef26-a612-4196-b779-df87e6a5a174
pool-1-thread-2: Executing select 1 queryID Optional[91bfa2ec-3767-4207-8f69-f4b5836512b5] queryId: 91bfa2ec-3767-4207-8f69-f4b5836512b5
pool-1-thread-3: Executing select 1 queryID Optional[dc94f299-92a6-487e-9dbb-d0943507352d] queryId: dc94f299-92a6-487e-9dbb-d0943507352d
pool-1-thread-2 sealing: 86939426533300 91bfa2ec-3767-4207-8f69-f4b5836512b5
pool-1-thread-1 sealing: 86939426523800 91bfa2ec-3767-4207-8f69-f4b5836512b5
pool-1-thread-3 sealing: 86939426576200 91bfa2ec-3767-4207-8f69-f4b5836512b5
pool-1-thread-1 error: Code: 216. DB::Exception: Query with id = 91bfa2ec-3767-4207-8f69-f4b5836512b5 is already running. (QUERY_WITH_SAME_ID_IS_ALREADY_RUNNING) (version 24.4.1.2088 (official build))

```
[...]

PR #1706, opened 2024-06-24
Fix duplicate query ID error 

chernser:
## Summary
In the JDBC driver every Statement has own request object, but it is shared among requests. If statement object is used by different threads then one thread may copy query_id set by another. Ignoring the fact that threads should not share Statements there should be a proper synchronization when working with shared request object. 
Current PR implements locking on request object while making copy (seal operation). 

Closes: 
- https://github.com/ClickHouse/clickhouse-java/issues/1529

## Checklist


chernser:
I've removed busy check from java11 HttpClientConnection because it doesn't make sense. This is similar to Apache connection but has busy check for some reason. I've inspected the code of the HttpClientConnection and it looks very simple and stateless - that means mostly thread-safe.

mzitnik (approved):

PR #1720, opened 2024-07-03
[clientk-v2] Fix async issues 

chernser:
## Summary
- Fixed handling async operations 
- Minimized number of threads when calling an old client
- Fixed passing custom settings to server (no more check if settings is enabled) 

## Checklist

### DIFF
--- a/client-v2/src/main/java/com/clickhouse/client/api/Client.java
+++ b/client-v2/src/main/java/com/clickhouse/client/api/Client.java
@@ -61,6 +61,7 @@ import java.util.concurrent.ExecutionException;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Executors;
 import java.util.concurrent.TimeUnit;
+import java.util.concurrent.TimeoutException;
 
 import static java.time.temporal.ChronoUnit.SECONDS;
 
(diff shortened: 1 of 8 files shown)


client-v2/src/main/java/com/clickhouse/client/api/Client.java
```diff
@@ -641,9 +641,11 @@ public CompletableFuture<InsertResponse> insert(String tableName,
 
             if (!responseFuture.isCompletedExceptionally()) {
                 try {
-                    InsertResponse response = new InsertResponse(client, future.get(), clientStats);
+                    InsertResponse response = new InsertResponse(client, future.get(TIMEOUT, TimeUnit.MILLISECONDS), clientStats);
```
mzitnik, line 644: should TIMEOUT also need to be configurable?

mzitnik (changes requested):

client-v2/src/main/java/com/clickhouse/client/api/Client.java
chernser: yes, but it would change with new client. So I have not done it. Will add.

chernser: Done.

mzitnik (approved):

PR #1721, opened 2024-07-06
Add retry on NoHttpResponseException

chernser:
## Summary
There is a problem with keep-alive connection. Sometimes a server closes it and a client knows nothing about it. Causes of happening it are different and it is complex to resolve with a single strategy. 
Apache HC throws `com.github.dockerjava.zerodep.shaded.org.apache.hc.core5.http.NoHttpResponseException` when request is sent over a "bad" connection. This PR will retry one more time immediately if `com.clickhouse.client.http.config.ClickHouseHttpOption#AHC_RETRY_ON_FAILURE` is enabled. 
However not every request may be retried safely. And here is another problem - it is extremely hard to detect what request is safe to retry. For example, it depends on if data stream is resettable in case of write operation. 

## Checklist


clickhouse-http-client/src/main/java/com/clickhouse/client/http/config/ClickHouseHttpOption.java
```diff
@@ -64,7 +65,22 @@ public enum ClickHouseHttpOption implements ClickHouseOption {
...
+     *     <li>{@code org.apache.hc.core5.http.ConnectionClosedException}</li>
+     *     <li>{@code org.apache.hc.core5.http.NoHttpResponseException}</li>
+     * </ul>
+     *
+     * And next status codes:
+     * <ul>
+     *     <li>{@code 503 Service Unavailable}</li>
+     * </ul>
+     */
+    AHC_RETRY_ON_FAILURE("ahc_retry_on_failure", false, "Whether to retry on failure with AsyncHttpClient.")
```
mzitnik, line 82: Can we change the name of the property?

chernser: Yes, sure. But let me explain why I've chosen this one - implementation works only for Apache HC and only for old client. 
In the new client I would name it `http.retry_on_failure` meaning it will be supported for every http implementation we ever have.

mzitnik (approved):

PR #1722, opened 2024-07-06
added settings for AHC to validate when stale

chernser:
## Summary
Apache HC has a great feature of checking for stale connection. Once this feature has been added to the our client but disappeared after some time due to refactoring (unfortunately it happens). 
Here is original PR https://github.com/ClickHouse/clickhouse-java/pull/760

New http client option is added to the `com.clickhouse.client.http.config.ClickHouseHttpOption`
```java 

    /**
     * The time in milliseconds after which the connection is validated after inactivity.
     * Default value is 5000 ms. If set to negative value, the connection is never validated.
     * It is used only for Apache Http Client connection provider.
     */
    AHC_VALIDATE_AFTER_INACTIVITY("ahc_validate_after_inactivity", 5000L,
            "The time in milliseconds after which the connection is validated after inactivity."),

```

## Checklist


mzitnik (approved):
Have you managed to reproduce it? Do you know how it works under the hood?

chernser:
> Have you managed to reproduce it? Do you know how it works under the hood?

Yes and no, I can reproduce same exception with wiremock but it is not a vanilla case. I mean it is not a 100% stale connection.

Issue #1729, opened 2024-07-09
instantiating a ClickHouseHttpClient takes almost a minute


vrischmann:
### Describe the bug

I'm currently testing the ClickHouse Java client on macOS and I think there's a bug.

Just instantiating a new ClickHouseHttpClient takes almost a minute on my system. The following code:
```java
        logger.info("Before client build");

        try (var client = new ClickHouseHttpClient()) {
            logger.info("Client created: {}", client);
        }

```

produces this:
```
07:47:30.336 [main] INFO com.batch.ebt.App -- Before client build
07:48:21.049 [main] INFO com.batch.ebt.App -- Client created: com.clickhouse.client.http.ClickHouseHttpClient@47fd17e3
```

I ran this code with a debugger, paused the execution and from what I can tell it is stuck doing a reverse-address lookup on a random IP address. The call to `getCanonicalHostname` in [getLocalhost](https://github.com/ClickHouse/clickhouse-java/blob/main/clickhouse-http-client/src/main/java/com/clickhouse/client/http/ClickHouseHttpClient.java#L45-L66) is what's blocking.

And here you can see the IP address it's calling `getCanonicalHostname` on: 
<img width="909" alt="Screenshot 2024-07-09 at 07 42 03" src="https://github.com/ClickHouse/clickhouse-java/assets/1916079/064eeea2-1112-46f0-a4c4-9bb92d22f718">

That IP comes from this interface but I have no clue what it's used for:
```
utun8: flags=8051<UP,POINTOPOINT,RUNNING,MULTICAST> mtu 1380
	inet6 fe80::88d:79aa:436e:dd72%utun8 prefixlen 64 scopeid 0x1a 
	nd6 options=201<PERFORMNUD,DAD>
```

I'm not sure what's the best way to resolve this because it seems specific to macOS, however it would be nice to have the ability to bypass this step and just provide this address ourselves somehow. Right now since it's executed in the ClickHouseHttpClient constructor we can't do anything.

### Steps to reproduce
1. Create a new basic main application, create a new instance of `ClickHouseHttpClient`
2. Run the code on macOS
3. Observe it takes a long time to finish

### Code example

See above

### Error log

Not applicable here.

#### Environment
* Client version: 0.6.2
* Language version: 22.0.1
* OS: macOS Sonoma 14.5

#### ClickHouse server
* ClickHouse Server version: version 24.5.4.49 (official build)

PR #1730, opened 2024-07-09
clickhouse-http-client: ignore link local addresses in getLocalhost

vrischmann:
## Summary

This a possible fix for #1729.

As far as I can tell link-local addresses are useless for the purpose of `getLocalhost` since it's used to get a referrer address, so I ignored them.

This solves my issue on my macOS system: for all others addresses the `getCanonicalHostName` is almost instantaneous.

Additionally, the comment says "use the first non-loopback address" but the code didn't actually do that because the break doesn't apply to the outer loop, I fixed that too.

## Checklist

### DIFF
--- a/clickhouse-http-client/src/main/java/com/clickhouse/client/http/ClickHouseHttpClient.java
+++ b/clickhouse-http-client/src/main/java/com/clickhouse/client/http/ClickHouseHttpClient.java
@@ -52,7 +52,7 @@ public class ClickHouseHttpClient extends AbstractClient<ClickHouseHttpConnectio
                 Enumeration<InetAddress> inetAddresses = ni.getInetAddresses();
                 for (InetAddress ia : Collections.list(inetAddresses)) {
                     // We just use the first non-loopback address
-                    if (!ia.isLoopbackAddress()) {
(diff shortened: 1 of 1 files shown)


chernser:
Good day, @vrischmann ! Thank you for the PR. We will review shortly.

mzitnik:
@vrischmann Thanks for the PR. Did you test it also on the Linux box? Maybe we should use it only macOS systems

vrischmann:
> @vrischmann Thanks for the PR. Did you test it also on the Linux box? Maybe we should use it only macOS systems

Sorry no, I didn't think of it.

chernser:
@vrischmann 
Would you be able to test on Linux, please?

vrischmann:
> @vrischmann 
> 
> Would you be able to test on Linux, please? 

Yes I can, I'll do it next week.

vrischmann:
Hi. I've tested the fix on a Debian server and it works fine, no blocking occurs.

chernser:
Thank you, @vrischmann! 
Will merge shortly

chernser (approved):

PR #1737, opened 2024-07-17
[client-v2] Implement Insert of POJOs with new HTTP implementation

chernser:
## Summary
 Implement Insert of POJOs with new HTTP implementation

### DIFF
--- a/client-v2/src/main/java/com/clickhouse/client/api/Client.java
+++ b/client-v2/src/main/java/com/clickhouse/client/api/Client.java
@@ -772,7 +772,7 @@ public class Client implements AutoCloseable {
                                          out -> {
                                              out.write("INSERT INTO ".getBytes());
                                              out.write(tableName.getBytes());
-                                             out.write(" \n FORMAT ".getBytes());
+                                             out.write(" FORMAT ".getBytes());
                                              out.write(format.name().getBytes());
                                              out.write(" \n".getBytes());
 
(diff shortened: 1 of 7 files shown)


client-v2/src/main/java/com/clickhouse/client/api/Client.java
```diff
@@ -746,47 +746,113 @@ public CompletableFuture<InsertResponse> insert(String tableName,
...
+                                         out -> {
+                                             out.write("INSERT INTO ".getBytes());
+                                             out.write(tableName.getBytes());
+                                             out.write(" \n FORMAT ".getBytes());
+                                             out.write(format.name().getBytes());
+                                             out.write(" \n".getBytes());
+
+                                             byte[] buffer = new byte[writeBufferSize];
+                                             int bytesRead;
+                                             while ((bytesRead = data.read(buffer)) != -1) {
```
mzitnik, line 780: If we try to retry, do not we already exhaust the stream by doing `data.read`?

chernser: Yes, this is correct. I need to try resetting the input stream. 
Will correct and create a test. 
However there is a story about retries and something smarter should be implemented
- check connection before use 
- cleanup connections by ttl

client-v2/src/main/java/com/clickhouse/client/api/Client.java
```diff
@@ -683,47 +747,122 @@ public CompletableFuture<InsertResponse> insert(String tableName,
...
+                // Selecting some node
+                ClickHouseNode selectedNode = getNextAliveNode();
+
+                for (int i = 0; i <= maxRetries; i++) {
+                    // Execute request
+                    try (ClassicHttpResponse httpResponse =
+                                 httpClientHelper.executeRequest(selectedNode, finalSettings.getAllSettings(),
+                                         out -> {
+                                             out.write("INSERT INTO ".getBytes());
+                                             out.write(tableName.getBytes());
```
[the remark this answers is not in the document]
chernser: Hmm.. this is actually how Mutation works today. 
I will add it to the header.

PR #1748, opened 2024-07-25
[client-v2] Proxy support

chernser:
## Summary
- Added forwarding proxy configuration to new http layer 
- Added auth configuration for proxy
- Added ability to disable storing cookies

## Checklist

### DIFF
--- a/client-v2/src/main/java/com/clickhouse/client/api/Client.java
+++ b/client-v2/src/main/java/com/clickhouse/client/api/Client.java
@@ -424,6 +424,12 @@ public class Client implements AutoCloseable {
             return this;
         }
 
+        public Builder setProxyCredentials(String user, String pass) {
+            this.configuration.put("proxy_user", user);
+            this.configuration.put("proxy_password", pass);
+            return this;
+        }
+
         /**
          * Sets the maximum time for operation to complete. By default, it is set to 3 hours.
          * @param timeout
@@ -963,6 +969,8 @@ public class Client implements AutoCloseable {
                         metrics.operationComplete();
 
                         return new QueryResponse(httpResponse, finalSettings, metrics);
+                    } catch (ClientException e) {
+                        throw e;
                     } catch (Exception e) {
                         throw new ClientException("Failed to execute query", e);
                     }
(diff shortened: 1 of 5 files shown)


client-v2/src/main/java/com/clickhouse/client/api/Client.java
```diff
@@ -444,6 +450,12 @@ public Builder useNewImplementation(boolean useNewImplementation) {
             return this;
         }
 
+        public Builder setHttpCookiesEnabled(boolean enabled) {
+            //TODO: extract to settings string constants
+            this.configuration.put("client.http.cookies_enabled", String.valueOf(enabled));
```
mzitnik, line 455: What is that used for?

client-v2/src/main/java/com/clickhouse/client/api/internal/HttpAPIClientHelper.java
```diff
@@ -50,25 +55,48 @@ public class HttpAPIClientHelper {
...
     private RequestConfig baseRequestConfig;
 
+    private String proxyAuthHeaderValue;
+
     public HttpAPIClientHelper(Map<String, String> configuration) {
         this.chConfiguration = configuration;
-        this.httpClient = createHttpClient(configuration, null);
+        this.httpClient = createHttpClient();
         this.baseRequestConfig = RequestConfig.custom()
                 .setConnectionRequestTimeout(1000, TimeUnit.MILLISECONDS)
```
mzitnik, line 64: Should it be configurable?

mzitnik (approved):

client-v2/src/main/java/com/clickhouse/client/api/Client.java
chernser: - sticky sessions 
- some enterprise proxies like F5 that store routing information in cookies

chernser: actually there may be opposite case when cookies should not be preserved to avoid sticky session when proxy behavior may not be changed.

client-v2/src/main/java/com/clickhouse/client/api/internal/HttpAPIClientHelper.java
chernser: Yes. Added more configuration parameters. Planning to add more later.

Issue #1717, opened 2024-07-01
[client-v2] Compression 


chernser:
## Compression 

Compression may be on transport or application layer. For example, HTTP compressed encoding is transport layer and LZ4 payload compression is what application does. 
Currently application compression is what we need to implement. First implementation should support bi-directional LZ4 encoding. 

Note 1: There is a satisfying implementation of LZ4 support but it uses own stream classes. Should be adopted for broader use. 

### Settings 

Next settings should be deprecated: 
- com.clickhouse.client.config.ClickHouseClientOption#COMPRESS "Whether the server will compress response it sends to client." - currently means server response compression and it should be named accordingly
- com.clickhouse.client.config.ClickHouseClientOption#DECOMPRESS "Whether the server will decompress request from client." - currently means client request compression and it should be named accordingly
- com.clickhouse.client.config.ClickHouseClientOption#COMPRESS_ALGORITHM "Algorithm used for server to compress response." - currently only LZ4 is used and this setting can be dropped. 
- com.clickhouse.client.config.ClickHouseClientOption#DECOMPRESS_ALGORITHM "Algorithm for server to decompress request." - currently only LZ4 is used and this setting can be dropped.
- com.clickhouse.client.config.ClickHouseClientOption#COMPRESS_LEVEL "Compression level for response, -1 standards for default" - currently means server response compression level and it should be named accordingly. But may be dropped. com.clickhouse.client.config.ClickHouseClientOption#DECOMPRESS_LEVEL "Compression level for request, -1 standards for default" - currently means client request compression level and it should be named accordingly. But may be dropped.

### Metrics 

Not yet

mshustov:
> com.clickhouse.client.config.ClickHouseClientOption#COMPRESS "Whether the server will compress response it sends to client." - currently means server response compression and it should be named accordingly
> com.clickhouse.client.config.ClickHouseClientOption#DECOMPRESS "Whether the server will decompress request from client." - currently means client request compression and it should be named accordingly

It's mirroring of ClickHouse settings https://clickhouse.com/docs/en/interfaces/http

> com.clickhouse.client.config.ClickHouseClientOption#COMPRESS_ALGORITHM "Algorithm used for server to compress response." - currently only LZ4 is used and this setting can be dropped.
> com.clickhouse.client.config.ClickHouseClientOption#DECOMPRESS_ALGORITHM "Algorithm for server to decompress request." - currently only LZ4 is used and this setting can be dropped.

[...]

PR #1761, opened 2024-08-02
[client-v2] Compression Support

chernser:
## Summary

Implements support for next compression scenarios:
- whole http response from a server is compressed with LZ4
    - enabled when server &  http compression is enabled 
- response body from a server is compressed with LZ4
- http request is compressed with LZ4 
- request body is compressed with LZ4  

Closes https://github.com/ClickHouse/clickhouse-java/issues/1717

### DIFF
--- a/client-v2/src/main/java/com/clickhouse/client/api/internal/ClickHouseLZ4InputStream.java
+++ b/client-v2/src/main/java/com/clickhouse/client/api/internal/ClickHouseLZ4InputStream.java
@@ -3,6 +3,7 @@ package com.clickhouse.client.api.internal;
 import com.clickhouse.client.api.ClientException;
 import com.clickhouse.data.ClickHouseByteUtils;
 import com.clickhouse.data.ClickHouseCityHash;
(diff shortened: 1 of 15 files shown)


client-v2/src/main/java/com/clickhouse/client/api/internal/ClickHouseLZ4InputStream.java
```diff
@@ -0,0 +1,175 @@
...
+    private ByteBuffer buffer;
+
+    private byte[] tmpBuffer = new byte[1];
+
+
+    public ClickHouseLZ4InputStream(InputStream in, LZ4FastDecompressor decompressor) {
+        super();
+        this.decompressor = decompressor;
+        this.in = in;
+        this.buffer = ByteBuffer.allocate(8192);
```
mzitnik, line 33: Let's extract it to const & make configurable

mzitnik (approved):
In general LGTM, but we need to add tests that can mimic different chunk sizes to simulate edge cases for the compression.

Issue #1692, opened 2024-06-17
[client-v2] Retries


chernser:
## Retries

Request to a server may fail and then retried depending on a reason. For example, when connection timeouted it may be retried with same target or next another one. 
Policy for retries is application configuration and Java Client should be very flexible in way automatic retries can be configured. 

Each retry attempt consumes some time and it may be a problem. For example, application sets maximum execution time and retry attempt take 90% of this limit. In such
case there is no benefit doing a retry attempt. Therefore retry logic should respect any configured time limitations.   

This is client feature and feature should be guaranteed by client. If underlying transport has retries mechanism it make sense to limit its use.  

### Settings 
 
- com.clickhouse.client.config.ClickHouseClientOption#RETRY "Maximum number of times retry can happen for a request, zero or negative value means no retry."
    This setting should be deprecated because it is has too broad meaning. Instead new should be created:
    - client.max_connection_retries - Maximum number of times a request can be retried if connection initiation error happens. 

- com.clickhouse.client.config.ClickHouseClientOption#MAX_EXECUTION_TIME "Maximum query execution time in seconds, 0 means no limit."
    This configuration setting should be respected when checking if retry is needed. 

- com.clickhouse.client.config.ClickHouseClientOption#REPEAT_ON_SESSION_LOCK "Whether to repeat execution when session is locked, until timed out(according to 'session_timeout' or 'connect_timeout')." 
    This options should be kept.

### Metrics 

- number_of_retries - counter of number of retries per operation type.

mzitnik:
@chernser i would like also impact the old client api

PR #1768, opened 2024-08-10
[client-v2] Retries on failures

chernser:
## Summary

Closes https://github.com/ClickHouse/clickhouse-java/issues/1692
## Checklist


client-v2/src/main/java/com/clickhouse/client/api/Client.java
```diff
@@ -776,65 +801,69 @@ public Client build() {
...
 
-            if (!userConfig.containsKey("connection_ttl")) {
-                userConfig.put("connection_ttl", "-1");
+            if (!configuration.containsKey("client_retry_on_failures")) {
+                retryOnFailures(ClientFaultCause.NoHttpResponse, ClientFaultCause.ConnectTimeout, ClientFaultCause.ConnectionRequestTimeout);
             }
 
-            if (!userConfig.containsKey("client_network_buffer_size")) {
+            if (!configuration.containsKey("client_network_buffer_size")) {
                 setClientNetworkBufferSize(8192);
```
mzitnik, line 861: lets add it as const

client-v2/src/main/java/com/clickhouse/client/api/Client.java
```diff
@@ -1166,25 +1199,27 @@ public CompletableFuture<InsertResponse> insert(String tableName,
...
                         return new InsertResponse(metrics);
-                    } catch (NoHttpResponseException e) {
-                        if (i < maxRetries) {
-                            try {
-                                data.reset();
-                            } catch (IOException ioe) {
-                                throw new ClientException("Failed to get response", e);
-                            }
-                            LOG.warn("Failed to get response. Retrying.", e);
+                    } catch ( NoHttpResponseException | ConnectionRequestTimeoutException | ConnectTimeoutException e) {
```
mzitnik, line 1202: Is it a duplicated code? (i see in other inserts as well)

mzitnik (approved):

client-v2/src/main/java/com/clickhouse/client/api/Client.java
chernser: done.

client-v2/src/main/java/com/clickhouse/client/api/Client.java
chernser: yes. I left it as duplicate because handling it in dedicated method would not be much shorter. I will review this code later if it can be extracted to a single method.

Issue #1770, opened 2024-08-12
[clickhouse-http-client] Extend Apache HTTP Client Connection Pool Config


chernser:
### Describe the bug
Connection pool controls how new connections are created and reused. It affects some common scenarios:
- Highly concurrent application - connection requests are blocking execution for too long 
- Clusters with many nodes - some nodes get less requests because most requests are sent thru established connection to other nodes. 

How to solve:
- it should be possible to configure TTL (or Keep-Alive) of a connection 
- it should be possible to configure connection request timeout to avoid waiting for potential free connection but fail fast 
- it should be possible to configure connection pool strategy (FIFO vs LIFO)

PR #1771, opened 2024-08-15
Connection pool config

chernser:
## Summary
Added new options:
- com.clickhouse.client.config.ClickHouseClientOption#CONNECTION_TTL - sets maximum connection time to leave. Overrides keep-alive from a server
- com.clickhouse.client.http.config.ClickHouseHttpOption#KEEP_ALIVE_TIMEOUT - sets keep alive for connections. Overrides keep-alive from a server 
- com.clickhouse.client.http.config.ClickHouseHttpOption#CONNECTION_REUSE_STRATEGY - sets strategy of connection reuse. By default LIFO is used so most fresh connections are used. Another value is "FIFO" that will make connection pool to use connections in fair way. 

Closes https://github.com/ClickHouse/clickhouse-java/issues/1770

## Checklist


clickhouse-http-client/src/main/java/com/clickhouse/client/http/ApacheHttpConnectionImpl.java
```diff
@@ -91,14 +96,31 @@ private CloseableHttpClient newConnection(ClickHouseConfig c) throws IOException
...
+                throw new IllegalArgumentException("Invalid connection reuse strategy: " + poolReuseStrategy);
+            }
+        }
+        log.info("Connection reuse strategy: %s", poolReusePolicy.name());
+        HttpConnectionManager connManager = new HttpConnectionManager(r.build(), c, PoolConcurrencyPolicy.LAX,
+                poolReusePolicy, TimeValue.ofMilliseconds(connectionTTL));
         int maxConnection = config.getIntOption(ClickHouseHttpOption.MAX_OPEN_CONNECTIONS);
 
-        connManager.setMaxTotal(maxConnection);
+        connManager.setMaxTotal(Integer.MAX_VALUE); // unlimited on global level
```
mzitnik, line 115: Can you explain why setting into MAX_VALUE

chernser: if we have different routes then it will override max connection per route
previous version would work with only one route well

PR #1794, opened 2024-08-27
[client-v2] POJO reader

chernser:
## Summary
In most cases data is read into DTO objects - custom application POJOs. When dataset is big it may be a significant overhead if data is read into internal storage or buffer then copied to a DTO. This PR implements reader using similar to `insert` API approach:
- class is registered and set of deserializers is generated
- each deserializer reads a single field and writes data to POJO

## Checklist


chernser:
POJO reader is promising if we would eliminate reflection calls. 
I've compared to the "classic" approach thru binary ready and it faster then POJO + objects pool. 
May be something is hiding from me.

mzitnik (approved):

PR #1795, opened 2024-08-28
[client-v2] Fixed reusing map without creating new nodes

chernser:
## Summary
Internally data is read into a map that represents a record. `clear()` was called each time before reading data to protect from stale data from a previous records. However it is not needed because it is always same number of columns are read. Only case with null should be handled correctly to clean up non-null values.

This single change reduces number of Map$Node object by magnitude. 

## Checklist


chernser:
Before fix:
top_classes_50max_connections

After fix:
top_classes_100max_connections_fix

mzitnik (approved):

PR #1797, opened 2024-08-29
[jdbc-refactor-main] Removing some outdated classes/interfaces

Paultagoras:
This is to convert the JDBC driver over, but it's still very much a WIP (until we convert it to a normal PR).


clickhouse-jdbc/src/main/java/com/clickhouse/jdbc/DriverImpl.java
chernser, line 1: lets name it just `Driver`. I see it is name pattern across some JDBC drivers.

chernser (approved):

PR #1803, opened 2024-09-03
Use debug logging for internal connection details

alxhill:
## Summary
Currently, connection TTL & queue strategy are logged at INFO level for every request. This can result in a lot of noise in logs:

Screenshot 2024-09-03 at 11 00 56 AM

This PR switches them to DEBUG log levels.

## Checklist

### DIFF
--- a/clickhouse-http-client/src/main/java/com/clickhouse/client/http/ApacheHttpConnectionImpl.java
+++ b/clickhouse-http-client/src/main/java/com/clickhouse/client/http/ApacheHttpConnectionImpl.java
@@ -95,7 +95,7 @@ public class ApacheHttpConnectionImpl extends ClickHouseHttpConnection {
         }
 
         long connectionTTL = config.getLongOption(ClickHouseClientOption.CONNECTION_TTL);
-        log.info("Connection TTL: %d ms", connectionTTL);
+        log.debug("Connection TTL: %d ms", connectionTTL);
         String poolReuseStrategy = c.getStrOption(ClickHouseHttpOption.CONNECTION_REUSE_STRATEGY);
         PoolReusePolicy poolReusePolicy = PoolReusePolicy.LIFO;
         if (poolReuseStrategy != null && !poolReuseStrategy.isEmpty()) {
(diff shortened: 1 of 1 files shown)


chernser:
Good day, @alxhill ! 
Thank you for reporting! 
There may be another problem. Would you please share your settings? 

It was expected that: 
- `ApacheHttpClientConnection` is reused many times (do you create a new client each request) 
- 0 ms setting seems not optimal because internal apache client connection  will not be reused.

alxhill:
I think the TTL being set to zero should mean unlimited TTL:

https://github.com/ClickHouse/clickhouse-java/blob/12f5edf4fb77f82bfefb5063273ad646ae6144b1/clickhouse-client/src/main/java/com/clickhouse/client/config/ClickHouseClientOption.java#L435-L437

Our code did used to re-use connections a lot but doesn't as much any more. It's possible testcontainers is the primary culprit - I made a minimal repro that creates a Client and runs the test query and it had a lot of logging:

```java
@Testcontainers
@ExtendWith(MockitoExtension.class)
class DemoTestContainersTest {

    @Container
    private final ClickHouseContainer clickHouseContainer = new ClickHouseContainer(ClickHouseConstants.IMAGE);

    private Client client;

    @BeforeEach
    void before() {
        client = new Client.Builder()
                .addEndpoint(
                        Protocol.HTTP, clickHouseContainer.getHost(), clickHouseContainer.getFirstMappedPort(), false)
                .setUsername(clickHouseContainer.getUsername())
                .setPassword(clickHouseContainer.getPassword())
                .compressServerResponse(true)
                .setSocketTimeout(30, ChronoUnit.SECONDS) // 30 seconds. Default is 2 seconds.
                .build();
    }

    @Test
    void connected() throws Exception {
        System.out.println("starting test query");
        try (QueryResponse response =
                client.query(clickHouseContainer.getTestQueryString()).join()) {
            System.out.println("response: " + response);
        }
        System.out.println("finished test query");
    }
}
```

Output logs here: [ch-connection.log](https://github.com/user-attachments/files/16853480/ch-connection.log)

chernser:
@alxhill thank you for the test! 

You seems using old version of the client-v2 (internally it uses old one but now has a new implementation). 
Latest version is 0.6.5 - it will utilize connection reuse. 
Also you may switch to the new implementation byte setting `com.clickhouse.client.api.Client.Builder#useNewImplementation` (method is deprecated because soon it will be only one internal implementation)

alxhill:
Those logs are using the `0.6.5` jar fwiw. Our codebase is _mostly_ client v2, but there's a few uses of client v1 floating around that will be producing excess logs too. As a sidenote, I did enable the `useNewImplementation` flag but it's causing some breaks - for example `date_time_input_format='best_effort'` doesn't successfully convert timestamps that were previously being parsed into `DateTime64`. Is GitHub issues a good place to report these?

Separate from fixing up our code, would still encourage switching this logline to `DEBUG` as it doesn't seem important enough to show up as `INFO` for all consumers

chernser:
@alxhill 
I see now - there is a lot records because client is called from multiple threads. 
Please create an issue with all feature you would like to see asap in client-v2 (list them all in one issue - I will split if needed).

chernser (approved):

PR #1814, opened 2024-09-08
Including support for bitmap serialization

javiercj93:
## Summary
I have a table structure which has a few aggregate function columns (groupBitmap) and I was unable to directly insert into it as the ClickhouseBitmapColumn was throwing an unsupported data type when serializing.

Checking the code I saw there was a `writeBitmap` function implemented in the BinaryStreamUtils but it was not being used in the SerializerUtils. Not sure if there was any reason behind it or if there are any other aggregate functions that could be serialized as well. 

- I did not add any unit tests as this class did not have them yet (there is no much logic in it but a wrapper to serialize)  but let me know fi that is a requirement and I'll add unit tests to cover the changes.

### DIFF
--- a/client-v2/src/main/java/com/clickhouse/client/api/internal/SerializerUtils.java
+++ b/client-v2/src/main/java/com/clickhouse/client/api/internal/SerializerUtils.java
@@ -2,8 +2,11 @@ package com.clickhouse.client.api.internal;
 
 import com.clickhouse.client.api.Client;
 import com.clickhouse.client.api.ClientFaultCause;
+import com.clickhouse.data.ClickHouseAggregateFunction;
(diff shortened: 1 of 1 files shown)


javiercj93:
> Hi @javiercj93 thanks for submitting this! Would you mind adding a test to cover this data as well?

Thanks for reviewing @Paultagoras! appreciate it! Please let me know if this change in an existing test would be enough or we need something more specific.

chernser:
@javiercj93 would you please resolve conflicts in the merge? 
I would do it with web editor but I need a permission to commit merge to your fork. 
Thanks in advance!

chernser:
@javiercj93 appreciate it! Thank you!

chernser (approved):

PR #1816, opened 2024-09-09
[client-v2] reader buffer optimization

chernser:
## Summary
A lot small buffers may be allocated while reading numeric information. 
This PR implements a reusing buffers withing a reader. It is not a default behavior. 
Introduced `Client.Builder#allowBinaryReaderToReuseBuffers` to control the behavior 

Note: 
- implementation is not thread-safe 
- reader should be called by one thread at a time
- small buffers will 64 bytes are pre-allocated 

## Checklist

### DIFF
--- a/client-v2/src/main/java/com/clickhouse/client/api/data_formats/internal/BinaryStreamReader.java
+++ b/client-v2/src/main/java/com/clickhouse/client/api/data_formats/internal/BinaryStreamReader.java
@@ -27,6 +27,11 @@ import java.util.Map;
 import java.util.TimeZone;
 import java.util.UUID;
 
+/**
+ * This class is not thread safe and should not be shared between multiple threads.
+ * Internally it may use a shared buffer to read data from the input stream.
+ * It is done mainly to reduce extra memory allocations for reading numbers.
+ */
 public class BinaryStreamReader {
 
     private final InputStream input;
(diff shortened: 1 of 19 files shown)


client-v2/src/main/java/com/clickhouse/client/api/data_formats/internal/BinaryStreamReader.java
```diff
@@ -543,4 +673,42 @@ private static int readByteOrEOF(InputStream input) throws IOException {
...
+           }
+        }
+
+        @Override
+        public byte[] allocate(int size) {
+            if (size < preallocated.length) {
+                return preallocated[size];
+            }
+
+            return new byte[size];
```
mzitnik, line 711: In what cases can the size bigger larger than `MAX_PREALLOCATED_SIZE` + 1?
if it meets the case, it always allocates byte[size]. Do we want to allocate and cache the array for future usage?

mzitnik: looking on `return readNBytesLE(input, bufferAllocator.allocate(len), 0, len);` this the only place that we can hit 
`return new byte[size];` Maybe we also need to reuse it? I think if we have a table with a lot of strings bigger than 32 we actually will allocate every time and since we use it between rows I think the change that we hit same size allocation is high.

mzitnik (commented):
I do not see tests that enable the flag `client_allow_binary_reader_to_reuse_buffers`

client-v2/src/main/java/com/clickhouse/client/api/data_formats/internal/BinaryStreamReader.java
chernser: This implementation is to solve only small buffer allocations for numbers. 
As for Strings, we may be need another approach. If string is small - we can cache them, if big - we would need to represent them with another class (`String` copies source buffer) but it will be confusing.

chernser:
> I do not see tests that enable the flag `client_allow_binary_reader_to_reuse_buffers`

I've added `client-v2/src/test/java/com/clickhouse/client/query/BinaryReadyReusesBuffersTests.java` that overrides `QueryTests` and sets `client_allow_binary_reader_to_reuse_buffers`

client-v2/src/main/java/com/clickhouse/client/api/data_formats/internal/BinaryStreamReader.java
mzitnik: This what i had in my mind we can do it later

mzitnik (approved):

PR #1826, opened 2024-09-17
[client-v2] primitive reading into pojo

chernser:
## Summary
Current implementation reads all values into objects even it can be done using primitive types. It was done to simplify code, but it may be not good for memory consumption.
Current PR:
- adds methods to read values into primitives 
- updates POJO deserialization code to use these new methods

## Checklist

### DIFF
--- a/client-v2/src/test/java/com/clickhouse/client/query/QueryTests.java
+++ b/client-v2/src/test/java/com/clickhouse/client/query/QueryTests.java
@@ -31,7 +31,6 @@ import com.clickhouse.data.ClickHouseFormat;
 import com.fasterxml.jackson.databind.JsonNode;
 import com.fasterxml.jackson.databind.MappingIterator;
 import com.fasterxml.jackson.databind.ObjectMapper;
-import org.testcontainers.shaded.com.google.common.collect.Table;
 import org.testng.Assert;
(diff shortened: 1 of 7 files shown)


client-v2/src/test/java/com/clickhouse/client/query/QueryTests.java
```diff
@@ -91,6 +89,10 @@ public class QueryTests extends BaseIntegrationTest {
 
     private boolean usePreallocatedBuffers = false;
 
+    static {
+//        System.setProperty("org.slf4j.simpleLogger.defaultLogLevel", "DEBUG");
```
mzitnik, line 93: Lets remove it

mzitnik:
Do we support POJO with types of Long, Integer, Short, and more class

client-v2/src/test/java/com/clickhouse/client/query/QueryTests.java
chernser: ok. this is only for current development. but will move into env settings on local

chernser:
> Do we support POJO with types of Long, Integer, Short, and more class

We do but with 1-to-1 match. But with primitives it should be easier to handle. 
I will update this part.

mzitnik (approved):

PR #1830, opened 2024-09-19
[client-v2] Tests and Fixes in scope of SerDe

chernser:
## Summary
- Added more data type tests 
- Fixed read error from response 
- Added possibility to register class against different tables. 

## Checklist


chernser:
I've found that reading AggregationFunction seems has some bug in reading code what causes reading far after real stream ended. I will work in this issue later. Currently it is only for writing.

Issue #1782, opened 2024-08-21
[client-v2] Custom http header, query parameters and URL parsing


chernser:
### Describe the bug

It should be possible to pass custom HTTP headers to the request 

Closes https://github.com/ClickHouse/clickhouse-java/issues/1750

chernser:
I would propose next solution: 
- create a client builder method: `setHttpHeaders(Map<String, String> headers)` and accept values as `String` in format supporting multiple values
- reserve `http_h_` prefix for options that should be passed as header values. This is required for cases when we parse connection URL into properties 
- create a client builder method: `setQueryOptions(Map<String, String> options)` to be passed as query parameters
- reserve `query_p_` prefix for options that should be passes as query parameters. This is required for cases when we parse connection URL into properties.

Issue #1750, opened 2024-07-27
input_format_null_as_default flag not available as a valid connection query param 


thiagohora:
### Describe your feedback

I have a use case where I would like to set this property `input_format_null_as_default` to `false`; however, as it is not a valid query param option, I would need to execute it in every request.

thiagohora:
I think so. However, typically, frameworks abstract the instantiation of the client/connection factory. Just a suggestion, but I believe this should be supported via connection string as a queryparam

mshustov:
> Just a suggestion, but I believe this should be supported via connection string as a queryparam

@chernser, that's a valid point. Other language clients support this method.
[python client](https://github.com/ClickHouse/clickhouse-connect/blob/01dcaf7c3494614e6c773ee10f744ff475d5b26e/clickhouse_connect/driver/__init__.py#L104)
[nodejs client](https://github.com/ClickHouse/clickhouse-js/pull/232)

@chernser  Once it's implemented, could you also update [the language client spec](https://docs.google.com/document/d/1924Dvy79KXIhfqKpi1EBVY3133pIdoMwgCQtZ-uhEKs/edit#heading=h.ah33hoz5xei2) with this requirement?

chernser:
@mshustov will do and will update the language spec.

Issue #1665, opened 2024-05-31
clickhouse-jdbc driver did not support connection url parameters


ldcsaa:
### Describe the bug

Clickhouse JDBC connection URL with parameters would be always uses `default` database, like this:
```yaml
# the default database is NOT `my_db`, but the `default` db.
url: jdbc:clickhouse://192.168.56.23:8123/my_db?custom_http_params=async_insert=1,wait_for_async_insert=1
```
If cut the url parameter, that's OK, like this:
```
# the default database is `my_db`
url: jdbc:clickhouse://192.168.56.23:8123/my_db
```

- I had test many version of clickhouse JDBC driver, 0.4.6, 0.5.0, 0.6.0 etc.
- ClickHouse Server version: 24.1

chernser:
Good day, @ldcsaa! Thank you for reporting the issue! 
Am I correct that there are two problems:
- when query parameters are set then wrong database is selected
- you would like to pass some parameters thru JDBC url

for the last one - have you tried to url encode this part - `async_insert=1,wait_for_async_insert=1` ?

ldcsaa:
> Good day, @ldcsaa! Thank you for reporting the issue! Am I correct that there are two problems:
> 
> * when query parameters are set then wrong database is selected
> * you would like to pass some parameters thru JDBC url
> 
> for the last one - have you tried to url encode this part - `async_insert=1,wait_for_async_insert=1` ?

I did not tried to encode URL, just refer to this guide:
[Asynchronous Inserts](https://clickhouse.com/docs/en/cloud/bestpractices/asynchronous-inserts)

![image](https://github.com/ClickHouse/clickhouse-java/assets/5990337/a84c5fc9-549c-49fc-bab4-1bbbbce6f6da)

chernser:
# Triage Report 
- Bug and should work as documented.

ningObito:
There is a version that solves this problem ?

chernser:
Good day, @ldcsaa! 
I've found the bug that causes the problem - `,` is used as delimiter for endpoints definitions (`http://host1,host2,host3/my_db`). This will be fixed in 0.7.0. 
However, `custom_settings` should be used instead `custom_http_params` (should not be used because depends on protocol and actually is not working as expected.)

wtao54101:
[...]

PR #1841, opened 2024-09-27
[client-v2] Added options for additional headers and server settings

chernser:
## Summary
Adds two new configuration options for client and operations:
- `httpHeader` - to set additional http headers or override already set one's 
- `serverSetting` - to pass settings to a server (thru query parameters in case of HTTP) 

Closes https://github.com/ClickHouse/clickhouse-java/issues/1782
Closes https://github.com/ClickHouse/clickhouse-java/issues/1750
Closes https://github.com/ClickHouse/clickhouse-java/issues/1665
## Checklist

### DIFF
--- a/clickhouse-client/src/main/java/com/clickhouse/client/ClickHouseNodes.java
+++ b/clickhouse-client/src/main/java/com/clickhouse/client/ClickHouseNodes.java
@@ -68,6 +68,7 @@ public class ClickHouseNodes implements ClickHouseNodeManager {
         } else {
             index = 0;
         }
+        
 
         String defaultParams = "";
         Set<String> list = new LinkedHashSet<>();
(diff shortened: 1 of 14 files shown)


clickhouse-client/src/test/java/com/clickhouse/client/ClickHouseNodeTest.java
```diff
@@ -438,4 +440,34 @@ public void testToUri() {
...
+                Assert.assertEquals(settingsMap.get("param1"), "value1");
+                Assert.assertEquals(settingsMap.get("param2"), "value2");
+            });
+        }
+    }
+
+    @DataProvider(name = "testPropertyWithValueList_endpoints")
+    public static Object[][] endpoints() {
+        return new Object[][] {
+            { "http://server1:9090/my_db?custom_settings=param1=value1,param2=value2", 1, new String[]{"http://server1:9090/"} },
```
mzitnik, line 467: Is this a valid URL?

clickhouse-client/src/main/java/com/clickhouse/client/ClickHouseNodes.java
```diff
@@ -95,14 +96,19 @@ static ClickHouseNodes create(String endpoints, Map<?, ?> defaultOptions) {
             }
 
             int endIndex = i;
+            // parsing host name
```
mzitnik, line 99: I do not fully understand if we are planning to remove this module. 
why we are changing `ClickHouseNodes`

clickhouse-client/src/test/java/com/clickhouse/client/ClickHouseNodeTest.java
chernser: it is not exactly URL - it is client endpoints spec. It may have multiple hosts for example. 
Current example has custom setting with multiple key-value pairs (what would be URL encoded to make it Web URL, but we do not need to do this until we send request - so it is all fine)

mzitnik (approved):

PR #1854, opened 2024-10-08
Executing SQL Queries

Paultagoras:
## Checklist

### DIFF
--- a/clickhouse-jdbc/src/main/java/com/clickhouse/jdbc/ResultSetImpl.java
+++ b/clickhouse-jdbc/src/main/java/com/clickhouse/jdbc/ResultSetImpl.java
@@ -134,6 +134,7 @@ public class ResultSetImpl implements ResultSet, JdbcWrapper {
     @Override
     public Date getDate(int columnIndex) throws SQLException {
         checkClosed();
+        //TODO: Add this to ClickHouseBinaryFormatReader
         LocalDate localDate = reader.getLocalDate(columnIndex);
         if(localDate == null) {
             return null;
@@ -164,7 +165,8 @@ public class ResultSetImpl implements ResultSet, JdbcWrapper {
     @Override
     public InputStream getAsciiStream(int columnIndex) throws SQLException {
         checkClosed();
-        return new ByteArrayInputStream(reader.getString(columnIndex).getBytes(StandardCharsets.UTF_8));
+        //TODO: Add this to ClickHouseBinaryFormatReader
+        throw new SQLFeatureNotSupportedException("AsciiStream is not yet supported.");
     }
 
     @Override
@@ -243,6 +245,7 @@ public class ResultSetImpl implements ResultSet, JdbcWrapper {
     @Override
     public Date getDate(String columnLabel) throws SQLException {
         checkClosed();
+        //TODO: Add this to ClickHouseBinaryFormatReader
         LocalDate localDate = reader.getLocalDate(columnLabel);
         if(localDate == null) {
             return null;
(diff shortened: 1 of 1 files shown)


clickhouse-jdbc/src/main/resources/META-INF/services/java.sql.Driver
```diff
@@ -1 +1 @@
-com.clickhouse.jdbc.ClickHouseDriver
```
chernser, line 1: Do we need this change?

clickhouse-jdbc/src/main/java/com/clickhouse/jdbc/ConnectionImpl.java
```diff
@@ -13,8 +13,8 @@
 public class ConnectionImpl implements Connection, JdbcWrapper {
     private static final Logger log = LoggerFactory.getLogger(ConnectionImpl.class);
 
-    private final Client client;
-    private final JdbcConfiguration config;
+    protected final Client client;
```
chernser, line 16: Do we need to have it protected?

chernser: because we need to see them in the same package.

clickhouse-jdbc/src/main/java/com/clickhouse/jdbc/ResultSetImpl.java
```diff
@@ -1,981 +1,1236 @@
...
+    private QueryResponse response;
+    protected ClickHouseBinaryFormatReader reader;
+    private final ResultSetMetaData metaData;
+    private boolean closed;
+
+    public ResultSetImpl(QueryResponse response, ClickHouseBinaryFormatReader reader) {
+        this.response = response;
+        this.reader = reader;
+        this.metaData = new com.clickhouse.jdbc.metadata.ResultSetMetaData();
+        this.closed = false;
```
chernser, line 33: may be bind to reader and response state? just for easier calculation?

clickhouse-jdbc/src/main/java/com/clickhouse/jdbc/ResultSetImpl.java
```diff
@@ -1,981 +1,1236 @@
...
-        return new byte[0];
+        checkClosed();
+        return reader.getByteArray(columnIndex);
     }
 
     @Override
     public Date getDate(int columnIndex) throws SQLException {
-        return null;
+        checkClosed();
+        LocalDate localDate = reader.getLocalDate(columnIndex);
```
chernser, line 137: This I think we need to move to the reader. Would you please add `TODO:` for me here in the code?

clickhouse-jdbc/src/main/java/com/clickhouse/jdbc/ResultSetImpl.java
```diff
@@ -1,981 +1,1236 @@
...
+            return null;
+        }
+        return Timestamp.valueOf(localDateTime);
     }
 
     @Override
     public InputStream getAsciiStream(int columnIndex) throws SQLException {
-        return null;
+        checkClosed();
+        return new ByteArrayInputStream(reader.getString(columnIndex).getBytes(StandardCharsets.UTF_8));
```
chernser, line 167: I think, this also should be part of the reader. Please add a TODO: 

Here is some info about stream content -https://docs.oracle.com/cd/A97335_01/apps.102/a83724/oralob2.htm#1043220
```
To read from a CLOB, use the getAsciiStream() or getCharacterStream() method of an oracle.sql.CLOB object to retrieve the entire CLOB as an input stream. The getAsciiStream() method returns an ASCII input stream in a java.io.InputStream object. The getCharacterStream() method returns a Unicode input stream in a java.io.Reader object.
```
So it is not correct to return UTF8, I think. 

lets just throw exception that it is not implemented. Will sort it out later.

Issue #1852, opened 2024-10-07
[client-v2] No conversion between number types (ex. short to int) 


chernser:
### Describe the bug
It should be possible to get number values in different convertible types.  Today it is required for application to know datatype of a column. 

For example, binary format reader:
```
java.lang.ClassCastException: class java.lang.Short cannot be cast to class java.lang.Integer (java.lang.Short and java.lang.Integer are in module java.base of loader 'bootstrap')

	at com.clickhouse.client.api.data_formats.internal.AbstractBinaryFormatReader.getInteger(AbstractBinaryFormatReader.java:484)
```

### Expected behavior
Convertion between number types should be done automatically where applicable without data truncation.

PR #1856, opened 2024-10-08
[client-v2] fixed number conversion in reader

chernser:
## Summary
This PR implement conversion between different number types. Today client API (readers) allow to get column value only in the type it is defined in a DB. It is very inconvenient because requires to follow schema very strictly. So this PR relaxes this rule and lets to:
- read a value of a number column as number with bigger range than it is defined in DB- 'int8' as `int`, `int32` as `long`. 
- read a value of a number column as number with smaller range than it is defined in DB if actual value fits it - `int32` value like `1000` can be read as `short` but not as `byte`.

Closes: https://github.com/ClickHouse/clickhouse-java/issues/1852

## Checklist


mzitnik (approved):
Can we tell if there any performance impact from this change?

chernser:
@mzitnik I may not tell exact numbers - will try running benchmark tests.

PR #1857, opened 2024-10-08
[client-v2] Nullable nested column

javiercj93:
## Summary
When using the client-v2 with nullable nested columns, client-v2 was unable to insert them as it was treating the inner items as non nullable. Have not tested it but I guess that the same thing was happening for Array(Nullable(column_type)).

These changes aim to check whether array base column is nullable and if the item in the array is also null, serialize null.

Closes: https://github.com/ClickHouse/clickhouse-java/issues/1858

## Checklist
- [x ] Unit and integration tests covering the common scenarios were added

### DIFF
--- a/client-v2/src/main/java/com/clickhouse/client/api/Client.java
+++ b/client-v2/src/main/java/com/clickhouse/client/api/Client.java
@@ -1068,9 +1068,13 @@ public class Client implements AutoCloseable {
                             }
                         }
                     } else {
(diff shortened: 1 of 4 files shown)


chernser:
@javiercj93 Thank you for the fix!

javiercj93:
@chernser could you trigger pipeline again, did not realise tests were failing when reading so I applied the fix there as well.

chernser (approved):

Issue #1836, opened 2024-09-25
[client-v2] Log Comment


chernser:
### Description 
Client v2 should support log comments passed thru query parameters

PR #1864, opened 2024-10-10
[client-v2] Added setting log_comment

chernser:
## Summary
There is an option to set a comment that will be seen in `system.query_log` record associated with a query. It is useful for troubleshooting. In case of HTTP interface client should send the comment thru query parameter `log_comment`. 

Closes: https://github.com/ClickHouse/clickhouse-java/issues/1836
## Checklist

### DIFF
--- a/client-v2/src/main/java/com/clickhouse/client/api/command/CommandResponse.java
+++ b/client-v2/src/main/java/com/clickhouse/client/api/command/CommandResponse.java
@@ -5,7 +5,7 @@ import com.clickhouse.client.api.metrics.OperationMetrics;
 import com.clickhouse.client.api.metrics.ServerMetrics;
 import com.clickhouse.client.api.query.QueryResponse;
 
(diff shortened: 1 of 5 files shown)


mzitnik (approved):
Looks like Query & Insert Settings share some variables do you think we should extract a common base for team?

chernser:
Thanks @mzitnik! I've replaced code with shared constant from `ClientSettings`.

mzitnik:
> Thanks @mzitnik! I've replaced code with shared constant from `ClientSettings`.

Actually was referring if we can have a common base class since there is shared code

chernser:
@mzitnik I think, we should not do it to keep developer away from idea of extracting common part. 
I do not plan to add all server settings in such way and it should not be a problem for maintenance. What I'm afraid of that joining two separate mechanisms will lead to more rigid code and it will not save time as intendent.

Issue #1866, opened 2024-10-11
[client-v2] Underscore in column name is removed


averemee-si:
### Describe the bug
According to the [documentation](https://clickhouse.com/docs/en/sql-reference/syntax), underscore is a valid character in a column name:

```
Non-quoted identifiers must match the regex ^[a-zA-Z_][0-9a-zA-Z_]*$ and can not be equal to [keywords](https://clickhouse.com/docs/en/sql-reference/syntax#keywords). Examples: x, _1, X_y__Z123_.
```
But when I try to register an POJO for a table in the Java client-v2, an error occurs for any column that contains an underscore in the name:
Exception in thread "main" java.lang.IllegalArgumentException: No serializer found for column 'ORA_ROW_ID'. Did you forget to register it?
	at com.clickhouse.client.api.Client.insert(Client.java:1178)
	at com.clickhouse.client.api.Client.insert(Client.java:1124)
	at solutions.a2.oracle.expimp.pipe.DtoTest.main(DtoTest.java:84)

### Steps to reproduce
1. Create project with Java client-v2, i.e. for Gradle add dependency:
```
	def clickHouseVersion = "0.7.0"
	// https://mvnrepository.com/artifact/com.clickhouse/clickhouse-client
	implementation("com.clickhouse:client-v2:$clickHouseVersion")
```

2. Create table
```
  CREATE TABLE TEST4CH(
    ORA_ROW_ID String,
  	OWNER String, 
	NAME String, 
	TYPE String,
	LINE UInt32,
	TEXT String, 
	ORIGIN_CON_ID Nullable(UInt32)) engine = MergeTree order by ();
```

3. Try to register POJO for this table and then insert to this table

### Expected behaviour
Successful execution

### Code example

```
		client.register(dynamicClass, client.getTableSchema("TEST4CH"));
		InsertSettings settings = new InsertSettings().compressClientRequest(true);
		try (InsertResponse response = client.insert("TEST4CH", list).get()) {
			System.out.println("rows written=" + response.getWrittenRows());
		} catch (InterruptedException e) {
			// TODO Auto-generated catch block
			e.printStackTrace();
		} catch (ExecutionException e) {
			// TODO Auto-generated catch block
			e.printStackTrace();
		}
		client.close();

```
### Error log
6:50:01,415  INFO main Client:build:886 - Using server timezone: UTC
06:50:01,712  INFO main Client:poolConnectionManager:201 - Connection reuse strategy: FIFO
[...]

PR #1867, opened 2024-10-11
register(): make underscore legal in column name

averemee-si:
## Summary
fix for [[client-v2] Underscore in column name is removed](https://github.com/ClickHouse/clickhouse-java/issues/1866) 

## Checklist

### DIFF
--- a/client-v2/src/main/java/com/clickhouse/client/api/Client.java
+++ b/client-v2/src/main/java/com/clickhouse/client/api/Client.java
@@ -1040,7 +1040,7 @@ public class Client implements AutoCloseable {
         boolean defaultsSupport = schema.hasDefaults();
         tableSchemaHasDefaults.put(schemaKey, defaultsSupport);
         for (ClickHouseColumn column : schema.getColumns()) {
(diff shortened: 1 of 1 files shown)


client-v2/src/main/java/com/clickhouse/client/api/Client.java
```diff
@@ -1040,7 +1040,7 @@ public synchronized void register(Class<?> clazz, TableSchema schema) {
         boolean defaultsSupport = schema.hasDefaults();
         tableSchemaHasDefaults.put(schemaKey, defaultsSupport);
         for (ClickHouseColumn column : schema.getColumns()) {
-            String propertyName = column.getColumnName().toLowerCase().replace("_", "").replace(".", "");
```
chernser, line 1043: the purpose of this replacement is to leave only words without delimiting characters for easier compare. 
For example, column name `first_name` should match getter method name `firstName`, that is why `_` is removed. 
If you have getters with underscores then is should be replaced from getter method name, too. 
So current fix breaks canonical case when getters are without `_`.

chernser:
I'm canceling this PR because it solves only part of the problem. Out team will work on the fix and will create a new PR for this issue.

PR #1872, opened 2024-10-17
Stable base for PreparedStatement

Paultagoras:
The one real thorny issue I was debating is how to actually "compile" the SQL statement. The approach I took here is to use the default toString() method of whatever object they submitted.

### DIFF
--- a/clickhouse-jdbc/src/main/java/com/clickhouse/jdbc/PreparedStatementImpl.java
+++ b/clickhouse-jdbc/src/main/java/com/clickhouse/jdbc/PreparedStatementImpl.java
@@ -1,7 +1,7 @@
 package com.clickhouse.jdbc;
 
-import com.clickhouse.logging.Logger;
-import com.clickhouse.logging.LoggerFactory;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 import java.io.InputStream;
 import java.io.Reader;
@@ -12,19 +12,19 @@ import java.sql.*;
 import java.util.Calendar;
 
 public class PreparedStatementImpl extends StatementImpl implements PreparedStatement, JdbcWrapper {
-    private static final Logger log = LoggerFactory.getLogger(PreparedStatementImpl.class);
+    private static final Logger LOG = LoggerFactory.getLogger(PreparedStatementImpl.class);
 
-    String initialSql;
+    String originalSql;
     String [] sqlSegments;
     Object [] parameters;
     public PreparedStatementImpl(ConnectionImpl connection, String sql) {
         super(connection);
-        this.initialSql = sql;
+        this.originalSql = sql;
         //Split the sql string into an array of strings around question mark tokens
         this.sqlSegments = sql.split("\\?");
 
         //Create an array of objects to store the parameters
-        if (initialSql.contains("?")) {
+        if (originalSql.contains("?")) {
             this.parameters = new Object[sqlSegments.length];
         } else {
             this.parameters = new Object[0];
@@ -39,8 +39,7 @@ public class PreparedStatementImpl extends StatementImpl implements PreparedStat
                 sb.append(parameters[i]);
             }
         }
-        log.trace("Compiled SQL: {}", sb);
-        System.out.println("Compiled SQL: " + sb);
+        LOG.trace("Compiled SQL: {}", sb);
         return sb.toString();
     }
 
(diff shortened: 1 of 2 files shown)


clickhouse-jdbc/src/main/java/com/clickhouse/jdbc/PreparedStatementImpl.java
```diff
@@ -1,306 +1,397 @@
 package com.clickhouse.jdbc;
 
+import com.clickhouse.logging.Logger;
```
chernser, line 3: I would like to go away from proprietary wrappers for very standard things. 
There is SLF4J that works perfectly. 
I think it is important to change it now because later it will require to cut off too many weed.

clickhouse-jdbc/src/main/java/com/clickhouse/jdbc/PreparedStatementImpl.java
```diff
@@ -1,306 +1,397 @@
...
 import java.io.Reader;
 import java.math.BigDecimal;
 import java.net.URL;
+import java.nio.charset.StandardCharsets;
 import java.sql.*;
 import java.util.Calendar;
 
 public class PreparedStatementImpl extends StatementImpl implements PreparedStatement, JdbcWrapper {
-    String sql;
+    private static final Logger log = LoggerFactory.getLogger(PreparedStatementImpl.class);
```
chernser, line 15: should be LOG because it is `static final`

clickhouse-jdbc/src/main/java/com/clickhouse/jdbc/PreparedStatementImpl.java
```diff
@@ -1,306 +1,397 @@
...
+
+    String initialSql;
+    String [] sqlSegments;
+    Object [] parameters;
     public PreparedStatementImpl(ConnectionImpl connection, String sql) {
         super(connection);
-        this.sql = sql;
+        this.initialSql = sql;
```
chernser, line 22: may be name it `originalSQL`?

clickhouse-jdbc/src/main/java/com/clickhouse/jdbc/PreparedStatementImpl.java
```diff
@@ -1,306 +1,397 @@
...
+        //Split the sql string into an array of strings around question mark tokens
+        this.sqlSegments = sql.split("\\?");
```
chernser, line 24: how would it handle `IN(?, ?, ?)` ? 
I think this logic is part of execution phase when we know more about parameters.

clickhouse-jdbc/src/main/java/com/clickhouse/jdbc/PreparedStatementImpl.java
```diff
@@ -1,306 +1,397 @@
...
+    private String compileSql() {
+        StringBuilder sb = new StringBuilder();
+        for (int i = 0; i < sqlSegments.length; i++) {
+            sb.append(sqlSegments[i]);
+            if (i < parameters.length) {
+                sb.append(parameters[i]);
+            }
+        }
+        log.trace("Compiled SQL: {}", sb);
+        System.out.println("Compiled SQL: " + sb);
```
chernser, line 43: lets use LOG instead.

chernser, line 43: remove?

clickhouse-jdbc/src/test/java/com/clickhouse/jdbc/PreparedStatementTest.java
```diff
@@ -0,0 +1,212 @@
...
+import java.sql.PreparedStatement;
+import java.sql.ResultSet;
+import java.sql.Types;
+
+import static org.testng.Assert.assertEquals;
+import static org.testng.Assert.assertFalse;
+import static org.testng.Assert.assertNull;
+import static org.testng.Assert.assertTrue;
+
+public class PreparedStatementTest extends JdbcIntegrationTest {
```
chernser, line 15: I would squash all these tests into one because each test will spin up a container just for few queries.

chernser: actually it will not - one instance of docker env will be reused in all methods in the class.

Issue #1827, opened 2024-09-19
JDBC driver prints stack trace to stdout


jjtt:
### Describe the bug

On some exceptions the JDBC driver prints a whole stack trace to stdout here: https://github.com/ClickHouse/clickhouse-java/blob/main/clickhouse-client/src/main/java/com/clickhouse/client/ClickHouseClient.java#L963

This can mess up structured logging output for an application that uses the ClickHouse JDBC driver.

### Steps to reproduce
1. Connect to a db from an application
2. Revoke the db credentials
3. See stack trace in application stdout

### Expected behaviour

Should use some common logging framework where the application can control logging from libraries. For example: SLF4J or logback.

Or throw an exception and let the application worry about it.

Am-phi:
Hello @chernser , 
Unless you already did something I'd be glad to solve this simple issue. 
My take would be to have the same behaviour as the client_v2 and simply return false. 

Even though I don't see generally a good idea to silence an exception, I wonder if adding a log is useful as the fact of having the clickhouse instance, we are trying to reach, unavailable is already part of the goal of calling this method. What is your opinion on the matter?

chernser:
Good day, @Am-phi ! 
Thank you for suggesting the help - I will appreciate it! 

What you are saying is true - we should always log exception or re-throw it. 
returning `false` in this case is a correct thing - because it indicates that server is not reachable. And logging the exception is important for investigation later.

PR #1880, opened 2024-10-23
Fix JDBC driver prints stack trace to stdout

Am-phi:
## Summary
This commit solves the issue #1827 by logging the exception instead of printing it to stdout. A log in the client v2 has also been added so that both clients are iso.

## Checklist

### DIFF
--- a/client-v2/src/main/java/com/clickhouse/client/api/Client.java
+++ b/client-v2/src/main/java/com/clickhouse/client/api/Client.java
@@ -1035,6 +1035,7 @@ public class Client implements AutoCloseable {
             try (QueryResponse response = query("SELECT 1 FORMAT TabSeparated").get(timeout, TimeUnit.MILLISECONDS)) {
                 return true;
             } catch (Exception e) {
+                LOG.debug("Failed to connect to the server", e);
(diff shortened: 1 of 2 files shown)


client-v2/src/main/java/com/clickhouse/client/api/Client.java
```diff
@@ -1032,9 +1032,10 @@ public boolean ping() {
      */
     public boolean ping(long timeout) {
         if (useNewImplementation) {
-            try (QueryResponse response = query("SELECT 1 FORMAT TabSeparated").get(timeout, TimeUnit.MILLISECONDS)) {
+            try (QueryResponse ignored = query("SELECT 1 FORMAT TabSeparated").get(timeout, TimeUnit.MILLISECONDS)) {
```
chernser, line 1035: it is good to keep PR changes as minimal as possible. 
I do not mind about this change, but we do not do like this:
- java has no good way to ignore returned value in this case  
- variable names should be meaningful.

chernser (approved):

chernser:
@Am-phi thank you so much for the contribution!

PR #1890, opened 2024-10-27
[jdbc] Enhance useragent to indicate Apache Spark

BentsiLeviav:
Many users use Apache Spark (or any other framework) via this JDBC connector.
This PR adds "framework detecting" functionality. It looks at the current stack trace, infers the framework usage based on a closed list of tracked frameworks, and adds an indication for it in the user agent string.

### DIFF
--- a/clickhouse-http-client/src/main/java/com/clickhouse/client/http/ClickHouseHttpConnection.java
+++ b/clickhouse-http-client/src/main/java/com/clickhouse/client/http/ClickHouseHttpConnection.java
@@ -53,7 +53,7 @@ public abstract class ClickHouseHttpConnection implements AutoCloseable {
     private static final byte[] SUFFIX_FORMAT = "_format\"\r\n\r\n".getBytes(StandardCharsets.US_ASCII);
     private static final byte[] SUFFIX_STRUCTURE = "_structure\"\r\n\r\n".getBytes(StandardCharsets.US_ASCII);
     private static final byte[] SUFFIX_FILENAME = "\"; filename=\"".getBytes(StandardCharsets.US_ASCII);
-    private static final List<String> FRAMEWORKS = List.of("apache.spark");
+    private static final List<String> FRAMEWORKS_TO_INFER = List.of("apache.spark");
 
     private static StringBuilder appendQueryParameter(StringBuilder builder, String key, String value) {
         return builder.append(urlEncode(key, StandardCharsets.UTF_8)).append('=')
@@ -414,7 +414,7 @@ public abstract class ClickHouseHttpConnection implements AutoCloseable {
     protected String getAdditionalFrameworkUserAgent() {
         Set<String> inferredFrameworks = new LinkedHashSet<>();
         for (StackTraceElement ste : Thread.currentThread().getStackTrace()) {
-            for (String framework : FRAMEWORKS) {
+            for (String framework : FRAMEWORKS_TO_INFER) {
                 if (ste.toString().contains(framework)) {
                     inferredFrameworks.add(String.format("(%s)", framework));
                 }
--- a/clickhouse-http-client/src/main/java/com/clickhouse/client/http/ClickHouseHttpConnection.java
+++ b/clickhouse-http-client/src/main/java/com/clickhouse/client/http/ClickHouseHttpConnection.java
@@ -53,7 +53,6 @@ public abstract class ClickHouseHttpConnection implements AutoCloseable {
     private static final byte[] SUFFIX_FORMAT = "_format\"\r\n\r\n".getBytes(StandardCharsets.US_ASCII);
     private static final byte[] SUFFIX_STRUCTURE = "_structure\"\r\n\r\n".getBytes(StandardCharsets.US_ASCII);
     private static final byte[] SUFFIX_FILENAME = "\"; filename=\"".getBytes(StandardCharsets.US_ASCII);
-    private static final List<String> FRAMEWORKS_TO_INFER = List.of("apache.spark");
 
     private static StringBuilder appendQueryParameter(StringBuilder builder, String key, String value) {
         return builder.append(urlEncode(key, StandardCharsets.UTF_8)).append('=')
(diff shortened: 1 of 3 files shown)


clickhouse-http-client/src/main/java/com/clickhouse/client/http/ClickHouseHttpConnection.java
```diff
@@ -417,10 +410,23 @@ protected String getDefaultUserAgent() {
         return config.getClientName();
     }
 
+    protected String getAdditionalFrameworkUserAgent() {
+        List<String> frameworks = List.of("apache.spark");
+        Set<String> inferredFrameworks = new LinkedHashSet<>();
+        for (StackTraceElement ste : Thread.currentThread().getStackTrace()) {
+            for (String framework : frameworks) {
+                if (ste.toString().contains(framework)) {
+                    inferredFrameworks.add(String.format("(%s)", framework));
```
mzitnik, line 419: Is the idea to collect all framework, can you provide an example?

mzitnik: or we should break once we detected one of them

clickhouse-http-client/src/main/java/com/clickhouse/client/http/ClickHouseHttpConnection.java
```diff
@@ -417,10 +410,23 @@ protected String getDefaultUserAgent() {
         return config.getClientName();
     }
 
+    protected String getAdditionalFrameworkUserAgent() {
```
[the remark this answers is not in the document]
BentsiLeviav: It would, but this is what we want.
Eventually, a user can write a Java application, using Spark with JDBC, but also might use just JDBC (without any framework). We want to classify each request and ensure we capture the right use case.
Do you happen to have another way to implement this to get the behavior I described?

clickhouse-http-client/src/main/java/com/clickhouse/client/http/ClickHouseHttpConnection.java
BentsiLeviav: IMHO we don't want to limit ourselves to a single framework use case. A nested or co-shared use case might be possible. From a complexity point of view, it is kind of the same, therefore I would prefer us not to limit this.

clickhouse-http-client/src/main/java/com/clickhouse/client/http/ClickHouseHttpConnection.java
```diff
@@ -417,10 +410,23 @@ protected String getDefaultUserAgent() {
...
+        List<String> frameworks = List.of("apache.spark");
```
chernser, line 414: This method will be called every time we do a request, `frameworks` is a static information. Please move it out of the method.

clickhouse-http-client/src/main/java/com/clickhouse/client/http/ClickHouseHttpConnection.java
```diff
@@ -417,10 +410,23 @@ protected String getDefaultUserAgent() {
...
+        Set<String> inferredFrameworks = new LinkedHashSet<>();
+        for (StackTraceElement ste : Thread.currentThread().getStackTrace()) {
```
chernser, line 416: Taking current thread stack trace approach would not work with asynchronous tasks (but we have moved away from this default behavior) because task may have very independent stack trace.

clickhouse-http-client/src/main/java/com/clickhouse/client/http/ClickHouseHttpConnection.java
chernser: I think, we need to exit loop when framework is detected. 
And may be remember it to not search again.

chernser:
Good day, @BentsiLeviav! 
This feature is only useful for clients that using ClickHouse Cloud. 
Other users should have an option to turn it off. 

Would you please also tell story behind this feature? What would we like to find out? 

Thanks!

clickhouse-http-client/src/main/java/com/clickhouse/client/http/ClickHouseHttpConnection.java
BentsiLeviav: Do you have any suggestions for that? If not, we better have something rather than nothing.

chernser: We may look up on startup for specific classes - usually frameworks have some.

clickhouse-http-client/src/main/java/com/clickhouse/client/http/ClickHouseHttpConnection.java
BentsiLeviav: @chernser 
You must search every time. Each request in an application doesn't necessarily mean we use the framework (you can write s Spark application, work with Spark JDBC, and later on open a regular JDBC connection and query CH).

In addition, could you elaborate on why do we need to exit the loop if a framework was detected? what if a user might use a combination of frameworks that we want to detect? (because as I said, from a complexity point of view there isn't much difference (the list of FRAMEWORKS_TO_INFER is going to be limited), so what is the rationality behind limiting this ability?)

BentsiLeviav:
@chernser 
This feature was meant for us to better monitor and understand the real usage of the JDBC driver. 
We want to be able to know exactly what kind of technology/framework our users use. 
As of today, the JDBC driver is the underlying connection for many frameworks/technologies such as:

- Apache Spark
- DBeaver
- Tableau

for some of these frameworks, we can't customize the product name. Therefore, I added this feature to infer what kind of framework is used (out of a close list we define), and add it to the user agent metadata.

I talked in this PR, I have moved the implementation to the JDBC (to avoid any async tasks) and minimized the usage by making sure the logic runs only on client creation.

clickhouse-jdbc/src/main/java/com/clickhouse/jdbc/ClickHouseDriver.java
```diff
@@ -152,6 +172,8 @@ public ClickHouseConnection connect(String url, Properties info) throws SQLExcep
         if (!acceptsURL(url)) {
             return null;
         }
+        if (!url.toLowerCase().contains("disable-frameworks-detection"))
```
chernser, line 175: not every case passes properties thru `url` - often properties are used. 
I suggest do this check in `com.clickhouse.jdbc.internal.ClickHouseConnectionImpl#ClickHouseConnectionImpl(com.clickhouse.jdbc.internal.ClickHouseJdbcUrlParser.ConnectionInfo)` - this method is called right after URL is parsed for properties so ConnectionInfo contains all of them.

chernser:
@BentsiLeviav thank you for explanation!

chernser (approved):
Please do

Issue #1305, opened 2023-04-03
clickhouse-jdbc password contains =？ Connection failed.


xiaowuyy:
### Describe the bug
password contains =？use clickhouse-jdbc connection failed.
### Steps to reproduce
1. clickhouse server version 21.8.8.1
2. password  AF.O4P=?m76.lHfa
3.     <dependency>
            <groupId>ru.yandex.clickhouse</groupId>
            <artifactId>clickhouse-jdbc</artifactId>
            <version>0.3.2</version>
        </dependency>
4. error message： 
Code: 516, e.displayText() = DB::Exception: apmhouse: Auth*****#*#*****he username or password*****#*#*****ion 21.8.8.1)[N], server ClickHouseNode(addr=http:ip:port, db=system)@-609109239
### Expected behaviour
connection success
### Code example
```java
        Properties properties = new Properties();
        properties.setProperty("ssl", "true");
        properties.setProperty("sslmode", "NONE");
        properties.setProperty("user", "apmhouse");
        properties.setProperty("password", "AF.O4P=?m76.lHfa");
        Connection connection = DriverManager.getConnection(url, properties);
```
### Error log
client:
Code: 516, e.displayText() = DB::Exception: apmhouse: Auth*****##he username or password##*****ion 21.8.8.1)[N], server ClickHouseNode(addr=http:ip:port, db=system)@-609109239
server:
2023.04.03 15:46:54.681974 [ 456 ] {} <Trace> DynamicQueryHandler: Request URI: /?extremes=0
2023.04.03 15:46:54.686019 [ 456 ] {} <Error> Access(user directories): apmhouse: Authentication failed: Code: 193, e.displayText() = DB::Exception: Invalid credentials, Stack trace (when copying this message, always include the lines below):

0. ? @ 0x10273542 in /opt/apm-clickhouse/programs/clickhouse
1. ? @ 0x519804f in /opt/apm-clickhouse/programs/clickhouse
2. ? @ 0x4aad274 in /opt/apm-clickhouse/programs/clickhouse
3. ? @ 0x4aaf101 in /opt/apm-clickhouse/programs/clickhouse
4. ? @ 0xc6f5d6a in /opt/apm-clickhouse/programs/clickhouse
5. ? @ 0xc71bb53 in /opt/apm-clickhouse/programs/clickhouse
6. ? @ 0xc6f5d6a in /opt/apm-clickhouse/programs/clickhouse
[...]

PR #1892, opened 2024-10-29
[client-v1,client-v2, auth] Implemented authentication via http basic auth

chernser:
## Summary
Password can be passed as value of the `X-ClickHouse-Key` header [docs](https://clickhouse.com/docs/en/interfaces/http#default-database). But header values have some limitations about what characters can be used (learned from practical tests and https://www.rfc-editor.org/rfc/rfc7230#section-3.2.6) what would require escaping special chars. Basic authentication helps to avoid all this and is standard authentication mechanism.
This PR changes both clients (v1, v2). 

New configuration parameter is added: 
- `com.clickhouse.client.http.config.ClickHouseHttpOption#USE_BASIC_AUTHENTICATION` - allows to switch using ClickHouse `X-` headers to authenticate. Default is using basic auth. 

Closes: https://github.com/ClickHouse/clickhouse-java/issues/1305

## Checklist

### DIFF
--- a/clickhouse-http-client/src/main/java/com/clickhouse/client/http/ClickHouseHttpConnection.java
+++ b/clickhouse-http-client/src/main/java/com/clickhouse/client/http/ClickHouseHttpConnection.java
@@ -234,12 +234,18 @@ public abstract class ClickHouseHttpConnection implements AutoCloseable {
             map.put("authorization", credentials.getAccessToken());
         } else if (!hasAuthorizationHeader) {
             if (config.isSsl() && !ClickHouseChecker.isNullOrEmpty(config.getSslCert())) {
-                map.put("x-clickhouse-user", credentials.getUserName());
-                map.put("x-clickhouse-ssl-certificate-auth", "on");
+                map.put(ClickHouseHttpProto.HEADER_DB_USER, credentials.getUserName());
+                map.put(ClickHouseHttpProto.HEADER_SSL_CERT_AUTH, "on");
             } else {
-                String password = credentials.getPassword() == null ? "" : credentials.getPassword();
-                map.put(HttpHeaders.AUTHORIZATION, "Basic " + Base64.getEncoder()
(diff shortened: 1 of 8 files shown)


clickhouse-http-client/src/main/java/com/clickhouse/client/http/ClickHouseHttpConnection.java
```diff
@@ -238,11 +240,13 @@ protected static Map<String, String> createDefaultHeaders(ClickHouseConfig confi
...
             map.put("authorization", credentials.getAccessToken());
         } else if (!hasAuthorizationHeader) {
-            map.put("x-clickhouse-user", credentials.getUserName());
             if (config.isSsl() && !ClickHouseChecker.isNullOrEmpty(config.getSslCert())) {
+                map.put("x-clickhouse-user", credentials.getUserName());
                 map.put("x-clickhouse-ssl-certificate-auth", "on");
-            } else if (!ClickHouseChecker.isNullOrEmpty(credentials.getPassword())) {
-                map.put("x-clickhouse-key", credentials.getPassword());
+            } else {
+                String password = credentials.getPassword() == null ? "" : credentials.getPassword();
```
mzitnik, line 247: Just to understand you have changed the logic to use Basic authorization to avoid the illegal characters.

chernser: Yes, I've changed this logic to avoid problem with different characters. 
So even sequence `=?` is getting problem.
Additionally it does things more standard. 
Javascript Client uses this approach as default - https://github.com/ClickHouse/clickhouse-js/blob/main/packages/client-node/src/connection/node_base_connection.ts#L87

mzitnik (approved):

mzitnik:
@chernser please validate discussed

mzitnik (approved):

Issue #1912, opened 2024-11-06
[JDBC] java.lang.NoClassDefFoundError: com/clickhouse/client/internal/apache/hc/core5/http2/HttpVersionPolicy


ivan-klass:
### Describe the bug
NoClassDefFoundError

### Steps to reproduce
I've updated from 0.6.5 to version 0.7.1.
I'm using "shaded" jar - so I expect not to care about any other thirdparty libraries.
The jar is "runtime" dependency, not used by compiler.

### Expected behaviour
Works as before

### Error log
```
java.lang.NoClassDefFoundError: com/clickhouse/client/internal/apache/hc/core5/http2/HttpVersionPolicy
	at com.clickhouse.client.internal.apache.hc.client5.http.config.TlsConfig$Builder.build(TlsConfig.java:211)
	at com.clickhouse.client.internal.apache.hc.client5.http.config.TlsConfig.<clinit>(TlsConfig.java:47)
	at com.clickhouse.client.internal.apache.hc.client5.http.impl.io.PoolingHttpClientConnectionManager.resolveTlsConfig(PoolingHttpClientConnectionManager.java:276)
	at com.clickhouse.client.internal.apache.hc.client5.http.impl.io.PoolingHttpClientConnectionManager.connect(PoolingHttpClientConnectionManager.java:441)
	at com.clickhouse.client.internal.apache.hc.client5.http.impl.classic.InternalExecRuntime.connectEndpoint(InternalExecRuntime.java:162)
	at com.clickhouse.client.internal.apache.hc.client5.http.impl.classic.InternalExecRuntime.connectEndpoint(InternalExecRuntime.java:172)
	at com.clickhouse.client.internal.apache.hc.client5.http.impl.classic.ConnectExec.execute(ConnectExec.java:142)
	at com.clickhouse.client.internal.apache.hc.client5.http.impl.classic.ExecChainElement.execute(ExecChainElement.java:51)
	at com.clickhouse.client.internal.apache.hc.client5.http.impl.classic.ProtocolExec.execute(ProtocolExec.java:192)
	at com.clickhouse.client.internal.apache.hc.client5.http.impl.classic.ExecChainElement.execute(ExecChainElement.java:51)
	at com.clickhouse.client.internal.apache.hc.client5.http.impl.classic.HttpRequestRetryExec.execute(HttpRequestRetryExec.java:96)
[...]

PR #1915, opened 2024-11-06
[JDBC] Added removed dependency to shaded jar. Corrected jdbc example to verify

chernser:
## Summary
There is a package classifier "http" with shaded AHC. While deprecation of some components one crucial dependency was removed that (surprise!) was used in base scenario. This PR returns this dependency and adds profiles to example so it can be verified. 

Closes: https://github.com/ClickHouse/clickhouse-java/issues/1912

## Checklist


ivan-klass:
@chernser thank you! It would be great for this added example execution to be a part of CI tests, isn't it?

ivan-klass:
I wish I can help with adding any "shaded" tests [here](https://github.com/ClickHouse/clickhouse-java/blob/main/.github/workflows/build.yml) but unfortunatelly I'm not familiar with `mvn` at all (only used Gradle and sbt so far)

PR #1920, opened 2024-11-07
Co-mingle Drivers

Paultagoras:
## Checklist

### DIFF
--- a/clickhouse-client/src/test/java/com/clickhouse/client/ClickHouseServerForTest.java
+++ b/clickhouse-client/src/test/java/com/clickhouse/client/ClickHouseServerForTest.java
@@ -339,8 +339,9 @@ public class ClickHouseServerForTest {
 
             try {
                 clickhouseContainer.start();
-                if (!runQuery("CREATE DATABASE IF NOT EXISTS " + database)) {
-                    throw new IllegalStateException("Failed to create database for testing.");
+
+                if (clickhouseContainer.isRunning()) {
+                    runQuery("CREATE DATABASE IF NOT EXISTS " + database);
                 }
             } catch (RuntimeException e) {
                 throw new IllegalStateException(new StringBuilder()


clickhouse-jdbc/src/main/java/com/clickhouse/jdbc/ClickHouseDriver.java
```diff
@@ -69,6 +69,9 @@ public static String getFrameworksDetected() {
     }
 
     static {
+        String versionString = System.getProperty("com.clickhouse.jdbc.version", "v2");
```
chernser, line 72: lets use another name like "com.clickhouse.jdbc.use_new_implementation" because 
- we may use com.clickhouse.jdbc.version for other reasons in the future 
- it may be too confusing

jdbc-v2/src/main/java/com/clickhouse/jdbc/metadata/DatabaseMetaData.java
```diff
@@ -9,13 +9,12 @@
...
 import com.clickhouse.jdbc.Driver;
-import com.clickhouse.jdbc.JdbcWrapper;
-import com.clickhouse.jdbc.ResultSetImpl;
+import com.clickhouse.jdbc.JdbcV2Wrapper;
 import com.clickhouse.jdbc.internal.JdbcUtils;
 import com.clickhouse.logging.Logger;
 import com.clickhouse.logging.LoggerFactory;
 
-public class DatabaseMetaData implements java.sql.DatabaseMetaData, JdbcWrapper {
+public class DatabaseMetaData implements java.sql.DatabaseMetaData, JdbcV2Wrapper {
```
chernser, line 17: Can we fix old JdbcWrapper and use it in both versions?

chernser (approved):

PR #1924, opened 2024-11-11
Proxy auth

NDHApps:
## Summary
Add basic auth support (username/password) for proxy

## Checklist
- [ X] A human-readable description of the changes was provided to include in CHANGELOG

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,8 @@
+## Latest
+
+### New Features
+- Added basic auth support for proxies. Now you can specify username/password when connecting via a proxy that requires it with HttpURLConnection and Apache HttpClient.
+
 ## 0.7.1-patch1
 
 ### Bug Fixes
--- a/clickhouse-client/src/main/java/com/clickhouse/client/ClickHouseConfig.java
+++ b/clickhouse-client/src/main/java/com/clickhouse/client/ClickHouseConfig.java
@@ -245,6 +245,8 @@ public class ClickHouseConfig implements ClickHouseDataConfig {
     private final ClickHouseProxyType proxyType;
     private final String proxyHost;
     private final int proxyPort;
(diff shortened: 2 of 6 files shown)


chernser:
Good day, @NDHApps ! 
Thank you for the contribution! However I should warn you that old client will be deprecated very soon. 

Would it be an option for you to migrate to the client-v2? Here is documentation https://clickhouse.com/docs/en/integrations/java/client-v2 

Thanks!

NDHApps:
> Good day, @NDHApps ! Thank you for the contribution! However I should warn you that old client will be deprecated very soon.
> 
> Would it be an option for you to migrate to the client-v2? Here is documentation https://clickhouse.com/docs/en/integrations/java/client-v2
> 
> Thanks!

Hey @chernser, thanks the the speedy response! This specific change is being used to unblock proxy support for the clickhouse-jdbc driver (I have been manually building it for a personal use for some time now and having to recompile it regularly with every change). I do see now that the V2 client already has a [setProxyCredentials](https://github.com/ClickHouse/clickhouse-java/blob/ff135812f8424fb056c1f74b0f3eb8b239d74e0f/client-v2/src/main/java/com/clickhouse/client/api/Client.java#L586C9-L590C10) method and that there is also a new jdbc-v2 library being built on top of this, which I will look into further. However, I'm not sure if or how soon my legacy application would be able to upgrade to this. Would this pending deprecation preclude contributions to the V1 library for the time being?

chernser:
@NDHApps I will merge this, no worries. 
If you are using jdbc then transition should be easier. We are working on new version that will have a few differences but more compliant to JDBC spec. 

Please keep in touch and let us know when you decide to migrate to the new version. Here is Community Slack Channel https://clickhouse.com/slack

chernser (approved):

Issue #1932, opened 2024-11-14
[jdbc] The ClickHouseHttpConnection overrides `result_overflow_mode`.  


chernser:
### Description 
There is a setting `result_overflow_mode` that controls what to do when result limit exceeded ([doc](https://clickhouse.com/docs/en/operations/settings/settings#result_overflow_mode)). Current JDBC implementation doesn't let an user to override this settings. It is especially needed when `use_query_cache` is set. In such cases `result_overflow_mode` would be `throw`. 

See also https://github.com/ClickHouse/ClickHouse/pull/69549 
Query Cache https://clickhouse.com/docs/en/operations/query-cache#configuration-settings-and-usage

saud-cloud:
I am facing same issue with overflow_mode

PR #1934, opened 2024-11-15
Fixed overriding max_result_rows settings

chernser:
## Summary
This PR fixes two problems: 
- force value for `result_overflow_mode` that was set to `break` and it was not possible to override it. Enforced in JDBC and in HTTP connection
- correct handling of `max_result_rows` - in attempt to make it 0 this setting was removed from request settings and because of this was set to client setting value what is wrong. It happens in ClickHouseHttpConnection. 

Closes: https://github.com/ClickHouse/clickhouse-java/issues/1932

## Checklist

### DIFF
--- a/clickhouse-http-client/src/main/java/com/clickhouse/client/http/ClickHouseHttpConnection.java
+++ b/clickhouse-http-client/src/main/java/com/clickhouse/client/http/ClickHouseHttpConnection.java
@@ -118,10 +118,23 @@ public abstract class ClickHouseHttpConnection implements AutoCloseable {
             appendQueryParameter(builder, settingKey, String.valueOf(config.getMaxExecutionTime()));
         }
         settingKey = ClickHouseClientOption.MAX_RESULT_ROWS.getKey();
-        if (config.getMaxResultRows() > 0L && !settings.containsKey(settingKey)) {
+        boolean hasRequestSetting = settings.containsKey(settingKey);
+        if (config.getMaxResultRows() > 0L && !hasRequestSetting) {
+            // set on client level
             appendQueryParameter(builder, settingKey, String.valueOf(config.getMaxResultRows()));
-            appendQueryParameter(builder, "result_overflow_mode", "break");
+        } else if (hasRequestSetting) {
+            // set on request level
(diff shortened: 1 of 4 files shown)


clickhouse-http-client/src/main/java/com/clickhouse/client/http/ClickHouseHttpConnection.java
```diff
@@ -118,10 +118,25 @@ static String buildQueryParams(ClickHouseRequest<?> request, Map<String, Seriali
...
-        if (config.getMaxResultRows() > 0L && !settings.containsKey(settingKey)) {
+        boolean hasRequestSetting = settings.containsKey(settingKey);
+        if (config.getMaxResultRows() > 0L && !hasRequestSetting) {
+            // set on client level
             appendQueryParameter(builder, settingKey, String.valueOf(config.getMaxResultRows()));
-            appendQueryParameter(builder, "result_overflow_mode", "break");
+        } else if (hasRequestSetting) {
+            // set on request level
+            Object value = settings.get(settingKey);
+            if (value instanceof Number && ((Number) value).longValue() > 0L) {
```
mzitnik, line 128: Why are we not checking the type on insert into the settings so we will not have instanceof on each time we build settings

chernser: This settings are coming from different places and it may be Number or String. I've made this check only for safety reasons.

mzitnik: Just saying we can make this check/conversion when the settings are updated, so when we run this method, we will deal with it, but correct me here it seems like it will deprecated and remove this part of code

mzitnik (approved):

clickhouse-http-client/src/main/java/com/clickhouse/client/http/ClickHouseHttpConnection.java
chernser: Yes, this will gone soon.

PR #1952, opened 2024-11-19
Make clickhouse-r2dbc use jdbc-v2 explicitly

mzitnik:
## Summary
`click house-r2dbc` uses the old driver explicitly so we can load according to testing. `jdbc-v2 `. I have replaced it with a new version.

## Checklist


mzitnik:
We have fixed the Driver code to proxy the `jdbc v2 `driver as default.

PR #1953, opened 2024-11-19
Updating where the proxy happens

Paultagoras:
## Checklist


clickhouse-jdbc/src/main/java/com/clickhouse/jdbc/ClickHouseDriver.java
```diff
@@ -1,232 +1,124 @@
...
-
-        log.debug("ClickHouse Driver {}(JDBC: {}) registered", driverVersion, specVersion);
     }
 
     public static void unload() {
         try {
-            log.info("Unregistering ClickHouse JDBC driver v1 ({})", driverVersion);
-            DriverManager.deregisterDriver(new ClickHouseDriver());
+            log.debug("Unloading the 'proxy' JDBC driver.");
+            java.sql.DriverManager.deregisterDriver(new ClickHouseDriver());
```
chernser, line 39: what if we have two different datasources? for old and the new version?

jdbc-v2/src/main/java/com/clickhouse/jdbc/ResultSetImpl.java
```diff
@@ -434,13 +434,21 @@ public ResultSetMetaData getMetaData() throws SQLException {
     @Override
     public Object getObject(int columnIndex) throws SQLException {
         checkClosed();
-        return null;
+        try {
+            return reader.readValue(columnIndex);
```
chernser, line 438: this is not correct - we have to do conversion to standard type.

chernser (approved):

chernser:
Lets merge and test how it would work with example applications.

PR #1961, opened 2024-11-21
Testing feedback

Paultagoras:
## Checklist


client-v2/src/main/java/com/clickhouse/client/api/internal/HttpAPIClientHelper.java
```diff
@@ -601,6 +601,15 @@ public static Map<String, String> parseUrlParameters(URL url) {
         Map<String, String> params = new HashMap<>();
 
         try {
+            String path = url.getPath();
```
chernser, line 604: just a note - this method should be outside of this class because I want to extract most common things.

chernser (approved):

PR #1965, opened 2024-11-22
Adding types, adjusting constructors

Paultagoras:
## Checklist


jdbc-v2/src/main/java/com/clickhouse/jdbc/ConnectionImpl.java
```diff
@@ -324,19 +323,31 @@ public PreparedStatement prepareStatement(String sql, String[] columnNames) thro
     @Override
     public Clob createClob() throws SQLException {
         checkOpen();
-        throw new SQLFeatureNotSupportedException("Clob not supported");
```
chernser, line 327: I think, we still need to throw exception as we do not support it and old driver either

chernser: just for reference - https://docs.oracle.com/javadb/10.10.1.2/ref/rrefclob.html
it can be added for example when we allow reading from a file, but today, I think it is not a widely used

jdbc-v2/src/main/java/com/clickhouse/jdbc/ConnectionImpl.java
```diff
@@ -358,30 +369,49 @@ public boolean isValid(int timeout) throws SQLException {
...
     public void setClientInfo(Properties properties) throws SQLClientInfoException {
-        throw new SQLClientInfoException("ClientInfo not supported", null);
+        try {
+            checkOpen();
+        } catch (SQLException e) {
+            throw new SQLClientInfoException("Failed to set client info.", new HashMap<>(), e);
+        }
+
+        for (Map.Entry<Object, Object> entry : properties.entrySet()) {
+            setClientInfo(entry.getKey().toString(), entry.getValue().toString());
```
chernser, line 389: ClientInfo is only what we can send to a server: 
```
The driver stores the value specified in a suitable location in the database.
```

and this one 

```
 The driver shall generate a warning if the client info name specified is not recognized by the driver.
```

Currently we may need to accept only 
```
    static final String PROP_APPLICATION_NAME = "ApplicationName";
``` 

I think this came from PostgreSQL JDBC driver - so we need to investigate more.

chernser: btw, each property is set as individual query in the PSQL JDBC

jdbc-v2/src/main/java/com/clickhouse/jdbc/ConnectionImpl.java
```diff
@@ -358,30 +369,49 @@ public boolean isValid(int timeout) throws SQLException {
...
+        }
     }
 
     @Override
     public String getClientInfo(String name) throws SQLException {
         checkOpen();
-        return null;
+        Object value = this.defaultQuerySettings.getAllSettings().get(name);
```
chernser, line 396: according to the comment above: we need to return only client info. In this case it will be value of 'User-Agent' header. 
I will do according change in the client.

jdbc-v2/src/main/java/com/clickhouse/jdbc/ConnectionImpl.java
```diff
@@ -358,30 +369,49 @@ public boolean isValid(int timeout) throws SQLException {
...
+        clientInfo.putAll(this.defaultQuerySettings.getAllSettings());
+        return clientInfo;
     }
 
     @Override
     public Array createArrayOf(String typeName, Object[] elements) throws SQLException {
-        //TODO: Should this be supported?
-        return null;
+        try {
+            return new com.clickhouse.jdbc.types.Array(List.of(elements));
```
chernser, line 411: this is reference to an old project. If you need to return a simple implementation - just copy into JDBC 2 please. 
There are a lot classes so I think we need to reply on old impl as less as possible

jdbc-v2/src/main/java/com/clickhouse/jdbc/types/Clob.java
```diff
@@ -0,0 +1,140 @@
+package com.clickhouse.jdbc.types;
+
+import java.io.InputStream;
+import java.io.OutputStream;
+import java.io.Reader;
+import java.io.Writer;
+import java.sql.SQLException;
+
+public class Clob implements java.sql.Clob {
+    String data;
```
chernser, line 10: it is a big amount of data - I think buffer or StringBuilder would work better.

chernser (approved):

PR #1986, opened 2024-11-29
Make ClickHouseBinaryFormatReader also extend Iterator

### LINKED ISSUE #1963: Why not `ClickHouseBinaryFormatReader` extends `Iterator` or `Stream` interface ?
### Describe your feedback

I think [ClickHouseBinaryFormatReader](https://github.com/ClickHouse/clickhouse-java/blob/main/client-v2/src/main/java/com/clickhouse/client/api/data_formats/ClickHouseBinaryFormatReader.java) should also extends `Iterator` or `Stream` interface, that will make handle large result set very easy to use Java Stream API.

### Code example

Maybe this will be a use case:

```java
QueryResponse response = this.client.query("select * from mytable").get();
ClickHouseBinaryFormatReader reader = client.newBinaryFormatReader(response);

// if ClickHouseBinaryFormatReader extends Stream, we can use
reader.map(....).filter(...)

// if ClickHouseBinaryFormatReader extends Iterator, we can use
StreamSupport.stream(Spliterators.spliteratorUnknownSize(reader, 0), false).map(...).filter(...)
```

Thanks.

abcfy2:
Fix #1963

## Summary

Make ClickHouseBinaryFormatReader also extend Iterator interface.

### DIFF
--- a/client-v2/src/main/java/com/clickhouse/client/api/data_formats/ClickHouseBinaryFormatReader.java
+++ b/client-v2/src/main/java/com/clickhouse/client/api/data_formats/ClickHouseBinaryFormatReader.java
@@ -16,11 +16,12 @@ import java.time.Instant;
 import java.time.LocalDate;
 import java.time.LocalDateTime;
 import java.time.ZonedDateTime;
+import java.util.Iterator;
 import java.util.List;
 import java.util.Map;
 import java.util.UUID;
 
-public interface ClickHouseBinaryFormatReader extends AutoCloseable {
+public interface ClickHouseBinaryFormatReader extends AutoCloseable, Iterator<Map<String, Object>> {
 
     /**
      * Reads a single value from the stream.
@@ -48,6 +49,7 @@ public interface ClickHouseBinaryFormatReader extends AutoCloseable {
      *
      * @return
      */
+    @Override
     boolean hasNext();
 
     /**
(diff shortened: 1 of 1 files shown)


chernser:
Good day, @abcfy2 ! 
However I do not plan to make reader as iterator. There is `com.clickhouse.client.api.Client#queryRecords(java.lang.String)` designed to be iterable reader. Would it work for your usecase? 

Thanks!

abcfy2:
I think it also makes sense to make ClickHouseBinaryFormatReader extend `Iterator` interface. Because `ClickHouseBinaryFormatReader` supports some methods like `getBoolean, getBigInteger` and so on.

abcfy2:
@chernser By the way, when will the new version release? I need the `queryRecords with prepared statement` feature.

I find `0.7.1` does not support this yet, but the code has been already added in the main branch.

chernser:
@abcfy2 the new version is planned by end of the December.
The prepared statements are done already in Nightly build, I think. (here is how to use nightly build https://gist.github.com/chernser/b4eacfde70093847449e3aef8e51ae8e) 

Are you looking for this method, right?:
```
public CompletableFuture<Records> queryRecords(String sqlQuery, Map<String, Object> params, QuerySettings settings)
```

abcfy2:
> @abcfy2 the new version is planned by end of the December. The prepared statements are done already in Nightly build, I think. (here is how to use nightly build https://gist.github.com/chernser/b4eacfde70093847449e3aef8e51ae8e)
> 
> Are you looking for this method, right?:
> 
> ```
> public CompletableFuture<Records> queryRecords(String sqlQuery, Map<String, Object> params, QuerySettings settings)
> ```

Yes. That's what I need. Current it's not in `0.7.1`. Waiting for `0.7.2`

PR #1992, opened 2024-12-02
Tweak version handling

Paultagoras:
## Checklist


jdbc-v2/src/main/java/com/clickhouse/jdbc/Driver.java
```diff
@@ -54,12 +58,32 @@ public static String getFrameworksDetected() {
         //If the version is not available, set it to 1.0
         if (tempDriverVersion == null || tempDriverVersion.isEmpty()) {
             log.warn("ClickHouse JDBC driver version is not available");
-            tempDriverVersion = "1.0";
+            tempDriverVersion = "1.0.0";
```
chernser, line 61: we need to keep static block as short as possible. 
In this case I would create separate static method and call it here.

PR #2004, opened 2024-12-06
Make GenericRecord and ClickHouseBinaryFormatReader easy to iterate

abcfy2:
## Summary

Add some useful methods to `GenericRecord` and `ClickHouseBinaryFormatReader` to make it easy to iterate.

Both `GenericRecord` and `ClickHouseBinaryFormatReader` could also get the column size, column names, and the full row values.

Use case:

```java
Records records = client.queryRecords("select * from mytable limit 100").get();
records.forEach(record -> {
    // easy to get column size, then user can use record.getObject(<index>) to get a specified column value
    System.out.println(record.getSize());

    // easy to get column names, then user can use record.getObject(<column_name>) to get a specified column value
    System.out.println(record.getNames());

    // easy to get all row values in kv format
    System.out.println(record.getAll());
});
```


client-v2/src/main/java/com/clickhouse/client/api/data_formats/internal/AbstractBinaryFormatReader.java
```diff
@@ -737,4 +732,19 @@ public ClickHouseBitmap getClickHouseBitmap(int index) {
     public void close() throws Exception {
         input.close();
     }
+
+    @Override
+    public int getColumnCount() {
+        return currentRecord.size();
```
chernser, line 738: It should be returned from schema because it may be a virtual columns (for example, when we do so for tuples)

client-v2/src/main/java/com/clickhouse/client/api/data_formats/internal/AbstractBinaryFormatReader.java
```diff
@@ -737,4 +732,19 @@ public ClickHouseBitmap getClickHouseBitmap(int index) {
...
     }
+
+    @Override
+    public int getColumnCount() {
+        return currentRecord.size();
+    }
+
+    @Override
+    public Set<String> getColumnNames() {
+        return currentRecord.keySet();
```
chernser, line 743: It should come from schema because it may have additional columns.

client-v2/src/main/java/com/clickhouse/client/api/data_formats/internal/AbstractBinaryFormatReader.java
```diff
@@ -737,4 +732,19 @@ public ClickHouseBitmap getClickHouseBitmap(int index) {
...
+    }
+
+    @Override
+    public Map<String, Object> current() {
+        return currentRecord;
```
chernser, line 748: This cannot be exposed because it is internal records and should not be accessed in such way. I plan to change internal later.

abcfy2:
So the new use case should be:

```java
records.forEach(record -> {
    TableSchema schema = record.getSchema();
    // easy to get column size, then user can use record.getObject(<index>) to get a specified column value
    System.out.println(schema.getColumns().size());

    // easy to get column names, then user can use record.getObject(<column_name>) to get a specified column value
    System.out.println(schema.getColumns().stream().map(ClickHouseColumn::getColumnName).toList());

    // easy to get all row values in kv format
    System.out.println(record.getAll());
});
```

client-v2/src/main/java/com/clickhouse/client/api/query/GenericRecord.java
```diff
@@ -500,4 +496,17 @@ public interface GenericRecord {
...
     ClickHouseBitmap getClickHouseBitmap(int index);
+
+    TableSchema getSchema();
+
+    /**
+     * Returns all values of this record in Map.
+     *
+     * @return a Map of column names and values.
+     */
+    default Map<String, Object> getAll() {
```
chernser, line 507: lets rename it to `getValues` or `asMap`. 
I would prefer having no default implementation in interfaces because `default` was designed to not break things when new stream API was introduced in java. I think, that only this case is valid usage of `default`.

chernser:
@abcfy2 than you for changing the PR! 
I've left 2 comments. I mainly worried about name of the method. 

Thanks!

abcfy2:
No worries @chernser . I've modified PR. Please check again.

Thanks.

chernser:
@abcfy2 looks good to me! 
Thank you!

abcfy2:
:raised_eyebrow: Some tests failed. But seems it's non of my mistake.

chernser (approved):

PR #2008, opened 2024-12-08
[jdbc-v2] JDBC Properties configuration 

chernser:
## Summary
Implements 
- merging properties from url 
- driver properties
- passing properties to client builder 
- 
## Checklist


jdbc-v2/src/test/java/com/clickhouse/jdbc/internal/JdbcConfigurationTest.java
```diff
@@ -1,31 +1,102 @@
...
-        url = "jdbc:clickhouse:https://localhost:8443/clickhouse?param1=value1&param2=value2";
-        cleanUrl = configuration.cleanUrl(url);
-        assertEquals(cleanUrl, "https://localhost:8443/clickhouse?param1=value1&param2=value2");
+public class JdbcConfigurationTest {
+
+    @Test(dataProvider = "testConnectionUrlDataProvider")
+    public void testConnectionUrl(String jdbcUrl, String connectionUrl, Properties properties, Map<String, String> expectedClientProps) throws Exception {
+        JdbcConfiguration configuration = new JdbcConfiguration(jdbcUrl, properties);
+        assertEquals(configuration.getConnectionUrl(), connectionUrl);
+        System.out.println(configuration.clientProperties);
```
mzitnik, line 25: Do you want to convert into logger instead of System.out

jdbc-v2/src/main/java/com/clickhouse/jdbc/internal/JdbcConfiguration.java
```diff
@@ -55,77 +39,206 @@ public boolean isIgnoreUnsupportedRequests() {
...
+     * Returned url has only schema and authority and doesn't have query parameters or path.
+     * JDBC URL should have only a single path parameter to specify database name.
+     * Note: Some BI tools do not let pass JDBC URL, so ssl is passed as property.
+     * @param url - JDBC url
+     * @param ssl - if SSL protocol should be used when protocol is not specified
+     * @return URL without JDBC prefix
+     */
+    static String createConnectionURL(String url, boolean ssl) throws SQLException {
         if (url.startsWith("//")) {
-            if (setSSL) {
```
mzitnik, line 85: Maybe i'am missing it. This we did for logic if we have properties like in tableau? are we covering it

chernser: yes, 
this method is simply builds string. `ssl` is now external parameter - it will be set to the value of `ssl` property once all properties from user and url are merged.

chernser: I have a test for ssl connection and have tests where ssl is passed as property.

jdbc-v2/src/test/java/com/clickhouse/jdbc/internal/JdbcConfigurationTest.java
chernser: I will remove it.

jdbc-v2/src/main/java/com/clickhouse/jdbc/internal/JdbcConfiguration.java
mzitnik: Great thanks for the clarification

mzitnik (approved):

PR #2013, opened 2024-12-12
Updating to address some seen issues/bugs in DataGrip testing

Paultagoras:
When testing against Datagrip I ran into some issues, this resolves them.


client-v2/src/main/java/com/clickhouse/client/api/data_formats/internal/AbstractBinaryFormatReader.java
```diff
@@ -618,7 +618,12 @@ public ClickHouseGeoMultiPolygonValue getGeoMultiPolygon(int index) {
 
     @Override
     public <T> List<T> getList(int index) {
-        return readValue(index);
+        try {
+            BinaryStreamReader.ArrayValue array = readValue(index);
+            return array.asList();
+        } catch (ClassCastException e) {
+            throw new ClientException("Column is not of array type", e);
```
chernser, line 625: would you please,  change to `        return getList(schema.indexToName(index));` it will call getList with the type check?
it will eliminate code duplication.

chernser (approved):

PR #2015, opened 2024-12-12
[client-v2] Fix check for ConnectTimeoutException in shouldRetry

janeklb:
## Summary
The [catch block accepts `ConnectTimeoutException`](https://github.com/ClickHouse/clickhouse-java/blob/02542cc5b7a2f9ad09a816afd25308ea7da57519/client-v2/src/main/java/com/clickhouse/client/api/Client.java#L1411); however, `shouldRetry` checks for `ConnectException`.

I think the solution is for `shouldRetry` to test for `ConnectTimeoutException`, please let me know if that's incorrect so that i can adjust the `catch` clause instead.

It's also worth noting that `wrapException` tests for `ConnectException`; however, `wrapException` will never only ever be called with `NoHttpResponseException | ConnectionRequestTimeoutException | ConnectTimeoutException`, which makes me wonder if:
- `ConnectException` there should be replaced with `NoHttpResponseException`
- _or_ `ConnectException` should be added to the multi-catch

## Checklist

### DIFF
--- a/client-v2/src/main/java/com/clickhouse/client/api/internal/HttpAPIClientHelper.java
+++ b/client-v2/src/main/java/com/clickhouse/client/api/internal/HttpAPIClientHelper.java
@@ -583,7 +583,7 @@ public class HttpAPIClientHelper {
             return retryCauses.contains(ClientFaultCause.NoHttpResponse);
         }
 
-        if (ex instanceof ConnectException) {
+        if (ex instanceof ConnectTimeoutException) {
             return retryCauses.contains(ClientFaultCause.ConnectTimeout);
         }
 
--- a/client-v2/src/main/java/com/clickhouse/client/api/internal/HttpAPIClientHelper.java
+++ b/client-v2/src/main/java/com/clickhouse/client/api/internal/HttpAPIClientHelper.java
@@ -583,7 +583,7 @@ public class HttpAPIClientHelper {
             return retryCauses.contains(ClientFaultCause.NoHttpResponse);
         }
 
-        if (ex instanceof ConnectTimeoutException) {
+        if (ex instanceof ConnectException || ex instanceof ConnectTimeoutException) {
             return retryCauses.contains(ClientFaultCause.ConnectTimeout);
         }
 
@@ -598,6 +598,7 @@ public class HttpAPIClientHelper {
     // ClientException will be also wrapped
     public ClientException wrapException(String message, Exception cause) {
         if (cause instanceof ConnectionRequestTimeoutException ||
+                cause instanceof NoHttpResponseException ||
                 cause instanceof ConnectTimeoutException ||
                 cause instanceof ConnectException) {
             return new ConnectionInitiationException(message, cause);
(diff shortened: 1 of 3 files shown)


client-v2/src/main/java/com/clickhouse/client/api/internal/HttpAPIClientHelper.java
```diff
@@ -583,7 +583,7 @@ public boolean shouldRetry(Exception ex, Map<String, Object> requestSettings) {
             return retryCauses.contains(ClientFaultCause.NoHttpResponse);
         }
 
-        if (ex instanceof ConnectException) {
+        if (ex instanceof ConnectTimeoutException) {
```
chernser, line 586: I think they should be both: 
- ConnectException - means that socket failed to connect. When we implement TCP protocol support we might use this exception 
- org.apache.hc.client5.http.ConnectTimeoutException is child of java.net.SocketTimeoutException - means that throws when read or accept (according to the JavaDoc of the last one). And ConnectTimeoutException is client specific exception. It will not be thrown by other client.

chernser:
Good day, @janeklb ! 

You have brought a very good question and I almost agree with your fix. 

ConnectionException should be in multi-catch. 
ConnectionException is about connection and thrown by low-level socket, NoHttpResponseException happens when stale connection is used and Apache Http Client detects it. 

When handling exceptions we are trying to be more generic and we try to avoid mentioning very specific exceptions. In this case both exceptions should be handles - unfortunately AHC 's ConnectTimeoutException doesn't derive from ConnectionException - I think this is wrong impl. 

Please see my comment in the code.

client-v2/src/main/java/com/clickhouse/client/api/internal/HttpAPIClientHelper.java
janeklb: > ConnectException - means that socket failed to connect. When we implement TCP protocol support we might use this exception

I agree that this is a useful exception to check for, but this particular if block then proceeds to check `retryCauses.contains(ClientFaultCause.ConnectTimeout)`, and so `ConnectException` doesn't seem applicable (to `ClientFaultCause.ConnectTimeout`)

janeklb:
> ConnectionException should be in multi-catch.

I've made this change in `Client.java`.

Do you think `wrapException` should be updated (to handle `NoHttpResponseException`)?

janeklb:
> Good day, @janeklb !

And good day to you too @chernser !

client-v2/src/main/java/com/clickhouse/client/api/internal/HttpAPIClientHelper.java
chernser: It is applicable - because ClientFaultCause.ConnectTimeout is about any error while connection initiation. 
ConnectException is all about it:
```
/**
 * Signals that an error occurred while attempting to connect a
 * socket to a remote address and port.  Typically, the connection
 * was refused remotely (e.g., no process is listening on the
 * remote address/port).
 *
 * @since   1.1
 */
public class ConnectException extends SocketException 
```

chernser:
> Do you think wrapException should be updated (to handle NoHttpResponseException)?
Lets do it - it is connection time exception in most cases.

janeklb:
Alrighty. I've pushed the changes I think you were suggesting

chernser:
Thank you!

chernser (approved):

PR #2029, opened 2024-12-17
[client-v2] Initial attempt at adding some server context to the client

### LINKED ISSUE #2027: [client-v2] Implement loading server context
### Description 
There is a need to load some server context after connecting to a server. Here is list of most important information:
- server timezone - it should be loaded to do proper DateTime conversion 
- user - it is important to know actual user if for example Bearer auth is used
- version - it is important to know when certain features are not working on older server versions and for debug purposes

Paultagoras:
## Summary
Closes #2027 
Adds a call to gather currentUser(), timezone(), and version() for the client

## Checklist
- [x] Closes issue 

### DIFF
--- a/client-v2/src/main/java/com/clickhouse/client/api/Client.java
+++ b/client-v2/src/main/java/com/clickhouse/client/api/Client.java
@@ -155,7 +155,6 @@ public class Client implements AutoCloseable {
     private final ColumnToMethodMatchingStrategy columnToMethodMatchingStrategy;
 
     // Server context
-    private String serverUser;
     private String serverVersion;
 
     private Client(Set<String> endpoints, Map<String,String> configuration, boolean useNewImplementation,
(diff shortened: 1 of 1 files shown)


client-v2/src/main/java/com/clickhouse/client/api/Client.java
```diff
@@ -154,6 +154,10 @@ public class Client implements AutoCloseable {
 
     private final ColumnToMethodMatchingStrategy columnToMethodMatchingStrategy;
 
+    // Server context
+    private String serverUser;
+    private String serverVersion;
```
chernser, line 159: please add `getServerVersion()`

chernser: for the use we need to update `com.clickhouse.client.api.Client#getUser`

chernser (changes requested):
please correct getters

chernser (approved):

mzitnik (commented):
General comment i think we should a way to disable if the user wants

PR #2030, opened 2024-12-17
No-op flags

Paultagoras:
## Summary
We should obey the no-op flag for unsupported features where possible.

## Checklist
- [x] Closes issue 

### DIFF
--- a/jdbc-v2/src/test/java/com/clickhouse/jdbc/StatementTest.java
+++ b/jdbc-v2/src/test/java/com/clickhouse/jdbc/StatementTest.java
@@ -4,6 +4,8 @@ import com.clickhouse.client.api.query.GenericRecord;
 import org.testng.Assert;
 import org.testng.annotations.Test;
 
+import java.net.Inet4Address;
+import java.net.Inet6Address;
 import java.sql.Array;
 import java.sql.Connection;
 import java.sql.Date;
@@ -427,5 +429,24 @@ public class StatementTest extends JdbcIntegrationTest {
         assertEquals(StatementImpl.parseStatementType(null), StatementImpl.StatementType.OTHER);
         assertEquals(StatementImpl.parseStatementType(""), StatementImpl.StatementType.OTHER);
         assertEquals(StatementImpl.parseStatementType("      "), StatementImpl.StatementType.OTHER);
+    }
+
+
+    @Test(groups = { "integration" })
+    public void testWithIPs() throws Exception {
+        try (Connection conn = getJdbcConnection()) {
+            try (Statement stmt = conn.createStatement()) {
+                try (ResultSet rs = stmt.executeQuery("SELECT toIPv4('127.0.0.1'), toIPv6('::1'), toIPv6('2001:438:ffff::407d:1bc1')")) {
+                    assertTrue(rs.next());
+                    assertEquals(rs.getString(1), "/127.0.0.1");
+                    assertEquals(rs.getObject(1), Inet4Address.getByName("127.0.0.1"));
+                    assertEquals(rs.getString(2), "/0:0:0:0:0:0:0:1");
+                    assertEquals(rs.getObject(2), Inet6Address.getByName("0:0:0:0:0:0:0:1"));
+                    assertEquals(rs.getString(3), "/2001:438:ffff:0:0:0:407d:1bc1");
+                    assertEquals(rs.getObject(3), Inet6Address.getByName("2001:438:ffff:0:0:0:407d:1bc1"));
+                    assertFalse(rs.next());
+                }
+            }
+        }
     }
 }


jdbc-v2/src/main/java/com/clickhouse/jdbc/ConnectionImpl.java
```diff
@@ -118,14 +118,22 @@ public PreparedStatement prepareStatement(String sql) throws SQLException {
     @Override
     public CallableStatement prepareCall(String sql) throws SQLException {
         checkOpen();
-        throw new SQLFeatureNotSupportedException("CallableStatement not supported", ExceptionUtils.SQL_STATE_FEATURE_NOT_SUPPORTED);
+        if (!config.isIgnoreUnsupportedRequests()) {
```
chernser, line 121: I would think this is totally valid exception because if this is called then NPE will be thrown anyway.
In most cases we can ignore parameters, but not a whole functionality.

jdbc-v2/src/main/java/com/clickhouse/jdbc/ConnectionImpl.java
```diff
@@ -118,14 +118,22 @@ public PreparedStatement prepareStatement(String sql) throws SQLException {
...
     @Override
     public String nativeSQL(String sql) throws SQLException {
         checkOpen();
         /// TODO: this is not implemented according to JDBC spec and may not be used.
-        throw new SQLFeatureNotSupportedException("nativeSQL not supported", ExceptionUtils.SQL_STATE_FEATURE_NOT_SUPPORTED);
+        if (!config.isIgnoreUnsupportedRequests()) {
+            throw new SQLFeatureNotSupportedException("nativeSQL not supported", ExceptionUtils.SQL_STATE_FEATURE_NOT_SUPPORTED);
+        }
+
+        return null;
```
chernser, line 136: may be return the same sql? + javadoc that it is not implemented according to the spec if flag is true

jdbc-v2/src/main/java/com/clickhouse/jdbc/ConnectionImpl.java
```diff
@@ -243,19 +251,29 @@ public PreparedStatement prepareStatement(String sql, int resultSetType, int res
     @Override
     public CallableStatement prepareCall(String sql, int resultSetType, int resultSetConcurrency) throws SQLException {
         checkOpen();
-        throw new SQLFeatureNotSupportedException("CallableStatement not supported", ExceptionUtils.SQL_STATE_FEATURE_NOT_SUPPORTED);
+        if (!config.isIgnoreUnsupportedRequests()) {
+            throw new SQLFeatureNotSupportedException("CallableStatement not supported", ExceptionUtils.SQL_STATE_FEATURE_NOT_SUPPORTED);
+        }
+
+        return null;
```
chernser, line 258: the same - this is whole functionality and it will be NPE anyway. so Exception should be thrown all the time

jdbc-v2/src/main/java/com/clickhouse/jdbc/ConnectionImpl.java
```diff
@@ -243,19 +251,29 @@ public PreparedStatement prepareStatement(String sql, int resultSetType, int res
...
 
     @Override
     public Map<String, Class<?>> getTypeMap() throws SQLException {
         checkOpen();
-        throw new SQLFeatureNotSupportedException("getTypeMap not supported", ExceptionUtils.SQL_STATE_FEATURE_NOT_SUPPORTED);
+        if (!config.isIgnoreUnsupportedRequests()) {
+            throw new SQLFeatureNotSupportedException("getTypeMap not supported", ExceptionUtils.SQL_STATE_FEATURE_NOT_SUPPORTED);
+        }
+
+        return null;
```
chernser, line 268: This kind of missing functionality. I would return an empty map at least to not cause NPE. 
But we need to create an issue and fill this properly

jdbc-v2/src/main/java/com/clickhouse/jdbc/ConnectionImpl.java
```diff
@@ -462,19 +514,27 @@ public String getSchema() throws SQLException {
...
+        if (!config.isIgnoreUnsupportedRequests()) {
+            throw new SQLFeatureNotSupportedException("setNetworkTimeout not supported", ExceptionUtils.SQL_STATE_FEATURE_NOT_SUPPORTED);
+        }
     }
 
     @Override
     public int getNetworkTimeout() throws SQLException {
         //TODO: Should this be supported?
-        throw new SQLFeatureNotSupportedException("getNetworkTimeout not supported", ExceptionUtils.SQL_STATE_FEATURE_NOT_SUPPORTED);
+        if (!config.isIgnoreUnsupportedRequests()) {
```
chernser, line 533: seems something missing

chernser: I mean it may be a useful timeout - we should take it from client.

chernser (approved):

PR #2033, opened 2024-12-18
Adding the types to be clearer

Paultagoras:
## Summary
* Return the full type with the ResultSet metadata (including the sub type)
* Return the CH type (rather than class name) for the Array

## Checklist
- [ ] Closes issue 


jdbc-v2/src/main/java/com/clickhouse/jdbc/ConnectionImpl.java
```diff
@@ -484,7 +484,7 @@ public Properties getClientInfo() throws SQLException {
     public Array createArrayOf(String typeName, Object[] elements) throws SQLException {
         try {
             // TODO: pass type name
-            return new com.clickhouse.jdbc.types.Array(List.of(elements), Types.OTHER);
+            return new com.clickhouse.jdbc.types.Array(List.of(elements), "", Types.OTHER);
```
chernser, line 487: I think we can determine type of the elements here and map it to certain types. 
JdbcUtils has such mapping alread.

chernser (approved):

Issue #2010, opened 2024-12-10
[client-v2] Inserting compressed data (using compressed input stream)


dizider:
### Describe your feedback
The client in version 2 allows to insert data from input byte stream. It also supports inserting data in compressed format (`decompress` option). I would like to be able to insert data from already compressed data stream. This change allows to create compressed in memory batches and then send them directly.

I have already looked at code and found that there is one major incompatibly with the current API - insert statement is a part of the request body and is added before each request. This is a problem because the whole body has to be compressed or not. So I think the options are to send the statement as query parameter of HTTP or to allow user defined insert statements. 

### Code example
Here are two examples of how I would like to use the client. The second example is my actual use case - create compressed batches in memory and then sending them. Creating these compressed batches should be more memory efficient which can be used in memory heavy applications.

DISCLAIMER: The following examples do not work with current implementation! they assume that insert query will be sent as an HTTP query parameter.

Compressed stream
```java
void compressedStream() {
	final String table = "test_table";  
	final ClickHouseFormat format = ClickHouseFormat.CSV;  
	try (Client client = new Client.Builder()  
	        .compressClientRequest(false)  
	        .compressServerResponse(false)  
	        .addEndpoint(Protocol.HTTP, "localhost", clickhouse.getMappedPort(8123), false)  
	        .setUsername(clickhouse.getUsername())  
	        .setPassword(clickhouse.getPassword())  
	        .useAsyncRequests(true)  
	        .build()) {  
	  
	    final var pipedOutputStream = ClickHouseDataStreamFactory.getInstance().createPipedOutputStream(new ClickHouseConfig());  
	    final OutputStream compressedOutputStream = new ClickHouseLZ4OutputStream(pipedOutputStream, LZ4Factory.fastestInstance().fastCompressor(), 8192);  
	    
	    final var futureResponse = client.insert(table, pipedOutputStream.getInputStream(), format, new InsertSettings().serverSetting("decompress", "1"));  
	  
	    // write data to insert to compressedOutputStream  
	    final int numberOfRows = 2;  
	    compressedOutputStream.write("1,foo\n".getBytes());  
	    compressedOutputStream.write("2,bar\n".getBytes());  
	    compressedOutputStream.close();  
	    pipedOutputStream.close();  
	  
	    // insert setting tells ClickHouse that data are compressed  
	    // but the insert doesn't work, because the `insert` method prepend the insert statement (and it isn't compressed)
	    try (var response = futureResponse.join()) {  
[...]

PR #2034, opened 2024-12-18
[client-v2] V2 implement writer api

chernser:
## Summary
Implements new Writer style API for insert operations. 
- Handles low-level writes
- Has advanced writer with set-then-commit style 
- Fixes some issues with date/datetime serialization (more convertible types) 

Writer API intended to work:
- when DTO class requires more precise control over serialization
- when data is compressed by application by a supported algorithm
- when data source is collection (queue, list) so wrapping with InputStream is too complex. 

Main change: 
- `com.clickhouse.client.api.Client#insert(java.lang.String, com.clickhouse.client.api.DataStreamWriter, com.clickhouse.data.ClickHouseFormat, com.clickhouse.client.api.insert.InsertSettings)` - intriduces ability to use data stream writer that can control how to write data. 

Current PR introduces two new experimental classes:
- `com.clickhouse.client.api.data_formats.RowBinaryFormatWriter` - Intend to simplify writing RowBinary encoded data 
- `com.clickhouse.client.api.data_formats.RowBinaryFormatSerializer` - very low-level serializer for RowBinary encoding. For very advanced users. 

## Checklist
- [x] Closes: https://github.com/ClickHouse/clickhouse-java/issues/2010


client-v2/src/main/java/com/clickhouse/client/api/Client.java
```diff
@@ -1479,7 +1453,49 @@ public CompletableFuture<InsertResponse> insert(String tableName, InputStream da
...
+                                                    ClickHouseFormat format,
+                                                    InsertSettings settings) {
+
+        final int writeBufferSize = settings.getInputStreamCopyBufferSize() <= 0 ?
+                Integer.parseInt(configuration.getOrDefault(ClientConfigProperties.CLIENT_NETWORK_BUFFER_SIZE.getKey(),
+                        ClientConfigProperties.CLIENT_NETWORK_BUFFER_SIZE.getDefaultValue())) :
+                settings.getInputStreamCopyBufferSize();
+
+        if (writeBufferSize <= 0) {
+            throw new IllegalArgumentException("Buffer size must be greater than 0");
```
mzitnik, line 1466: Why we can not do that in `InsertSettings` and throw the exception from there?

client-v2/src/main/java/com/clickhouse/client/api/ClientConfigProperties.java
```diff
@@ -104,7 +104,7 @@ public enum ClientConfigProperties {
 
     QUERY_ID("query_id"), // actually a server setting, but has client effect too
 
-    CLIENT_NETWORK_BUFFER_SIZE("client_network_buffer_size"),
+    CLIENT_NETWORK_BUFFER_SIZE("client_network_buffer_size", "8192"),
```
mzitnik, line 107: lets extract it to a const

mzitnik (approved):

client-v2/src/main/java/com/clickhouse/client/api/ClientConfigProperties.java
chernser: Done -> `com.clickhouse.client.api.Client.Builder#DEFAULT_BUFFER_SIZE`

client-v2/src/main/java/com/clickhouse/client/api/Client.java
chernser: this value can be set by two method `setInputStreamCopyBufferSize` and `setOption` and here I'm doing specific calculation. 
One more reason - I keep validation here because it is implementation specific code for this operation. InsertSettings should not be aware of this validation because they may be used for other purposes.

PR #2041, opened 2024-12-19
[jdbc-v2] Fixed closing prev result set

chernser:
## Summary
Fixes closing prev. result set in `execureQuery` method of `StatementImpl`. 
Also fixed passing `SET` queries to server and not just handling roles staff. 

## Checklist
- [ ] Closes issue 


mzitnik (approved):

jdbc-v2/src/main/java/com/clickhouse/jdbc/StatementImpl.java
```diff
@@ -289,6 +301,7 @@ private boolean execute(String sql, QuerySettings settings) throws SQLException
             executeQuery(sql, settings); // keep open to allow getResultSet()
             return true;
         } else if(type == StatementType.SET) {
+            executeUpdate(sql, settings);
```
slvrtrn, line 304: @chernser 
I don't think that we need to execute this update; the role URL setting is enough, as the HTTP interface is stateless.

Issue #2039, opened 2024-12-19
[jdbc-v2] getObject(Type) should try and do some type conversion


Paultagoras:
Right now we ignore when a Type is passed and just return the original - we should TRY and do some basic conversion.

Issue #1409, opened 2023-08-01
Cannot use ResultSet.getObject to retrieve java.sql.Date (JDBC specification violation?)


LDVSOFT:
### Describe the bug
At some point ClickHouse JDBC driver switched to using `java.time.` types to represent temporal values. While I salute this change and would prefer too to abandon `java.sql.Date`, the way it was done, unfortunately, breaks some end consumers that expect old types to be returned by default. While originally we detected it by trying to read data from ClickHouse to Spark (by using the latter), I was able to provide a more direct counterexample.

### Steps to reproduce
1. Setup a JDBC connection using driver version _0.4.6_,
1. Perform a `SELECT d FROM some_table` where `d` has type `Date` or similar (nullable or array),
2. Access it via `resultSet.getObject("d")` or `resultSet.getObject("d", java.sql.Date.class)`.
   * First will give `java.time.LocalDate`, that *could be correct* under some type maps but not the default one,
   * Second one fails with a `ClassCastException`, that in my eyes violates JDBC specification directly.

To be exact, I am referencing to JDBC 4.3 specification, appendix B (quotation with my style):
> ## B.3 JDBC Types Mapped to Java Object Types
> `ResultSet.getObject` and `CallableStatement.getObject` use the mapping shown in TABLE B-3 for standard mappings.
> | JDBC Type | Java Object Type |  
> | --------- | ---------------- |
> | _skipped_ | _skipped_        |
> | `DATE`    | `java.sql.Date`  | 
> | _skipped_ | _skipped_        |
All `ResultSet.getObject` methods that don't take a type map directly tell in their documentation that they follow the mapping above.

Related:
* `ResultSet.getDate` works as expected.
* Arrays of dates (accessed via `resultSet.getArray("ds").getArray()`) are broken the same way (`java.time.LocalDate[]` instead of `java.sql.Date[]`). 
* I've not tested this yet but I'd expect the same problem to arise for JDBC `TIMESTAMP` with `java.time.Instant` vs `java.sql.Timestamp`. 
* I've not tested other JDBC methods, like setting `PreparedStatement` parameters, but I would check their behaviour too.

I've checked the behaviour of other JDBC drivers against the respectful databases I use, including PostgreSQL, and they work as expected. Older ClickHouse JDBC drvier (I guess around _0.3.1_) worked for us.

### Expected behaviour
I guess the right thing to do would be one or more of:
[...]

PR #2045, opened 2024-12-20
Cleanup + Type Conversion

Paultagoras:
## Summary
* Cleaning up methods that use columnIndex to use columnLabel instead (so that we can edit in one place)
* Adding some type conversion methods for convenience in JDBC
* Tweaking to use JDBCType enum rather than just the int values
* Closes https://github.com/ClickHouse/clickhouse-java/issues/2039

## Checklist
- [x] Closes issue 

### DIFF
--- a/jdbc-v2/src/test/java/com/clickhouse/jdbc/DataTypeTests.java
+++ b/jdbc-v2/src/test/java/com/clickhouse/jdbc/DataTypeTests.java
@@ -8,7 +8,6 @@ import org.testng.annotations.Test;
 
 import java.math.BigDecimal;
 import java.math.BigInteger;
-import java.sql.Blob;
 import java.sql.Connection;
 import java.sql.Date;
 import java.sql.DriverManager;
@@ -18,10 +17,8 @@ import java.sql.ResultSet;
 import java.sql.SQLException;
 import java.sql.Statement;
 import java.sql.Timestamp;
-import java.sql.Types;
 import java.time.LocalDate;
 import java.time.LocalDateTime;
-import java.time.LocalTime;
 import java.time.OffsetDateTime;
 import java.time.ZoneId;
 import java.time.ZonedDateTime;


jdbc-v2/src/test/java/com/clickhouse/jdbc/DataTypeTests.java
```diff
@@ -846,4 +855,64 @@ public void testDynamicTypesSimpleStatement() throws SQLException {
...
+                    assertEquals(String.valueOf(rs.getObject(2, new HashMap<String, Class<?>>(){{put(JDBCType.BOOLEAN.getName(), Boolean.class);}})), "true");
+
+                    assertEquals(rs.getFloat(3), 1.0f);
+                    assertEquals(String.valueOf(rs.getObject(3)), "1.0");
+                    assertEquals(rs.getObject(3, Float.class), 1.0f);
+                    assertEquals(rs.getObject(3, Double.class), 1.0);
+                    assertEquals(String.valueOf(rs.getObject(3, new HashMap<String, Class<?>>(){{put(JDBCType.FLOAT.getName(), Float.class);}})), "1.0");
+
+                    assertEquals(rs.getDate(4), Date.valueOf("2024-12-01"));
+                    assertTrue(rs.getObject(4) instanceof ZonedDateTime);//ZonedDateTime
```
chernser, line 885: This is not true - according to spec we need to return sql.Date or sql.Timestamp

jdbc-v2/src/test/java/com/clickhouse/jdbc/DataTypeTests.java
```diff
@@ -846,4 +855,64 @@ public void testDynamicTypesSimpleStatement() throws SQLException {
...
+                    assertEquals(String.valueOf(rs.getObject(4)), "2024-12-01T00:00Z[UTC]");//ZonedDateTime
```
chernser, line 886: here we need do `rs.getString()` . Current assertion doesn't reflect how we convert date to string

jdbc-v2/src/test/java/com/clickhouse/jdbc/DataTypeTests.java
```diff
@@ -846,4 +855,64 @@ public void testDynamicTypesSimpleStatement() throws SQLException {
...
+                    assertEquals(rs.getObject(4, LocalDate.class), LocalDate.of(2024, 12, 1));
```
chernser, line 887: should be also conversion to ZonedDateTime may be

jdbc-v2/src/test/java/com/clickhouse/jdbc/DataTypeTests.java
```diff
@@ -846,4 +855,64 @@ public void testDynamicTypesSimpleStatement() throws SQLException {
...
+                    assertEquals(String.valueOf(rs.getObject(4, new HashMap<String, Class<?>>(){{put(JDBCType.DATE.getName(), LocalDate.class);}})), "2024-12-01");
```
chernser, line 888: This is cool check! I like it.

jdbc-v2/src/test/java/com/clickhouse/jdbc/DataTypeTests.java
```diff
@@ -846,4 +855,64 @@ public void testDynamicTypesSimpleStatement() throws SQLException {
...
+
+                    assertEquals(rs.getTimestamp(5), Timestamp.valueOf("2024-12-01 12:34:56"));
+                    assertTrue(rs.getObject(5) instanceof ZonedDateTime);//ZonedDateTime
```
chernser, line 891: here is related issue https://github.com/ClickHouse/clickhouse-java/issues/1409 
`getObject()` always return some default JDBC objects. We cannot change it.

jdbc-v2/src/main/java/com/clickhouse/jdbc/ResultSetImpl.java
```diff
@@ -1221,50 +1153,38 @@ public Statement getStatement() throws SQLException {
...
-        checkClosed();
-        return getArray(reader.getSchema().columnIndexToName(columnIndex));
+        return getArray(getSchema().columnIndexToName(columnIndex));
     }
 
     @Override
     public Object getObject(String columnLabel, Map<String, Class<?>> map) throws SQLException {
         checkClosed();
-        return getObject(columnLabel);
+        log.debug("getObject(columnLabel={}, map={})", columnLabel, map);
```
chernser, line 1182: I'd make it `trace` or remove at all - logs will be flooded

jdbc-v2/src/main/java/com/clickhouse/jdbc/ResultSetImpl.java
```diff
@@ -1221,50 +1153,38 @@ public Statement getStatement() throws SQLException {
...
+        if (map == null) {
```
chernser, line 1183: Just a note: 
In such cases I usually try to use one way of getting result. I would have a default map of type mappings, so only one generic method is used.

chernser (approved):

Issue #2007, opened 2024-12-08
Fix version detection


Paultagoras:
Problems to solve:
- Reports a wrong client version or update didn’t happen.
- Reports never released kafka connector version (`v1.2.10`)
- The existing tests haven't caught the problem

Currently we're defaulting to 0.6.3:

public static final String LATEST_KNOWN_VERSION = "0.6.3";

We need to find a more reliable way to determine the version...

mshustov:
or don't guess and set a well-known indicator like `unknown`, which would help us to detect problems earlier and fix them

chernser:
@mshustov we need to know at least aprox. version - so I've added `~` to version constant. 
Event if we missed updating it - we will get something `~0.7.2`. 
Please let me know if this will work. 

Thanks!

PR #2054, opened 2024-12-27
[client-v2] Fix user agent

chernser:
## Summary
Fixes several issues:
- client-v2 old `product_name` property should be alias to a new one `client_name`
- client-v2 client name update 
- jdbc-v2 proper setting client name thru application name 

## Checklist
- [X] Closes issue https://github.com/ClickHouse/clickhouse-java/issues/2007

### DIFF
--- a/clickhouse-http-client/src/test/java/com/clickhouse/client/http/ClickHouseHttpClientTest.java
+++ b/clickhouse-http-client/src/test/java/com/clickhouse/client/http/ClickHouseHttpClientTest.java
@@ -174,6 +174,7 @@ public class ClickHouseHttpClientTest extends ClientIntegrationTest {
                          .executeAndWait()) {
                 if (response.records().iterator().hasNext()) {
                     String result = response.firstRecord().getValue(0).asString();
+                    System.out.println(result);
                     Assert.assertTrue(result.startsWith(optionValue + " ClickHouse-JavaClient/"));
                     Assert.assertTrue(result.indexOf("Http") > 0);
                     break;
--- a/clickhouse-http-client/src/test/java/com/clickhouse/client/http/ClickHouseHttpClientTest.java
+++ b/clickhouse-http-client/src/test/java/com/clickhouse/client/http/ClickHouseHttpClientTest.java
@@ -174,7 +174,6 @@ public class ClickHouseHttpClientTest extends ClientIntegrationTest {
                          .executeAndWait()) {
                 if (response.records().iterator().hasNext()) {
                     String result = response.firstRecord().getValue(0).asString();
-                    System.out.println(result);
                     Assert.assertTrue(result.startsWith(optionValue + " ClickHouse-JavaClient/"));
                     Assert.assertTrue(result.indexOf("Http") > 0);
                     break;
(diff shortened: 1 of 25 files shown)


clickhouse-http-client/src/test/java/com/clickhouse/client/http/ClickHouseHttpClientTest.java
```diff
@@ -174,6 +174,7 @@ private void testUserAgent(ClickHouseOption option, String optionValue) throws E
                          .executeAndWait()) {
                 if (response.records().iterator().hasNext()) {
                     String result = response.firstRecord().getValue(0).asString();
+                    System.out.println(result);
```
mzitnik, line 177: Lets remove it

client-v2/src/main/java/com/clickhouse/client/api/Client.java
```diff
@@ -2186,8 +2160,28 @@ public void updateClientName(String name) {
         this.configuration.put(ClientConfigProperties.CLIENT_NAME.getKey(), name);
     }
 
+    public static final String clientVersion = getPackageVersion();
+    public static final String CLIENT_USER_AGENT = "clickhouse-java-v2/";
+
     private Collection<String> unmodifiableDbRolesView = Collections.emptyList();
 
+    private static String getPackageVersion() {
```
mzitnik, line 2168: This code is duplicated. Can we use some utility function

client-v2/src/main/java/com/clickhouse/client/api/internal/HttpAPIClientHelper.java
```diff
@@ -646,6 +644,56 @@ public static Map<String, String> parseUrlParameters(URL url) {
         return params;
     }
 
+
+    private void correctUserAgentHeader(HttpRequest request, Map<String, Object> requestConfig) {
```
mzitnik, line 648: I guess it will be called every request why not prepare it once?

mzitnik (approved):

mzitnik (approved):

clickhouse-http-client/src/test/java/com/clickhouse/client/http/ClickHouseHttpClientTest.java
chernser: will do

client-v2/src/main/java/com/clickhouse/client/api/Client.java
chernser: I thought about it, but it is in different packages. 
We can have it in old client project because it is used everywhere. Will see what I can do.

client-v2/src/main/java/com/clickhouse/client/api/internal/HttpAPIClientHelper.java
chernser: yes.
we can restrict only changing client name, but not a header. Because we need to alter header value - we have this code. 
We can ignore the case when user completely overrides operation header but it is not how prev. client works. 
I've tried to  minimize work here.

Issue #1988, opened 2024-11-29
JWT support in JDBC driver


mshustov:
scope:
- ability to attach an JWT token
- ability to overwrite the token on the client instance

blocked by https://github.com/ClickHouse/clickhouse-java/issues/1834

chernser:
Passing JWT token in initial config is not a problem (may be thru URL will be problematic, but I hope it will not be a case). 
The problem is to update. Usually there is some token manager with registered callback or function that returns token for operation. 

JDBC is configured thru one of two ways:
1. url parameters 
2. properties

There is not problem to pass some instance of a class that returns tokens thru properties. Using URL we can pass only string. 
But we can limit solution only for properties.

mshustov:
closed by https://github.com/ClickHouse/clickhouse-java/pull/2055

PR #2055, opened 2024-12-27
Adding support to JWT in JDBC V2

mzitnik:
## Summary
Adding support to JWT in JDBC V2

## Checklist
- [x] Closes issue https://github.com/ClickHouse/clickhouse-java/issues/1988

### DIFF
--- a/jdbc-v2/src/main/java/com/clickhouse/jdbc/internal/DriverProperties.java
+++ b/jdbc-v2/src/main/java/com/clickhouse/jdbc/internal/DriverProperties.java
@@ -22,8 +22,7 @@ public enum DriverProperties {
      * query settings to be passed along with query operation.
      * {@see com.clickhouse.client.api.query.QuerySettings}
      */
-    DEFAULT_QUERY_SETTINGS("default_query_settings", null),
-    BEARER_TOKEN("jdbc_bearer_token", null);
+    DEFAULT_QUERY_SETTINGS("default_query_settings", null);
     private final String key;
 
     private final String defaultValue;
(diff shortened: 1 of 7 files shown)


jdbc-v2/src/main/java/com/clickhouse/jdbc/internal/JdbcConfiguration.java
```diff
@@ -240,7 +240,9 @@ public String getDriverProperty(String key, String defaultValue) {
     public Client.Builder applyClientProperties(Client.Builder builder) {
         builder.addEndpoint(connectionUrl)
                 .setOptions(clientProperties);
-
+        if (clientProperties.containsKey("access_token")) {
```
chernser, line 243: I think we may be need to move this code to the client:
1. create ClientConfigProperties,BEARERTOKEN_AUTH ("bearer_token")
2. handle it in Client#setOption

or create com.clickhouse.jdbc.internal.DriverProperties#BEARER_TOKEN("jdbc_bearer_token")
and use here instead of `access_token`;

Current `access_token` may conflict with existing com.clickhouse.client.api.ClientConfigProperties#ACCESS_TOKEN

jdbc-v2/src/main/java/com/clickhouse/jdbc/internal/DriverProperties.java
```diff
@@ -22,8 +22,8 @@ public enum DriverProperties {
      * query settings to be passed along with query operation.
      * {@see com.clickhouse.client.api.query.QuerySettings}
      */
-    DEFAULT_QUERY_SETTINGS("default_query_settings", null);
-
+    DEFAULT_QUERY_SETTINGS("default_query_settings", null),
+    BEARER_TOKEN("jdbc_bearer_token", null);
```
chernser, line 26: if there is a property in Client - we do not need this.
user will input "bearer_token" and it will be passed to the client.

chernser (approved):

PR #2056, opened 2024-12-28
[jdbc-v2] Fixed Tests

chernser:
## Checklist
- [ ] Closes issue 


mzitnik (approved):
Just for general knowledge why we decrease the limit number?
What are we testing here where is no assert in the test body

chernser:
Bigger number cause timeout on cloud. 
But test is verifying that cancel doesn't stuck or throw exception. 
we would need a better test for this.

PR #2062, opened 2025-01-01
Tableau fix getDate with calendar

mzitnik:
## Summary
Tableau fix getDate with calendar

## Checklist
- [ ] Closes issue 

### DIFF
--- a/client-v2/src/main/java/com/clickhouse/client/api/data_formats/internal/AbstractBinaryFormatReader.java
+++ b/client-v2/src/main/java/com/clickhouse/client/api/data_formats/internal/AbstractBinaryFormatReader.java
@@ -51,14 +51,9 @@ public abstract class AbstractBinaryFormatReader implements ClickHouseBinaryForm
     protected BinaryStreamReader binaryStreamReader;
 
     private TableSchema schema;
-
     private ClickHouseColumn[] columns;
-
     private Map[] convertions;
-
     private volatile boolean hasNext = true;
-
-    private DateTimeFormatter dateTimeFormatter = DateTimeFormatter.ofPattern("yyyy-MM-dd");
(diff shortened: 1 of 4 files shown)


client-v2/src/main/java/com/clickhouse/client/api/data_formats/internal/AbstractBinaryFormatReader.java
```diff
@@ -57,7 +58,7 @@ public abstract class AbstractBinaryFormatReader implements ClickHouseBinaryForm
 
     private volatile boolean hasNext = true;
 
-
+    private DateTimeFormatter dateTimeFormatter = DateTimeFormatter.ofPattern("yyyy-MM-dd");
```
chernser, line 61: we have this formatter com.clickhouse.client.api.DataTypeUtils#DATE_FORMATTER

jdbc-v2/src/main/java/com/clickhouse/jdbc/ResultSetImpl.java
```diff
@@ -41,6 +44,7 @@ public ResultSetImpl(StatementImpl parentStatement, QueryResponse response, Clic
         this.metaData = new com.clickhouse.jdbc.metadata.ResultSetMetaData(this);
         this.closed = false;
         this.wasNull = false;
+        this.defaultCalendar = new GregorianCalendar(TimeZone.getTimeZone("UTC"));
```
chernser, line 47: there is a feature to use server timezone for date. So it may be applicable here.

mzitnik: I will check the JDBC v1 implementation to how we deal with that in the past

Issue #2065, opened 2025-01-02
[jdbc] Incorrect DateTime results when writing numeric data with use_time_zone property


haishui126:
### Describe the bug

The DateTime results will -8h when inserting long values using PrepareStatement with use_server_time_zone=false&use_time_zone=Asia/Shanghai.

Expected: '1970-01-01 08:00:00.001000000'
Actual: '1970-01-01 00:00:00.001000000'
(timezone is Asia/Shanghai)

### Code example
```sql
CREATE TABLE t_test
(
    `timestamp` DateTime64(3)
)
ENGINE = Log;
```
```java
try (
    Connection connection = DriverManager.getConnection("jdbc:ch://localhost:8123/default?use_server_time_zone=false&use_time_zone=Asia/Shanghai", "default", "");
    PreparedStatement preparedStatement = connection.prepareStatement("INSERT INTO t_test VALUES (?)")
) {
    preparedStatement.setLong(1, 1);
    preparedStatement.execute();
}
```

#### Environment
* Client version: 0.7.2
* Language version: java11

haishui126:
Hi team,
I think the method in com.clickhouse.data.ClickHouseValues.java:282 is incorrect. Why should a numeric type representing a Unix timestamp, which inherently lacks timezone context, be directly converted to a LocalDateTime assuming it's in the UTC time zone?
```java
public static LocalDateTime convertToDateTime(BigDecimal value) {
    return value != null ? LocalDateTime.ofInstant(convertToInstant(value), ZoneOffset.UTC) : null;
}
```
The concern here is that the conversion assumes the input Unix timestamp is in UTC without verifying if this is correct for all cases, which can lead to time zone discrepancies when clients operate in different time zones.

dyjwi4-wiqvyr-cyZbem:
LhmRUXg7XomRqVMmVAz7Hu1k8ovSZ7v12G

chernser:
Good day, @haishui126 ! 
Sorry for a delayed answer and thank you for reporting the issue! 

We will look into it but this is a v1. 
Would you please try v2 (set by default after 0.8.0)? Thanks!

dyjwi4-wiqvyr-cyZbem:
Не чего не поняла телеграмм восстановите на номере или как ?почему номер
заблокирован пишет ?

Чт, 28 авг. 2025 г. в 23:21, Sergey Chernov ***@***.***>:

chernser:
Good day, @haishui126 ! 

[...]

PR #2066, opened 2025-01-03
[jdbc] Incorrect DateTime results when writing numeric data with use_time_zone property

haishui126:
## Summary

Refer to #2065.
When I use ClickHouse JDBC to insert data, the server's time zone is UTC, while my client's time zone is UTC+8. The JDBC first converts my numeric data to a LocalDateTime in the UTC time zone. This time object is then treated as a LocalDateTime in the UTC+8 time zone when written into ClickHouse, resulting in the written result being 8 hours behind.

## Checklist
- [x] Closes issue #2065 

### DIFF
--- a/clickhouse-data/src/main/java/com/clickhouse/data/value/ClickHouseDateTimeValue.java
+++ b/clickhouse-data/src/main/java/com/clickhouse/data/value/ClickHouseDateTimeValue.java
@@ -285,9 +285,9 @@ public class ClickHouseDateTimeValue extends ClickHouseObjectValue<LocalDateTime
         if (value == null) {
             resetToNullOrEmpty();
         } else if (scale == 0) {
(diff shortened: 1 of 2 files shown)


chernser:
This PR need to be ported/verified with jdbc-v2

haishui126:
@chernser This PR addresses the time zone issue when converting Long type to LocalDateTime type in the clickhouse-data module. The jdbc-v2 does not use the clickhouse-data module, so this change will not affect v2. However, I tested v2 in 0.8.0, and found that writing Long type to DateTime64 is fine, but when writing LocalDateTime type to DateTime64, it seems that the use_time_zone setting does not take effect.

chernser:
@haishui126 thank you for the info! 
I've created the issue https://github.com/ClickHouse/clickhouse-java/issues/2134

Issue #2112, opened 2025-01-25
getColumnClassName needs to be implemented


livk-cloud:
### Describe your feedback

**com.clickhouse.jdbc.metadata.ResultSetMetaData#getColumnClassName** needs to be implemented like **com.clickhouse.jdbc.ClickHouseResultSetMetaData**, otherwise Spring Jdbc and Mybatis will not work

```java
public class ResultSetMetaData implements java.sql.ResultSetMetaData, JdbcV2Wrapper {

       public String getColumnClassName(int column) throws SQLException {
        throw new SQLException("Not implemented", "0A000");
    }
}
```

```java
public class ClickHouseResultSetMetaData extends JdbcWrapper implements ResultSetMetaData {
    public String getColumnClassName(int column) throws SQLException {
        return this.mapper.toJavaClass(this.getColumn(column), this.typeMap).getCanonicalName();
    }
}
```

livk-cloud:
https://github.com/ClickHouse/clickhouse-java/pull/2113

chernser:
Good day, @livk-cloud ! 
We appreciate your contribution!

chernser:
@livk-cloud is there any document that points out how JDBC driver should behave and what should be returned? 
I mean any guidelines from Spring or Mybatis how they "see" type mapping? I think, it is critical now to understand to make a review of your PR. 
Thanks!

dspangen:
Is this intended to be implemented? The linked PR just throws a `null`.

We're trying to migrate an integration with Clickhouse (built on v1) and we do rely on `getColumnClassName(int index)`. Should we re-write or do you intend to actually implement this method?

chernser:
Good day, @dspangen ! 
We will implement this method in similar way as V1.

PR #2113, opened 2025-01-25
feat: implemented getColumnClassName

livk-cloud:
Implement the **getColumnClassName** method to accommodate Spring Jdbc and Mybatis

### DIFF
--- a/clickhouse-data/src/main/java/com/clickhouse/data/ClickHouseUtils.java
+++ b/clickhouse-data/src/main/java/com/clickhouse/data/ClickHouseUtils.java
@@ -362,7 +362,7 @@ public final class ClickHouseUtils {
                 pattern.chars().anyMatch(
                         value -> {
                             if (value < ' ' || reservedCharsWindows.indexOf(value) != -1) {
-                                throw new IllegalArgumentException("File path contains reserved character <%s>".formatted((char) value));
+                                throw new IllegalArgumentException(String.format("File path contains reserved character <%s>", value));
                             }
                             return false;
                         }
--- a/client-v2/src/main/java/com/clickhouse/client/api/Client.java
+++ b/client-v2/src/main/java/com/clickhouse/client/api/Client.java
@@ -2006,6 +2006,7 @@ public class Client implements AutoCloseable {
         return isAsync ? CompletableFuture.supplyAsync(resultSupplier, sharedOperationExecutor) : CompletableFuture.completedFuture(resultSupplier.get());
     }
 
+    @Override
     public String toString() {
         return "Client{" +
                 "endpoints=" + endpoints +
--- a/client-v2/src/main/java/com/clickhouse/client/api/data_formats/ClickHouseBinaryFormatReader.java
+++ b/client-v2/src/main/java/com/clickhouse/client/api/data_formats/ClickHouseBinaryFormatReader.java
@@ -11,11 +11,7 @@ import java.math.BigDecimal;
 import java.math.BigInteger;
 import java.net.Inet4Address;
 import java.net.Inet6Address;
-import java.time.Duration;
-import java.time.Instant;
-import java.time.LocalDate;
-import java.time.LocalDateTime;
-import java.time.ZonedDateTime;
+import java.time.*;
 import java.util.List;
 import java.util.Map;
 import java.util.UUID;
@@ -534,6 +530,10 @@ public interface ClickHouseBinaryFormatReader extends AutoCloseable {
     LocalDateTime getLocalDateTime(String colName);
 
     LocalDateTime getLocalDateTime(int index);
+
+    OffsetDateTime getOffsetDateTime(String colName);
+
+    OffsetDateTime getOffsetDateTime(int index);
 
     TableSchema getSchema();
 
--- a/client-v2/src/main/java/com/clickhouse/client/api/data_formats/RowBinaryFormatSerializer.java
+++ b/client-v2/src/main/java/com/clickhouse/client/api/data_formats/RowBinaryFormatSerializer.java
@@ -12,6 +12,8 @@ import java.math.BigDecimal;
 import java.math.BigInteger;
 import java.net.Inet4Address;
 import java.net.Inet6Address;
+import java.time.Instant;
+import java.time.OffsetDateTime;
 import java.time.ZoneId;
 import java.time.ZonedDateTime;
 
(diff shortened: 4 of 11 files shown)


livk-cloud:
https://github.com/ClickHouse/clickhouse-java/issues/2112

chernser:
@livk-cloud the build failed: 
```
Error:  Failed to execute goal org.apache.maven.plugins:maven-compiler-plugin:3.13.0:compile (default-compile) on project jdbc-v2: Compilation failure: Compilation failure: 
Error:  /home/runner/work/clickhouse-java/clickhouse-java/jdbc-v2/src/main/java/com/clickhouse/jdbc/metadata/ResultSetMetaData.java:[7,27] cannot find symbol
Error:    symbol:   class JdbcTypeMapping
Error:    location: package com.clickhouse.jdbc
Error:  /home/runner/work/clickhouse-java/clickhouse-java/jdbc-v2/src/main/java/com/clickhouse/jdbc/metadata/ResultSetMetaData.java:[169,9] cannot find symbol
Error:    symbol:   class JdbcTypeMapping
Error:    location: class com.clickhouse.jdbc.metadata.ResultSetMetaData
Error:  /home/runner/work/clickhouse-java/clickhouse-java/jdbc-v2/src/main/java/com/clickhouse/jdbc/metadata/ResultSetMetaData.java:[169,35] cannot find symbol
Error:    symbol:   variable JdbcTypeMapping
Error:    location: class com.clickhouse.jdbc.metadata.ResultSetMetaData
Error:  /home/runner/work/clickhouse-java/clickhouse-java/jdbc-v2/src/main/java/com/clickhouse/jdbc/metadata/ResultSetMetaData.java:[170,63] cannot find symbol
Error:    symbol:   variable Map
Error:    location: class com.clickhouse.jdbc.metadata.ResultSetMetaData
Error:  -> [Help 1]
Error:  
```

livk-cloud:
@chernser Hey, bro。I've tried to fix the above and remove the API introduced in JDK9，Thanks for the review

jdbc-v2/src/main/java/com/clickhouse/jdbc/JdbcTypeMapping.java
```diff
@@ -0,0 +1,655 @@
...
+                case Ring:
+                case Polygon:
+                case MultiPolygon:
+                    typeName = "ARRAY";
+                    break;
+                case Enum8:
+                case Enum16:
+                case IPv4:
+                case IPv6:
+                case JSON:
```
chernser, line 103: `JSON` - it depends on query settings if JSON should be treated as string. 
So it should be `OTHER`, I think.
`Object` - is deprecated. We should use `OTHER` .

jdbc-v2/src/main/java/com/clickhouse/jdbc/JdbcTypeMapping.java
```diff
@@ -0,0 +1,655 @@
...
+                    break;
+                case Point:
+                case Ring:
+                case Polygon:
+                case MultiPolygon:
+                    typeName = "ARRAY";
+                    break;
+                case Enum8:
+                case Enum16:
+                case IPv4:
```
chernser, line 101: Technically IP addresses are stored as series of numbers, but we can return them either way. If we put `VARCHAR` for it - what will be Spring behavior?

jdbc-v2/src/main/java/com/clickhouse/jdbc/JdbcTypeMapping.java
```diff
@@ -0,0 +1,655 @@
...
+                    typeName = "DOUBLE PRECISION";
+                    break;
+                case Point:
+                case Ring:
+                case Polygon:
+                case MultiPolygon:
+                    typeName = "ARRAY";
+                    break;
+                case Enum8:
+                case Enum16:
```
chernser, line 100: Enum s are `SMALLINT` as they are stored as int8, int16. 
What will be Spring behavior in this case?

jdbc-v2/src/main/java/com/clickhouse/jdbc/JdbcTypeMapping.java
```diff
@@ -0,0 +1,655 @@
...
+            return sqlType;
+        }
+
+        @Override
+        public String toNativeType(ClickHouseColumn column) {
+            return toAnsiSqlType(column, new StringBuilder());
+        }
+
+        @Override
+        public int toSqlType(ClickHouseColumn column, Map<String, Class<?>> typeMap) {
```
chernser, line 190: we have this mapping already com.clickhouse.jdbc.internal.JdbcUtils#generateTypeMap

jdbc-v2/src/main/java/com/clickhouse/jdbc/JdbcTypeMapping.java
```diff
@@ -0,0 +1,655 @@
...
+            }
+
+            return sqlType;
+        }
+    }
+
+    /**
+     * Inner class for static initialization.
+     */
+    static final class InstanceHolder {
```
chernser, line 289: Why do we need an instance holder? can we have just simple static block in JdbcTypeMapping class? 
We are trying to have classes with some meaningful functions that can be reused in different places. Current class is private and just holds static values.

jdbc-v2/src/main/java/com/clickhouse/jdbc/JdbcTypeMapping.java
```diff
@@ -0,0 +1,655 @@
...
+        return null;
+    }
+
+    /**
+     * Gets corresponding {@link ClickHouseDataType} of the given {@link Types}.
+     *
+     * @param sqlType generic SQL types defined in JDBC
+     * @return non-null ClickHouse data type
+     */
+    protected ClickHouseDataType getDataType(int sqlType) {
```
chernser, line 342: we have this mapping in com.clickhouse.jdbc.internal.JdbcUtils#SQL_TYPE_TO_CLASS_MAP

livk-cloud:
> There is good idea of customizable mappings. With different dialects and variety of tools should help. But I think, it needs more work to fit it current JDBC implementation.

> I'm coming up with a New Year's holiday in the coming days, and I probably don't have much time to code. Considering the multiple dialects, I think there should be a common interface that uses SPI for loading processing, what do you think about this?

chernser:
@livk-cloud I think we can do with simple properties. Besides not always it is possible to manipulate a service loader when only URL is available to setup. 
What mapping should be returned for your case?

livk-cloud:
> @livk-cloud I think we can do with simple properties. Besides not always it is possible to manipulate a service loader when only URL is available to setup. What mapping should be returned for your case?

> After some trying, I found that the **getColumnClassName** method returns Null and works fine with these frameworks, what do you think?

chernser:
Hi @livk-cloud ! 
I think we can just return null as a quick fix or just use `com.clickhouse.jdbc.internal.JdbcUtils#CLICKHOUSE_TO_SQL_TYPE_MAP`. 
Current PR seems more complex for this problem.

livk-cloud:
@chernser Of course, the easiest way to do this is to return Null. But I don't know if this has a bad effect on other ORM frameworks

PR #2124, opened 2025-01-30
[client-v2] Fixed cleaning expired connections from the pool

### LINKED ISSUE #2123: [client-v2] Number of active connections never drop after reaching max.
### Describe the bug

It was observed then when  number of active connections are riches its max it never gets down. This behavior makes metrics kind of useless because doesn't tell actual state of a connection pool and looks like a problem of stuck connections .

chernser:
## Summary
Expired connections are not removed from the pool. It causes two problem with metrics: they do not reflect actual state. 
This PR calls `poolControl.closeExpired()` before each request. It should not create a problem with pool. 

Closes https://github.com/ClickHouse/clickhouse-java/issues/2123
## Checklist
- [x] Closes #2123 


chernser:
Call to clear expired connection is not among hot spots:
image

Issue #1926, opened 2024-11-11
[client-v2] Variant Type Support 


chernser:
### Description

Variant type example definition: 
```
CREATE TABLE external_projects ( id Variant(UUID, UInt32, String), name String) Engine=MergeTree ORDER BY ()
```

Binary Encoded as: `[ord number of type from definition when types are sorted alphabetically (String = 0)][... encoded value of selected type]` 

This type is similar to Tuple, but only one value is present at a time. 
[Variants can be accessed by sub column names](https://clickhouse.com/docs/en/sql-reference/data-types/variant#reading-variant-nested-types-as-subcolumns). This should be implemented as ability of `ClickHouseBinaryReader`:

```
      // for query select id, name from external_projects 
      String id = reader.getString("id.String"); // should return string value or throw exception that value missing or null if nullable column
      String id = reader.getInteger("id.String"); // should throw exception about type mismatch even if column is nullable

```
There should be possibility to get variant type. This requires to add `getValueType` method. 

When encoding POJO: 
- Most suitable type is selected 
- Numbers are not converted to strings
- Strings not converted to DateTime values and vice versa 

[Variant Type Documentation](https://clickhouse.com/docs/en/sql-reference/data-types/variant)

mshustov:
@chernser closed by https://github.com/ClickHouse/clickhouse-java/pull/2077?

Issue #1833, opened 2024-09-25
[client-v2] JSON Support


chernser:
### Description
Client should support JSON Data type (https://clickhouse.com/docs/en/integrations/data-formats/json/overview). 

**Considerations:**
- Client should provide convenient API to get access to row data
- Client should have examples of using a few commonly used JSON parsing libraries
- Client should not implement own JSON parser in any way
- Client documentation should list all details of work with JSON format

----
See also:
- https://github.com/ClickHouse/ClickHouse/pull/70288
- https://github.com/Avogar/ClickHouse/blob/master/src/DataTypes/Serializations/SerializationObject.cpp

dzmitry-biatenia-smop:
@chernser any idea when 0.7.2 Java client will be released?

chernser:
Hi @dzmitry-biatenia-smop! 
We plan to release 0.7.2 by end of November. However this functionality is not in the 0.7.2: it is a complex and experimental feature - so it will take time to add it to the client. 
However  0.7.2 might have a minimal support when JSON columns can be read/written as a string. 

How would you use JSON? Do you have a DTO object or do you work directly with some generic classes generated by GSON or Jackson ? 

Thanks!

dzmitry-biatenia-smop:
@chernser the basic case - I was able to create a column with JSON type in Clickhouse table with flag _allow_experimental_json_type_ enabled, and even was able to insert some JSON data there with com.clickhouse.client.api.Client#execute(), but when can running any com.clickhouse.client.api.Client.query("select * from table_with_json_column") I'm getting **Unsupported format: "JSON"** coming from [here](https://github.com/ClickHouse/clickhouse-java/blob/eb0bfaf20a7ed0fd37d8e31e368df486757f1ea2/client-v2/src/main/java/com/clickhouse/client/api/Client.java#L1906) because response has JSON type here.
I tried to set _output_format_binary_write_json_as_string_ , but seems it's not supported by my version of server 24.8.4

chernser:
@dzmitry-biatenia-smop 
I will fix this issue for sure.

dzmitry-biatenia-smop:
@chernser are you going to release that fix separately? any ETA for that fix?

chernser:
@dzmitry-biatenia-smop 
patch should be release by end of this week. 
release is by end of the month
another option is to take nightly build.

dzmitry-biatenia-smop:
[...]

PR #2130, opened 2025-02-04
[client-v2] Support Dynamic and JSON Data Types

### LINKED ISSUE #1925: [client-v2] Dynamic Data Type Support
### Description

[Data Type documentation](https://clickhouse.com/docs/en/sql-reference/data-types/dynamic)

chernser:
## Summary

This PR adds support for Variant, Dynamic and JSON for reading and writing. Here is details what is supported 
All this is try only for `RowBinary` format. `Native` will need another implementation because it has another structure. 

| DataType |    POJO |  Binary Reader | JDBC |
|----------|:-------------:|------:|------:|
| Variant|  Read/Write | Yes | Yes |
| Dynamic |  Read/Write | Yes | Yes |
| JSON | No | Yes | Yes/No | 

### JSON Support 
JSON in RowBinary is encoded as map of path-value (for example "a.b" = 1). Reader returns a flat map for this. 
It is convenient when  path is known and hardcoded. But inconvenient for parsing. It should be possible to get Map-of-Map structure. However this is most heavy-weight solution. 
Writing JSON from POJO is not done in this PR. It can be done in two ways:
- in-house solution - tedious work with reflection
- plugin GSON/Jackson libraries to make Map-of-Map structure so can be easily transformed into a flat map.

PS: There are some test fixes. 

Closes https://github.com/ClickHouse/clickhouse-java/issues/1925
Closes https://github.com/ClickHouse/clickhouse-java/issues/1833
Closes https://github.com/ClickHouse/clickhouse-java/issues/1926

## Checklist
- [x] Closes #1925, #1833,#1926


clickhouse-data/src/main/java/com/clickhouse/data/ClickHouseDataType.java
```diff
@@ -49,74 +49,75 @@
...
+    IntervalSecond(Long.class, false, true, true, 8, 19, 0, 0, 0, false, 0x22),
+    IntervalMicrosecond(Long.class, false, true, true, 8, 19, 0, 0, 0, false, 0x22),
+    IntervalMillisecond(Long.class, false, true, true, 8, 19, 0, 0, 0, false, 0x22),
+    IntervalNanosecond(Long.class, false, true, true, 8, 19, 0, 0, 0, false, 0x22),
+    UInt64(UnsignedLong.class, false, true, false, 8, 20, 0, 0, 0, false, 0x04, "BIGINT UNSIGNED", "BIT", "SET"),
+    Int128(BigInteger.class, false, true, true, 16, 39, 0, 0, 0, false, 0x0B),
+    UInt128(BigInteger.class, false, true, false, 16, 39, 0, 0, 0, false, 0x05),
+    Int256(BigInteger.class, false, true, true, 32, 77, 0, 0, 0, false, 0x0C),
+    UInt256(BigInteger.class, false, true, false, 32, 78, 0, 0, 0, false, 0x06),
     Decimal(BigDecimal.class, true, false, true, 0, 76, 0, 0, 76, false, "DEC", "FIXED", "NUMERIC"),
```
chernser, line 88: Virtual type - no bin tag

clickhouse-data/src/main/java/com/clickhouse/data/ClickHouseDataType.java
```diff
@@ -212,13 +213,81 @@ static Map<ClickHouseDataType, Set<Class<?>>> dataTypeClassMap() {
...
+        map.put(IntervalNanosecond, setOfAllNumberClasses);
+
         return map;
     }
 
     private static Set<Class<?>> setOf(Class<?>... args) {
         return Collections.unmodifiableSet(new HashSet<>(Arrays.stream(args).collect(Collectors.toList())));
     }
 
+    public static final byte INTERVAL_BIN_TAG = 0x22;
```
chernser, line 236: following tags are not actual types or require special processing.

clickhouse-data/src/main/java/com/clickhouse/data/ClickHouseDataType.java
```diff
@@ -258,6 +333,21 @@ private static Set<Class<?>> setOf(Class<?>... args) {
 
         allAliases = Collections.unmodifiableSet(set);
         name2type = Collections.unmodifiableMap(map);
+
```
chernser, line 336: Creating lookup maps for types and interval kinds

clickhouse-data/src/main/java/com/clickhouse/data/ClickHouseEnum.java
```diff
@@ -51,7 +51,7 @@ protected ClickHouseEnum(Collection<String> params) {
         }
     }
 
-    protected ClickHouseEnum(String[] names, int[] values) {
+    public ClickHouseEnum(String[] names, int[] values) {
```
chernser, line 54: no need to protect the constructor of such simple class.

client-v2/src/main/java/com/clickhouse/client/api/data_formats/internal/BinaryStreamReader.java
```diff
@@ -641,6 +658,38 @@ public synchronized <T> List<T> asList() {
         }
     }
 
+    public static class EnumValue extends Number {
```
chernser, line 661: Have to create the class to overcome problem:
- when dynamic column then each enum value may have different set of columns
- when enum column is requested as String we need string constant but normally it is stored in column definition but it will be dynamic.

client-v2/src/main/java/com/clickhouse/client/api/data_formats/internal/BinaryStreamReader.java
```diff
@@ -956,4 +1005,81 @@ public byte[] allocate(int size) {
             return new byte[size];
         }
     }
+
+    private static final Set<Byte> DECIMAL_TAGS = Collections.unmodifiableSet(new HashSet<>(Arrays.asList(
```
chernser, line 1009: decimal has special handling because of arguments. This set just simplifies if statements.

client-v2/src/main/java/com/clickhouse/client/api/data_formats/internal/BinaryStreamReader.java
```diff
@@ -956,4 +1005,81 @@ public byte[] allocate(int size) {
...
+
+    private static final Set<Byte> DECIMAL_TAGS = Collections.unmodifiableSet(new HashSet<>(Arrays.asList(
+            ClickHouseDataType.Decimal.getBinTag(),
+            ClickHouseDataType.Decimal32.getBinTag(),
+            ClickHouseDataType.Decimal64.getBinTag(),
+            ClickHouseDataType.Decimal128.getBinTag(),
+            ClickHouseDataType.Decimal256.getBinTag()
+    )));
+
+    private ClickHouseColumn readDynamicData() throws IOException {
```
chernser, line 1017: reads metadata for dynamic value

client-v2/src/main/java/com/clickhouse/client/api/data_formats/internal/SerializerUtils.java
```diff
@@ -87,35 +90,326 @@ public static void serializeData(OutputStream stream, Object value, ClickHouseCo
...
                 serializePrimitiveData(stream, value, column);
                 break;
 
         }
     }
 
-    private static void serializeArrayData(OutputStream stream, Object value, ClickHouseColumn column) throws IOException {
+    private static final Map<Class<?>, ClickHouseColumn> PREDEFINED_TYPE_COLUMNS = getPredefinedTypeColumnsMap();
+
+    private static Map<Class<?>, ClickHouseColumn> getPredefinedTypeColumnsMap() {
```
chernser, line 107: We use Column object to properly serialize/deserialize values, so I've created predefined list to avoid creation all the time. 
But it can be more lightweight - will redesign in the future.

client-v2/src/main/java/com/clickhouse/client/api/data_formats/internal/SerializerUtils.java
```diff
@@ -87,35 +90,326 @@ public static void serializeData(OutputStream stream, Object value, ClickHouseCo
...
-            //Serialize the array to the stream
-            //The array is a list of values
-            List<?> values = (List<?>) value;
-            writeVarInt(stream, values.size());
-            for (Object val : values) {
-                if (column.getArrayBaseColumn().isNullable()) {
-                    if (val == null) {
-                        writeNull(stream);
-                        continue;
+    public static ClickHouseColumn valueToColumnForDynamicType(Object value) {
```
chernser, line 161: we need to crate column definition for each value so serializer can use it like a normal column serialization.

client-v2/src/test/java/com/clickhouse/client/insert/InsertTests.java
```diff
@@ -638,8 +640,57 @@ public static Object[][] testAppCompressionDataProvider() {
         };
     }
 
+    @Test(groups = { "integration" }, enabled = true)
+    public void testPOJOWithDynamicType() throws Exception {
+        if (isVersionMatch("(,24.8]")) {
```
mshustov, line 645: should it be a proper comparison with a semver version ( `greater or equal  24.8`)? as done below https://github.com/ClickHouse/clickhouse-java/pull/2130/files#diff-89c11757b9243105d9931d9b667e953b7beabe6ca45ab2f333e905bb7003f8a9R944

client-v2/src/test/java/com/clickhouse/client/datatypes/DataTypeTests.java
```diff
@@ -388,6 +370,207 @@ public void testVariantWithTuple() throws Exception {
...
+    }
+
+    @Data
+    @AllArgsConstructor
+    public static class DTOForDynamicPrimitivesTests {
+        private int rowId;
+        private Object field;
+    }
+
+    private void testDynamicWith(String withWhat, Object[] values, String[] expectedStrValues) throws Exception {
```
mshustov, line 550: do we need separate json type tests?

mshustov (commented):
> Writing JSON from POJO is not done in this PR. It can be done in two ways:

+1 to use Gson or jackson 

> JSON + JDBC

What is our plan for JSON type usage via JDBC?

client-v2/src/test/java/com/clickhouse/client/insert/InsertTests.java
chernser: This expression does the same - it is from old tests and I just reuse it. We will unify such things later when we have one client and can have a utility method.

client-v2/src/test/java/com/clickhouse/client/datatypes/DataTypeTests.java
chernser: I think no - JSON uses the same code as Dynamic for values. Paths are just string additions to each value. 
The rest of logic is tested in InsertTests/QueryTests thru POJO.

client-v2/src/main/java/com/clickhouse/client/api/data_formats/internal/BinaryStreamReader.java
```diff
@@ -130,41 +143,41 @@ public <T> T readValue(ClickHouseColumn column, Class<?> typeHint) throws IOExce
...
                 case Bool:
                     return (T) Boolean.valueOf(readByteOrEOF(input) == 1);
-                case Enum8:
-                    return (T) Byte.valueOf((byte) readUnsignedByte());
-                case Enum16:
-                    return (T) Short.valueOf((short) readUnsignedShortLE());
+                case Enum8: {
+                    byte enum8Val = (byte) readUnsignedByte();
+                    String name = actualColumn.getEnumConstants().nameNullable(enum8Val);
+                    return (T) new EnumValue(name == null ? "<unknown>" : name, enum8Val);
```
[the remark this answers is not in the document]
chernser: `""` will be a blank space and it will be hard to detect error. `<unknown>`  will attract attention.

clickhouse-data/src/main/java/com/clickhouse/data/ClickHouseEnum.java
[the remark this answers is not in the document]
chernser: Current approach is more memory efficient. We can introduce map and measure performance.

client-v2/src/test/java/com/clickhouse/client/datatypes/DataTypeTests.java
mshustov: > JSON uses the same code as Dynamic for values

are we sure this assumption will always be valid?

mzitnik (approved):

PR #2131, opened 2025-02-06
[DO NOT MERGE] connection metrics

chernser:
## Summary
This PR is just showcase for what can be done to collect certain metrics 

Closes 
## Checklist
- [ ] Closes #<issue ref>

### DIFF
--- a/client-v2/src/main/java/com/clickhouse/client/api/internal/HttpAPIClientHelper.java
+++ b/client-v2/src/main/java/com/clickhouse/client/api/internal/HttpAPIClientHelper.java
@@ -13,27 +13,38 @@ import com.clickhouse.client.api.ServerException;
 import com.clickhouse.client.api.data_formats.internal.SerializerUtils;
 import com.clickhouse.client.api.enums.ProxyType;
 import com.clickhouse.client.api.http.ClickHouseHttpProto;
+import org.apache.hc.client5.http.AuthenticationStrategy;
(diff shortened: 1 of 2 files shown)


client-v2/src/main/java/com/clickhouse/client/api/internal/HttpAPIClientHelper.java
```diff
@@ -220,12 +234,23 @@ private HttpClientConnectionManager poolConnectionManager(LayeredConnectionSocke
                         .setBufferSize(networkBufferSize)
                         .build(),
                 CharCodingConfig.DEFAULT,
-                DefaultHttpResponseParserFactory.INSTANCE);
+                DefaultHttpResponseParserFactory.INSTANCE) {
```
chernser, line 237: We would need normal class somewhere at the bottom of this helper class, just to keep code clear. 
There is open question how to report this to micrometer. 
Current integration is do for pool by calling getters when metrics snapshot is collected - so our factory should have similar getter.

client-v2/src/test/java/com/clickhouse/client/query/QueryTests.java
```diff
@@ -2056,4 +2056,14 @@ public void testGettingRowsBeforeLimit() throws Exception {
             Assert.assertEquals(response.getTotalRowsToRead(), expectedTotalRowsToRead);
         }
     }
+
+    static {
```
chernser, line 2060: please delete debug from  this test.

Issue #2122, opened 2025-01-28
[client-v2] Error when inserting DTOs with null value representing Nullable(Int) field of Tuple or Array


medovgeo:
### Describe the bug
Exception occurs when inserting DTOs with `null` elements(of List) that represents a `Nullable(numeric type)` of `Tuple` or `Array`. 
I checked it only on some numeric types and strings and for strings it doesn't reproduce (I believe this is merit of java standard library).  
Seems that java client doesn't expect nullable values when parses elements of Tuples or Arrays. 

reproducible example with java code and table creation statement below as per template

### Steps to reproduce
1. Create table with `Tuple(Nullable(Int))` field -- _reproduces at least with all numeric primitives_
2. Insert data using DTOs - `insert(String tableName, List<?> data, InsertSettings settings)`

### Expected behaviour
Successful insertion

### Code example
```java
import com.clickhouse.client.api.Client;
import com.clickhouse.client.api.enums.Protocol;

import java.util.ArrayList;
import java.util.List;

public class Reproducer {
    public static void main(String[] args) throws Exception {
        var client = new Client.Builder()
                    .addEndpoint(Protocol.HTTP, "localhost", 8123, false)
                    .setUsername("user")
                    .setPassword("password")
                    .build();

            String tableName= "reproducer";
            var sample = new ArrayList<Object>();
            sample.add(null);
            client.register(Message.class, client.getTableSchema(tableName));
            client.insert(tableName, List.of(new Message("id", sample)));
    }

    public static class Message {
        public String id;
        public List<Object> sample;

        public Message(String id, List<Object> sample) {
            this.id = id;
            this.sample = sample;
        }

        public String getId() {
            return id;
        }
        public List<Object> getSample() {
            return sample;
        }
    }
}
```
### Error log
```java
Exception in thread "main" com.clickhouse.client.api.ClientException: Failed to execute request
	at com.clickhouse.client.api.internal.HttpAPIClientHelper.executeRequest(HttpAPIClientHelper.java:423)
	at com.clickhouse.client.api.Client.lambda$insert$6(Client.java:1336)
	at com.clickhouse.client.api.Client.runAsyncOperation(Client.java:1994)
	at com.clickhouse.client.api.Client.insert(Client.java:1386)
	at com.clickhouse.client.api.Client.insert(Client.java:1265)
	at Reproducer.main(Reproducer.java:20)
Caused by: java.lang.IllegalArgumentException: Cannot convert null to Integer
[...]

PR #2149, opened 2025-02-14
Send null when nullable

Paultagoras:
## Summary
* Send null when we have null primitives and a nullable column

Closes https://github.com/ClickHouse/clickhouse-java/issues/2122
## Checklist
- [x] Closes #<issue ref>

### DIFF
--- a/client-v2/src/main/java/com/clickhouse/client/api/data_formats/internal/SerializerUtils.java
+++ b/client-v2/src/main/java/com/clickhouse/client/api/data_formats/internal/SerializerUtils.java
@@ -462,6 +462,12 @@ public class SerializerUtils {
     }
 
     private static void serializePrimitiveData(OutputStream stream, Object value, ClickHouseColumn column) throws IOException {
+        //Handle null values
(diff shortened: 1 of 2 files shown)


chernser:
@Paultagoras would you please add a test for JDBC to check that we are getting null? I remember it has separate class or so. 
Also no `Array(Nullable(Int32))` test case.

client-v2/src/main/java/com/clickhouse/client/api/data_formats/internal/SerializerUtils.java
```diff
@@ -462,6 +462,12 @@ private static void serializeMapData(OutputStream stream, Object value, ClickHou
     }
 
     private static void serializePrimitiveData(OutputStream stream, Object value, ClickHouseColumn column) throws IOException {
+        //Handle null values
+        if (value == null && column.isNullable()) {//Only nullable columns can have null values
```
mzitnik, line 466: If it null and not nullable should we throw an exception like data is not valid

mzitnik (approved):

PR #2164, opened 2025-02-20
[client-v1] Removed unnecessary object creation. Added executor pool configuration.

chernser:
## Summary
This PR improves v1 performance in different ways: 
- no unnecessary creation of column index in each value
- bounded queue for executor and non-zero TTL for core threads to avoid spawning new thread for each request
- network buffer configuration to let allocate bigger buffer for copying data from socket

The first one improvement is significant because:
- reduces number of objects - so less GC work 
- reduces memory overhead
- reduces CPU time 

Closes 
## Checklist
- [ ] Closes #<issue ref>


clickhouse-data/src/main/java/com/clickhouse/data/ClickHouseSimpleRecord.java
```diff
@@ -86,25 +79,11 @@ public ClickHouseValue getValue(int index) {
...
 
     @Override
     public int size() {
         return values.length;
     }
-
-    private int computeColumnIndex(String name) {
-        int index = 0;
-        for (ClickHouseColumn c : columns) {
-            if (c.getColumnName().equalsIgnoreCase(name)) {
```
chernser, line 103: I think, we need to drop this behavior. 
Client-v2 uses exact match.

mshustov (approved):

PR #2195, opened 2025-03-04
Refactoring and adding DataSet

Paultagoras:
## Summary
* Adjusting the tests to be in a singular place, and have a more extensible footprint.

### DIFF
--- a/clickhouse-jdbc/pom.xml
+++ b/clickhouse-jdbc/pom.xml
@@ -137,6 +137,19 @@
             <version>${revision}</version>
             <scope>compile</scope>
         </dependency>
+        <dependency>
+            <groupId>org.openjdk.jmh</groupId>
+            <artifactId>jmh-core</artifactId>
+            <version>${jmh.version}</version>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>org.openjdk.jmh</groupId>
+            <artifactId>jmh-generator-annprocess</artifactId>
+            <version>${jmh.version}</version>
+            <scope>test</scope>
+        </dependency>
+
     </dependencies>
 
     <profiles>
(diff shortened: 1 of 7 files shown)


performance/src/test/com/clickhouse/benchmark/clients/ClientV2.java
```diff
@@ -0,0 +1,103 @@
...
+        ClickHouseNode node = getServer();
+        client = new Client.Builder()
+                .addEndpoint(Protocol.HTTP, node.getHost(), node.getPort(), isCloud())
+                .setUsername(getUsername())
+                .setPassword(getPassword())
+                .compressClientRequest(true)
+                .setMaxRetries(0)
+                .setLZ4UncompressedBufferSize(2 * 1024 * 1024)
+                .setClientNetworkBufferSize(20 * 1024 * 1024)
+                .setSocketRcvbuf(2 * 1024 * 1024)
```
mshustov, line 48: it's not a default experience. How many users know what parameters should be adjusted and how?

performance/src/test/com/clickhouse/benchmark/clients/BenchmarkBase.java
```diff
@@ -0,0 +1,104 @@
...
+        ClickHouseNode node = getServer();
+        try (Client client = new Client.Builder()
+                .addEndpoint(Protocol.HTTP, node.getHost(), node.getPort(), isCloud())
+                .setUsername(getUsername())
+                .setPassword(getPassword())
+                .compressClientRequest(true)
+                .setDefaultDatabase(DB_NAME)
+                .build();
+             InsertResponse response = client.insert(tableName, dataStream, format).get()) {
+            LOGGER.info("Rows inserted: {}", response.getWrittenRows());
```
mshustov, line 98: we need to validate the data is written correctly. at least a simple sanity check for a single row. can be done in a follow-up

chernser (approved):

PR #2212, opened 2025-03-12
corrected v1 tests and config to match v2

chernser:
## Summary

Closes 
## Checklist
- [ ] Closes #<issue ref>

### DIFF
--- a/performance/src/test/com/clickhouse/benchmark/BenchmarkRunner.java
+++ b/performance/src/test/com/clickhouse/benchmark/BenchmarkRunner.java
@@ -45,6 +45,7 @@ public class BenchmarkRunner {
                 .jvmArgs("-Xms8g", "-Xmx8g")
                 .measurementTime(TimeValue.seconds(isCloud() ? 30 : 10))
                 .resultFormat(ResultFormatType.JSON)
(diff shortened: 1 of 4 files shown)


laeg:
should we flip it the other way around, that v2 defaults match v1 defaults as most folks will be running stock v1?

mshustov:
> should we flip it the other way around, that v2 defaults match v1 defaults as most folks will be running stock v1?

Not necessarily. I think compression should be enabled by default in most cases: in the real-world case, the client and server are located on different physical machines with a network hop between.

> ideally, we have 1 test with v2 defaults and another that has v2 matching v1 defaults so we can compare

+1

laeg:
> > should we flip it the other way around, that v2 defaults match v1 defaults as most folks will be running stock v1?
> 
> Not necessarily. I think compression should be enabled by default in most cases: in the real-world case, the client and server are located on different physical machines with a network hop between, so in the most of the cases having compression is more beneficial
> 
Sorry, maybe I've misunderstood. I thought we were turning off compression in v1 with this PR? I agree that compression should be on by default, hence why i suggested that we get v2 (uncompressed) to match v1 (compressed).

PR #2216, opened 2025-03-13
Fix typo in param

lucapette:
Sorry I know it's a super trivial change but it caught my eyes while reading the source code and fixing it took less time than writing this description


chernser:
@lucapette - any contribution is very welcome!

chernser (approved):

PR #2225, opened 2025-03-14
[perf] fixed LZ4OutputStream

chernser:
## Summary

Closes 
## Checklist
- [ ] Closes #<issue ref>


client-v2/src/main/java/com/clickhouse/client/api/internal/ClickHouseLZ4OutputStream.java
```diff
@@ -26,16 +26,27 @@ public class ClickHouseLZ4OutputStream extends OutputStream {
...
-        write(tmpBuffer, 0, 1);
+        if (inBuffer.remaining() == 0) {
+            flush();
+        }
+        inBuffer.put((byte) b);
+    }
+
+    @Override
+    public void write(byte[] b) throws IOException {
+        if (b.length == 1) {
```
mzitnik, line 45: In the other method, you are doing some checks like `if (b == null)` should we also do it here

chernser: I honestly do not see any use of checking for null an throwing own NPE exception.

PR #2270, opened 2025-04-01
[perf] Jmh app

chernser:
## Summary
Makes a benchmark project as standalone application that can be run with options. 
Adds JMH workflow to run minimal benchmarks 

## Checklist
- [ ] Closes #<issue ref>


.github/workflows/benchmarks.yml
```diff
@@ -0,0 +1,74 @@
...
+on:
+  schedule:
+    # Build and publish nightly snapshot at 3:55pm every day
+    - cron: "55 15 * * *"
+  workflow_dispatch:
+    inputs:
+      pr:
+        description: "Pull request#"
+        required: false
+  pull_request:
```
chernser, line 13: This should be removed - here just for testing.

chernser: Actually we may run this if PR marked with performance label.

mzitnik (changes requested):

mzitnik (approved):

Issue #2206, opened 2025-03-10
Null SSLSocket causes NPE when jdbc url contains "ssl=true"


kslex:
### Describe the bug
The DummySSLConnectionSocketFactory creates a null SSLSocket because initSslContext in HttpAPIClientHelper#createHttpClient is false. 
The root cause is that JdbcConfiguration does not read the ssl property from jdbc url.

### Steps to reproduce
1. set dependency version between 0.8.0 to 0.8.2
2. set jdbc url like: jdbc:clickhouse://localhost:8123/testdb?ssl=true

### Expected behaviour
SSLSocket not null

### Code example
```java
url: jdbc:clickhouse://localhost:8123/testdb?ssl=true
username: xxx
password: xxx
driver-class-name: com.clickhouse.jdbc.ClickHouseDriver

<dependency>
    <groupId>com.clickhouse</groupId>
    <artifactId>clickhouse-jdbc</artifactId>
    <version>0.8.0</version>
</dependency>
```
### Error log
Caused by: java.lang.NullPointerException: SSL Socket
java.base/java.util.Objects.requireNonNull(Objects.java:259)
 org.apache.hc.core5.util.Args.notNull(Args.java:165)
 org.apache.hc.core5.http.impl.io.SocketHolder.<init>(SocketHolder.java:58)
 org.apache.hc.client5.http.impl.io.DefaultManagedHttpClientConnection.bind(DefaultManagedHttpClientConnection.java:188)
 org.apache.hc.client5.http.impl.io.DefaultHttpClientConnectionOperator.connect(DefaultHttpClientConnectionOperator.java:232)
 org.apache.hc.client5.http.impl.io.PoolingHttpClientConnectionManager.connect(PoolingHttpClientConnectionManager.java:490)
 org.apache.hc.client5.http.impl.classic.InternalExecRuntime.connectEndpoint(InternalExecRuntime.java:164)
 org.apache.hc.client5.http.impl.classic.InternalExecRuntime.connectEndpoint(InternalExecRuntime.java:174)
 org.apache.hc.client5.http.impl.classic.ConnectExec.execute(ConnectExec.java:144)
 org.apache.hc.client5.http.impl.classic.ExecChainElement.execute(ExecChainElement.java:51)
 org.apache.hc.client5.http.impl.classic.ProtocolExec.execute(ProtocolExec.java:192)
 org.apache.hc.client5.http.impl.classic.ExecChainElement.execute(ExecChainElement.java:51)
[...]

PR #2271, opened 2025-04-02
fix null SSLSocket (#2206)

kslex:
## Summary
- read ssl property from clientProperties that have already been initialized
- The clientProperties already contain bearerToken

Closes [#2206 ](https://github.com/ClickHouse/clickhouse-java/issues/2206)
## Checklist
- [ ] Closes [#2206 ](https://github.com/ClickHouse/clickhouse-java/issues/2206)

### DIFF
--- a/jdbc-v2/src/main/java/com/clickhouse/jdbc/internal/JdbcConfiguration.java
+++ b/jdbc-v2/src/main/java/com/clickhouse/jdbc/internal/JdbcConfiguration.java
@@ -57,11 +57,7 @@ public class JdbcConfiguration {
         initProperties(urlProperties, info);
 
         // after initializing all properties - set final connection URL
-        boolean useSSL = Boolean.parseBoolean(info.getProperty(DriverProperties.SECURE_CONNECTION.getKey(), "false"));
-        String bearerToken = info.getProperty(ClientConfigProperties.BEARERTOKEN_AUTH.getKey(), null);
-        if (bearerToken != null) {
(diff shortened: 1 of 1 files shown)


jdbc-v2/src/main/java/com/clickhouse/jdbc/internal/JdbcConfiguration.java
```diff
@@ -57,11 +57,7 @@ public JdbcConfiguration(String url, Properties info) throws SQLException {
         initProperties(urlProperties, info);
 
         // after initializing all properties - set final connection URL
-        boolean useSSL = Boolean.parseBoolean(info.getProperty(DriverProperties.SECURE_CONNECTION.getKey(), "false"));
-        String bearerToken = info.getProperty(ClientConfigProperties.BEARERTOKEN_AUTH.getKey(), null);
```
chernser, line 61: Looks like code related to bearerToken should be here.

chernser:
@kslex would you please add at least one test and return code about bearer token back? 
Thanks you for the fix!

jdbc-v2/src/main/java/com/clickhouse/jdbc/internal/JdbcConfiguration.java
kslex: the clientProperties already contain bearerToken, so this code is not necessary

mzitnik:
We will provide the fix for that issue for the next release.

mshustov:
fixed by https://github.com/ClickHouse/clickhouse-java/pull/2304

PR #2277, opened 2025-04-03
Clean empty lines in sql query

### LINKED ISSUE #2217: Newlines after the SQL comment break `parseStatementType`
Taken from https://github.com/ClickHouse/metabase-clickhouse-driver/issues/297#issuecomment-2690817946

DDLs: 
```sql
CREATE TABLE default.balance ( `id` UUID, `currency` String, `amount` Decimal(64, 18), `create_time` DateTime64(6), `_version` UInt64, `_sign` UInt8 ) ENGINE = ReplacingMergeTree PRIMARY KEY id ORDER BY id;

INSERT INTO balance VALUES (generateUUIDv4(), 'EUR', '42.42', now(), 144, 255);
```

both of these queries

```sql
-- SELECT * FROM balance

WITH balance_cte AS (
SELECT
id, currency, amount
FROM balance
LIMIT 10
)
SELECT * FROM balance_cte;
```

and

```sql
-- SELECT amount FROM default.balance FINAL;

SELECT amount FROM default.balance FINAL;
```

fail in Metabase with

```
Select statement did not produce a ResultSet for native query
```

However, if I remove an extra newline from it, it will work fine, e.g.

```sql
-- SELECT amount FROM default.balance FINAL;
SELECT amount FROM default.balance FINAL;
```

It returns `42.42` as expected; the query with CTE also works.

Most likely, newlines after the comment break `parseStatementType`, so it is not recognized as `SELECT`; then, when `getResultSet` is called on the `Statement`, `checkClosed` throws, as the statement is kept open only for specific statements types.

mzitnik:
## Summary

Closes #2217 
## Checklist
- [ ] Closes #2217 

### DIFF
--- a/jdbc-v2/src/test/java/com/clickhouse/jdbc/StatementTest.java
+++ b/jdbc-v2/src/test/java/com/clickhouse/jdbc/StatementTest.java
@@ -614,6 +614,27 @@ public class StatementTest extends JdbcIntegrationTest {
                 ResultSet rs = stmt.executeQuery(sqlSelect);
                 assertTrue(rs.next());
             }
+            try (Statement stmt = conn.createStatement()) {
+                String sqlSelect = new StringBuilder("-- SELECT amount FROM balance FINAL;\n")
+                        .append("\n")
+                        .append("SELECT amount /* test */FROM balance FINAL;").toString();
+                ResultSet rs = stmt.executeQuery(sqlSelect);
+                assertTrue(rs.next());
+            }
+            try (Statement stmt = conn.createStatement()) {
+                String sqlSelect = new StringBuilder("-- SELECT amount FROM balance FINAL;\n")
+                        .append("\n")
+                        .append("SELECT amount FROM balance FINAL; /* test */").toString();
+                ResultSet rs = stmt.executeQuery(sqlSelect);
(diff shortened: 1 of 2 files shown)


jdbc-v2/src/test/java/com/clickhouse/jdbc/StatementTest.java
```diff
@@ -573,4 +573,47 @@ public void testSwitchDatabase() throws Exception {
...
+                int r = stmt.executeUpdate(sqlCreate);
+                assertEquals(r, 0);
+            }
+            try (Statement stmt = conn.createStatement()) {
+                String sqlInsert = "INSERT INTO balance VALUES (generateUUIDv4(), 'EUR', '42.42', now(), 144, 255);";
+                int r = stmt.executeUpdate(sqlInsert);
+                assertEquals(r, 1);
+            }
+            try (Statement stmt = conn.createStatement()) {
+                String sqlSelect = new StringBuilder("-- SELECT amount FROM balance FINAL;\n")
```
mshustov, line 591: can we add tests with a comment
- in the middle of a query
- at the end of the query

mshustov (approved):

Issue #2284, opened 2025-04-04
[client-v2] Support an arbitrary column list for insert(...)


Paultagoras:
Right now we hard code the insert statement without including columns:

`final String sqlStmt = "INSERT INTO " + tableName + " FORMAT " + format.name();`

We should allow someone to specify a List<String> of columnNames to include:

`final String sqlStmt = "INSERT INTO " + tableName + " (col1, col2, etc...) FORMAT " + format.name();`

mshustov:
how it's done in clickhouse-js https://github.com/ClickHouse/clickhouse-js/blob/main/examples/insert_exclude_columns.ts

PR #2289, opened 2025-04-07
Add support for partial insert

Paultagoras:
## Summary
* Adds a List<String> to allow partial insert with a subset of columns

Closes https://github.com/ClickHouse/clickhouse-java/issues/2284
## Checklist
- [x] Closes #<issue ref>

### DIFF
--- a/client-v2/src/main/java/com/clickhouse/client/api/Client.java
+++ b/client-v2/src/main/java/com/clickhouse/client/api/Client.java
@@ -1521,14 +1521,14 @@ public class Client implements AutoCloseable {
      * Does an insert request to a server. Data is pushed when a {@link DataStreamWriter#onOutput(OutputStream)} is called.
      *
      * @param tableName - target table name
-     * @param columns - list of column names to insert data into. If null or empty, all columns will be used.
+     * @param columnNames - list of column names to insert data into. If null or empty, all columns will be used.
      * @param writer - {@link DataStreamWriter} implementation
      * @param format - source format
      * @param settings - operation settings
      * @return {@code CompletableFuture<InsertResponse>} - a promise to insert response
      */
     public CompletableFuture<InsertResponse> insert(String tableName,
-                                     List<String> columns,
+                                     List<String> columnNames,
                                      DataStreamWriter writer,
                                      ClickHouseFormat format,
                                      InsertSettings settings) {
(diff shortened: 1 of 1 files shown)


client-v2/src/main/java/com/clickhouse/client/api/Client.java
```diff
@@ -1425,9 +1425,23 @@ public CompletableFuture<InsertResponse> insert(String tableName, List<?> data,
...
      */
     public CompletableFuture<InsertResponse> insert(String tableName, InputStream data, ClickHouseFormat format) {
-        return insert(tableName, data, format, new InsertSettings());
+        return insert(tableName, Collections.emptyList(), data, format, new InsertSettings());
     }
 
+    /**
+     * <p>Sends write request to database. Input data is read from the input stream.</p>
+     *
+     * @param tableName - destination table name
```
chernser, line 1434: please update javadoc according to the method parameters. `columnNames` is missing for example.

client-v2/src/main/java/com/clickhouse/client/api/Client.java
```diff
@@ -1509,7 +1558,20 @@ public CompletableFuture<InsertResponse> insert(String tableName,
...
         settings.setOption(ClientConfigProperties.INPUT_OUTPUT_FORMAT.getKey(), format.name());
         final InsertSettings finalSettings = settings;
-        final String sqlStmt = "INSERT INTO " + tableName + " FORMAT " + format.name();
+
+        StringBuilder columnString = new StringBuilder();
+        if (columns != null && !columns.isEmpty()) {
+            columnString.append("(");
+            for (int i = 0; i < columns.size(); i++) {
+                columnString.append(columns.get(i));
+                if (i < columns.size() - 1) {
```
chernser, line 1567: with builder you may avoid this check by just cutting 2 chars at the end.
also for loop will be for each loop `for (String column : columns)` (No need to use stream)

client-v2/src/main/java/com/clickhouse/client/api/Client.java
chernser: javadoc missing this parameter

chernser:
```
Sends write request to database. Input data is read from the input stream.
If columnNames is not empty then only listed columns will be used in final insert statement.
```

```
@param columnNames - name of the columns to use in insert
```

chernser (approved):

client-v2/src/main/java/com/clickhouse/client/api/Client.java
```diff
@@ -1509,8 +1558,18 @@ public CompletableFuture<InsertResponse> insert(String tableName,
 
         settings.setOption(ClientConfigProperties.INPUT_OUTPUT_FORMAT.getKey(), format.name());
         final InsertSettings finalSettings = settings;
-        final String sqlStmt = "INSERT INTO " + tableName + " FORMAT " + format.name();
-        finalSettings.serverSetting(ClickHouseHttpProto.QPARAM_QUERY_STMT, sqlStmt);
+
```
mzitnik, line 1561: Should it have a condition if empty `"INSERT INTO " + tableName + " FORMAT " + format.name();` else do the loop

client-v2/src/main/java/com/clickhouse/client/api/Client.java
```diff
@@ -1509,8 +1558,18 @@ public CompletableFuture<InsertResponse> insert(String tableName,
...
+        StringBuilder sqlStmt = new StringBuilder("INSERT INTO ").append(tableName);
```
mzitnik, line 1562: Do we need to add ` table name?

Issue #2290, opened 2025-04-08
Question marks '?' break queries with filters


luizarakaki:
### Describe the bug

When a query, which contains a question mark ? and a Metabase filter is executed agaisnt a Clickhouse database, the ? is replaced by null and the query execution fails.

A similar query on Postgres JDBC doesn't fail

### To Reproduce

1. Prepare the following query in Metabase (using new SQL Query) and select a Clickhouse database
```
WITH 
data as (SELECT '1' as `field?`)
SELECT *
FROM data
WHERE `field?` = {{data_filter}}
```

2. Set the data_filter to type string and value 1
3. Execute the query
4. Query fails with the following error
```
Code: 47. DB::Exception: Unknown expression or function identifier `fieldnull` in scope WITH data AS (SELECT '1' AS `field'1'`) SELECT * FROM data WHERE fieldnull = NULL. (UNKNOWN_IDENTIFIER) (version 24.9.3.128 (official build))
```

When either the ? or WHERE clause is removed, the query is executed successfully.
This behavior cannot be reproduced with a similar query against a PostgreSQL database.

### Expected behavior

The Query returns '1'

#### Environment
* Client version: metabase-clickhouse-driver 1.53.3, used JDBC-v2 `0.8.2`
* Language version:
* OS:

related to https://github.com/metabase/metabase/issues/56364

PR #2302, opened 2025-04-12
Fixes and improvements for multiple problems of jdbc_v2 and client_v2

### LINKED ISSUE #2292: [jdbc-v2] Prepared statement metadata returns null
### Description 

When `com.clickhouse.jdbc.PreparedStatementImpl#getMetaData` called it returns `null`

Expected behavior: return metadata for the prepared statement.

oleg-zinovev:
## Summary

Contains several improvements and fixes for the v2 of the client and driver

Improvements: 
- PreparedStatementImpl: getMetaData returns result set metadata using describe query
- StatementImpl: maxRows handling
- Driver: unload method fixed (DriverManager search driver for unregister by reference equality)
- HttpAPIClientHelper: added sslmode option (presents in v1), support to skip ssl validation 

Fixes:
 - PreparedStatementImpl: ignores question mark (?) in comments, string literals and quoted identifiers
 - StatementImpl: multiline block comments trim
 - HttpAPIClientHelper: null pointer exception with unpooled connection (poolControl is null)

Closes #2292, #2290
## Checklist
- [x] Closes #2292, #2290

### DIFF
--- a/client-v2/src/main/java/com/clickhouse/client/api/ClientConfigProperties.java
+++ b/client-v2/src/main/java/com/clickhouse/client/api/ClientConfigProperties.java
@@ -92,6 +92,8 @@ public enum ClientConfigProperties {
 
     SSL_KEY_STORE_PASSWORD("key_store_password"),
 
+    SSL_MODE("sslmode", "strict", Arrays.asList("strict", "none")),
+
     SSL_KEY("ssl_key"),
 
     CA_CERTIFICATE("sslrootcert"),
--- a/client-v2/src/main/java/com/clickhouse/client/api/ClientConfigProperties.java
+++ b/client-v2/src/main/java/com/clickhouse/client/api/ClientConfigProperties.java
@@ -92,8 +92,6 @@ public enum ClientConfigProperties {
 
     SSL_KEY_STORE_PASSWORD("key_store_password"),
 
-    SSL_MODE("sslmode", "strict", Arrays.asList("strict", "none")),
-
     SSL_KEY("ssl_key"),
 
     CA_CERTIFICATE("sslrootcert"),
--- a/client-v2/src/main/java/com/clickhouse/client/api/internal/HttpAPIClientHelper.java
+++ b/client-v2/src/main/java/com/clickhouse/client/api/internal/HttpAPIClientHelper.java
@@ -58,8 +58,11 @@ import org.apache.hc.core5.util.TimeValue;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+import javax.net.ssl.KeyManager;
 import javax.net.ssl.SSLContext;
 import javax.net.ssl.SSLException;
+import javax.net.ssl.TrustManager;
+import javax.net.ssl.X509TrustManager;
 import java.io.IOException;
 import java.io.InputStream;
 import java.io.OutputStream;
@@ -74,7 +77,10 @@ import java.net.URISyntaxException;
 import java.net.URL;
 import java.net.UnknownHostException;
 import java.nio.charset.StandardCharsets;
+import java.security.KeyManagementException;
 import java.security.NoSuchAlgorithmException;
+import java.security.SecureRandom;
+import java.security.cert.X509Certificate;
 import java.util.Base64;
 import java.util.Collection;
 import java.util.Collections;
@@ -168,6 +174,13 @@ public class HttpAPIClientHelper {
                 );
             } catch (SSLException e) {
                 throw new ClientMisconfigurationException("Failed to create SSL context from certificates", e);
+            }
+        } else if ("none".equals(chConfiguration.get(ClientConfigProperties.SSL_MODE.getKey()))) {
+            try {
+                sslContext = SSLContext.getInstance("TLS");
+                sslContext.init(new KeyManager[0], new TrustManager[]{new TrustAllManager()}, new SecureRandom());
+            } catch (NoSuchAlgorithmException | KeyManagementException e) {
+                throw new ClientException("Failed to create none validating SSL context", e);
             }
         }
         return sslContext;
@@ -380,7 +393,7 @@ public class HttpAPIClientHelper {
 
     public ClassicHttpResponse executeRequest(ClickHouseNode server, Map<String, Object> requestConfig, LZ4Factory lz4Factory,
                                              IOCallback<OutputStream> writeCallback) throws IOException {
-        if (timeToPoolVent.get() < System.currentTimeMillis()) {
+        if (poolControl != null && timeToPoolVent.get() < System.currentTimeMillis()) {
             timeToPoolVent.set(System.currentTimeMillis() + POOL_VENT_TIMEOUT);
             poolControl.closeExpired();
         }
@@ -826,6 +839,24 @@ public class HttpAPIClientHelper {
             }
 
             return count > 0 ? runningAverage / count : 0;
+        }
+    }
+
+    private static final class TrustAllManager implements X509TrustManager {
+
+        @Override
+        public void checkClientTrusted(X509Certificate[] chain, String authType) {
+            // ignore
+        }
+
+        @Override
+        public void checkServerTrusted(X509Certificate[] chain, String authType) {
+            // ignore
+        }
+
+        @Override
+        public X509Certificate[] getAcceptedIssuers() {
+            return new X509Certificate[0];
         }
     }
 }
--- a/client-v2/src/main/java/com/clickhouse/client/api/internal/HttpAPIClientHelper.java
+++ b/client-v2/src/main/java/com/clickhouse/client/api/internal/HttpAPIClientHelper.java
@@ -175,13 +175,6 @@ public class HttpAPIClientHelper {
             } catch (SSLException e) {
                 throw new ClientMisconfigurationException("Failed to create SSL context from certificates", e);
             }
-        } else if ("none".equals(chConfiguration.get(ClientConfigProperties.SSL_MODE.getKey()))) {
-            try {
-                sslContext = SSLContext.getInstance("TLS");
-                sslContext.init(new KeyManager[0], new TrustManager[]{new TrustAllManager()}, new SecureRandom());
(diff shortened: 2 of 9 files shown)


client-v2/src/main/java/com/clickhouse/client/api/ClientConfigProperties.java
```diff
@@ -92,6 +92,8 @@ public enum ClientConfigProperties {
 
     SSL_KEY_STORE_PASSWORD("key_store_password"),
 
+    SSL_MODE("sslmode", "strict", Arrays.asList("strict", "none")),
```
chernser, line 95: How would this parameter work? Just toggle support of SSL? 
How would it work when `https` endpoint is set and `sslmode=none` ?

client-v2/src/main/java/com/clickhouse/client/api/internal/HttpAPIClientHelper.java
```diff
@@ -169,6 +175,13 @@ public SSLContext createSSLContext() {
             } catch (SSLException e) {
                 throw new ClientMisconfigurationException("Failed to create SSL context from certificates", e);
             }
+        } else if ("none".equals(chConfiguration.get(ClientConfigProperties.SSL_MODE.getKey()))) {
+            try {
+                sslContext = SSLContext.getInstance("TLS");
```
chernser, line 180: Here you are going to initialize a sslContex, but ssl mode is "none" and no-one seems will use HTTPs then. Do we need initialize it anyway for some case?

client-v2/src/test/java/com/clickhouse/client/ClientTests.java
```diff
@@ -64,6 +65,12 @@ private static Client[] secureClientProvider() throws Exception {
                         .setPassword("")
                         .setRootCertificate("containers/clickhouse-server/certs/localhost.crt")
                         .build(),
+                new Client.Builder()
```
chernser, line 68: In what case this is used?

jdbc-v2/src/main/java/com/clickhouse/jdbc/Driver.java
```diff
@@ -163,4 +163,8 @@ public static String chSettingKey(String key) {
     public java.util.logging.Logger getParentLogger() throws SQLFeatureNotSupportedException {
         throw new SQLFeatureNotSupportedException("Method not supported", ExceptionUtils.SQL_STATE_FEATURE_NOT_SUPPORTED);
     }
+
+    private static final class DriverHolder {
```
chernser, line 167: I think, we do not need this wrapper class. or do we?

jdbc-v2/src/main/java/com/clickhouse/jdbc/PreparedStatementImpl.java
```diff
@@ -305,7 +303,26 @@ public void setArray(int parameterIndex, Array x) throws SQLException {
     @Override
     public ResultSetMetaData getMetaData() throws SQLException {
         checkClosed();
-        return null;
+        if (this.currentResultSet != null) {
+            return currentResultSet.getMetaData();
+        } else if (statementType != StatementType.SELECT) {
+            return null;
```
chernser, line 309: Should it return an empty result set? 
I think, null is not expected.

jdbc-v2/src/main/java/com/clickhouse/jdbc/PreparedStatementImpl.java
```diff
@@ -305,7 +303,26 @@ public void setArray(int parameterIndex, Array x) throws SQLException {
...
+        }
+
+        String sql = compileSql(sqlSegments);
+        String describe = String.format("describe (\n%s\n)", sql);
```
chernser, line 313: This is already implemented in `com.clickhouse.client.api.Client#getTableSchemaFromQuery(java.lang.String)`

chernser:
@oleg-zinovev thank you for the contribution! 
Would you please exclude: 
- PreparedStatementImpl: getMetaData returns result set metadata using describe query - I'm working on it and have a bigger change to support more cases 
- HttpAPIClientHelper: added sslmode option (presents in v1), support to skip ssl validation - We need to discuss this because why to exclude ssl validation when you do not use it. In other cases we need to support self signed certificates but need another way to do so. 

Thanks!

client-v2/src/main/java/com/clickhouse/client/api/ClientConfigProperties.java
oleg-zinovev: Behavior is completely identical to the v1 version of client:
- sslmode=none disables server certificate validation

You can see v1 code here: com/clickhouse/client/config/ClickHouseDefaultSslContextProvider.java:151

client-v2/src/main/java/com/clickhouse/client/api/internal/HttpAPIClientHelper.java
oleg-zinovev: https://clickhouse.com/docs/integrations/language-clients/java/jdbc-v1#ssl-properties

According to this doc, sslmode means 'Whether to verify SSL/TLS certificate'

client-v2/src/test/java/com/clickhouse/client/ClientTests.java
oleg-zinovev: This code adds testing of https connection with disabled SSL certificate validation.

Configured clients used in the testAddSecureEndpoint test.

jdbc-v2/src/main/java/com/clickhouse/jdbc/PreparedStatementImpl.java
oleg-zinovev: Null is expected according to PreparedStatement javadoc:

` * @return the description of a {@code ResultSet} object's columns or
     *         {@code null} if the driver cannot return a
     *         {@code ResultSetMetaData} object`

oleg-zinovev:
> @oleg-zinovev thank you for the contribution! Would you please exclude:

> * HttpAPIClientHelper: added sslmode option (presents in v1), support to skip ssl validation - We need to discuss this because why to exclude ssl validation when you do not use it. In other cases we need to support self signed certificates but need another way to do so.
> 
> Thanks!


An example of using the ssl mode in our case:

sslmode is convenient for DBeaver (and similar programs). It is much easier (in the case of a self-signed certificate) to disable its verification than to explain how to specify the server certificate correctly.
Unfortunately, in our case, the use of endpoints without encryption is prohibited.

client-v2/src/main/java/com/clickhouse/client/api/ClientConfigProperties.java
oleg-zinovev: @chernser Removed

client-v2/src/main/java/com/clickhouse/client/api/internal/HttpAPIClientHelper.java
oleg-zinovev: @chernser Removed

jdbc-v2/src/main/java/com/clickhouse/jdbc/Driver.java
oleg-zinovev: @chernser Removed

jdbc-v2/src/main/java/com/clickhouse/jdbc/PreparedStatementImpl.java
oleg-zinovev: @chernser Removed

jdbc-v2/src/main/java/com/clickhouse/jdbc/PreparedStatementImpl.java
```diff
@@ -305,7 +303,26 @@ public void setArray(int parameterIndex, Array x) throws SQLException {
     @Override
     public ResultSetMetaData getMetaData() throws SQLException {
         checkClosed();
-        return null;
+        if (this.currentResultSet != null) {
+            return currentResultSet.getMetaData();
+        } else if (statementType != StatementType.SELECT) {
```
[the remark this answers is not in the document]
oleg-zinovev: @chernser Removed

jdbc-v2/src/main/java/com/clickhouse/jdbc/PreparedStatementImpl.java
oleg-zinovev: @chernser Removed

jdbc-v2/src/main/java/com/clickhouse/jdbc/PreparedStatementImpl.java
chernser: Thank you! 

I agree that specification allows to return null, but some frameworks doesn't accept it so we need to implement this another way - at least return an empty result set.

jdbc-v2/src/main/java/com/clickhouse/jdbc/Driver.java
chernser: Thank you!

client-v2/src/main/java/com/clickhouse/client/api/internal/HttpAPIClientHelper.java
chernser: As I've mentioned we will port this feature with some changes. 
Thanks!

client-v2/src/main/java/com/clickhouse/client/api/ClientConfigProperties.java
chernser: Thank you!

chernser (approved):

chernser:
@oleg-zinovev thank you for the update! 
I've run CI and will approve shortly. 

As for SSL strict mode I've created a separate issue https://github.com/ClickHouse/clickhouse-java/issues/2309 
I would like implement it in the way:
- It doesn't confuses users with configuration. SSL mode can be theoretically applied to SSL auth and may be need to have another configuration 
- User may need more configuration over SSL (we had a request for custom SSL factory) 
So yes, we will work on it.

Issue #2286, opened 2025-04-07
If not implemented, null should not be returned


nieqiurong:
https://github.com/ClickHouse/clickhouse-java/blob/7a139706364f2888450405542e8a3df0589653ee/jdbc-v2/src/main/java/com/clickhouse/jdbc/metadata/DatabaseMetaData.java#L1029C1-L1032C1

I see that there are two implementations in the driver package, and the other ClickHouseBaseMetaData implements the getIndexInfo logic, but the implementation of BaseMetaData returns null. It would be better to return an empty ResultSet

chernser:
Good day, @nieqiurong! 
Thank you for reporting it!

PR #2315, opened 2025-04-17
Adjust getIndexInfo to return an empty ResultSet instead of a null value

Paultagoras:
## Summary
* Tweaked DatabaseMetaData.getIndexInfo to return an empty ResultSet rather than null

Closes https://github.com/ClickHouse/clickhouse-java/issues/2286
## Checklist
- [x] Closes #<issue ref>


jdbc-v2/src/main/java/com/clickhouse/jdbc/metadata/DatabaseMetaDataImpl.java
```diff
@@ -1065,7 +1065,14 @@ private static String getDataTypeInfoSql() {
 
     @Override
     public ResultSet getIndexInfo(String catalog, String schema, String table, boolean unique, boolean approximate) throws SQLException {
-        return null;
+        try {
+            String sql = "SELECT null AS TABLE_CAT, null AS TABLE_SCHEM, null AS TABLE_NAME, null AS NON_UNIQUE," +
```
chernser, line 1069: I would expect that we use https://clickhouse.com/docs/sql-reference/statements/show#show-index to get information about indexes.

PR #2316, opened 2025-04-18
[jdbc-v2] Base implementation of RowBinaryFormatWriter support in JDBC

### LINKED ISSUE #2171: [jdbc-v2] JDBC Bulk Insert performance using RowBinary
### Describe your feedback

Improving JDBC Insert performance by using RowBinary protocol instead of `INSERT INTO` SQL  

### Code example
```java
try (PreparedStatement stmt = conn.prepareStatement(insertQuery)) {
    for (int i = 0; i < ROWS; i++) {
        stmt.setInt(1, i);
        stmt.setString(2, "name" + i);
        stmt.addBatch();
    }
    long startBatchTime = System.currentTimeMillis();
    stmt.executeBatch();
    long endBatchTime = System.currentTimeMillis();
    System.out.println("Insertion Time for Final Batch: " + (endBatchTime - startBatchTime) + " ms");
}
```

@chernser: 
RowBinary format works well with many common use cases. JDBC v2 should support it. 
Integration should be done as follows:
1. Complete RowBinaryWriter implementation and test as part of client-v2.
2. Integrate the writer to JDBC v2 code.

Problems to solve: 
1. RowBinaryWriter is designed to work with writer API inside a writer callback. There is no good API to get output buffer outside of this callback. A new API may be need to be drafted specially for cases when output stream should be passed out of the caller scope. 
2. When writer operation return control - output stream is closed and request is finished. In case of JDBC it doesn't work. Because output stream should be open all the time until `execute()` or `executeBatch()` is called. Simplest way is to create a stream and some sync primitives, but may be there is another way to solve the problem. 
3. Currently most logic is within one class of `PreparedStatementImpl`.  However prepared statements may be INSERT or SELECT at a time. Would it make sense to split this two branches to make code super clear? Another concern that while mixing logic it leads to creating a "unused variable" problem in runtime because not everything useful for both cases. 

  

This issued is blocked by https://github.com/ClickHouse/clickhouse-java/issues/2177

chernser:
## Summary
Adds support for RowBinaryFormatWriter and fixes some issues:
- Implementation forks PreparedStatementImpl and uses RowBinaryFormatWriter for simple "INSERT INTO t VALUES ()" case
- Implementation is simply collecting values as encoded bytes and writes out when batch is executed.

Additional work is needed further: 
- support "INSER INTO t (col1, col2) VALUES (?, ?)". It requires better parsing code. 

Closes https://github.com/ClickHouse/clickhouse-java/issues/2171
## Checklist
- [ ] Closes #2171 

### DIFF
--- a/jdbc-v2/src/main/java/com/clickhouse/jdbc/ConnectionImpl.java
+++ b/jdbc-v2/src/main/java/com/clickhouse/jdbc/ConnectionImpl.java
@@ -359,7 +359,14 @@ public class ConnectionImpl implements Connection, JdbcV2Wrapper {
     @Override
     public PreparedStatement prepareStatement(String sql, int resultSetType, int resultSetConcurrency, int resultSetHoldability) throws SQLException {
         checkOpen();
-        return new PreparedStatementImpl(this, sql);
+
+        StatementImpl.StatementType statementType = StatementImpl.parseStatementType(sql);
+        if (statementType == StatementImpl.StatementType.INSERT) {
+            if (!PreparedStatementImpl.FUNC_DETECT_REGEXP.matcher(sql).find()) {
+                return new WriterStatementImpl(this, sql, statementType);
+            }
+        }
+        return new PreparedStatementImpl(this, sql, statementType);
     }
 
     @Override
--- a/jdbc-v2/src/main/java/com/clickhouse/jdbc/PreparedStatementImpl.java
+++ b/jdbc-v2/src/main/java/com/clickhouse/jdbc/PreparedStatementImpl.java
@@ -44,11 +44,13 @@ import java.time.format.DateTimeFormatter;
 import java.time.format.DateTimeFormatterBuilder;
 import java.time.temporal.ChronoField;
 import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.Calendar;
 import java.util.Collection;
 import java.util.Collections;
 import java.util.List;
 import java.util.Map;
+import java.util.regex.Pattern;
 
 public class PreparedStatementImpl extends StatementImpl implements PreparedStatement, JdbcV2Wrapper {
     private static final Logger LOG = LoggerFactory.getLogger(PreparedStatementImpl.class);
@@ -61,23 +63,33 @@ public class PreparedStatementImpl extends StatementImpl implements PreparedStat
 
     private final Calendar defaultCalendar;
 
-    String originalSql;
-    String [] sqlSegments;
-    String [] valueSegments;
-    Object [] parameters;
-    String insertIntoSQL;
-    StatementType statementType;
+    // common fields
+    private final String originalSql;
+    private final String [] sqlSegments;
+    private final Object [] parameters;
+    private final StatementType statementType;
+
+    // insert
+    private String [] valueSegments;
+    private String insertIntoSQL;
 
     private final ParameterMetaData parameterMetaData;
 
     private ResultSetMetaData resultSetMetaData = null;
 
-    public PreparedStatementImpl(ConnectionImpl connection, String sql) throws SQLException {
+    // Detects if any of the arguments is withing function parameters
+    static final Pattern FUNC_DETECT_REGEXP = Pattern.compile(
+            "\\b(?!values?\\b)[A-Za-z_]\\w*\\([^)]*\\?[^)]*\\)",
+            Pattern.CASE_INSENSITIVE);
+
+    private static final Pattern VALUES_PARAMETER_SPLIT = Pattern.compile("\\?(?=(?:[^']*'[^']*')*[^']*$)");
+
+    public PreparedStatementImpl(ConnectionImpl connection, String sql, StatementType statementType) throws SQLException {
         super(connection);
         this.originalSql = sql.trim();
         //Split the sql string into an array of strings around question mark tokens
         this.sqlSegments = splitStatement(originalSql);
-        this.statementType = parseStatementType(originalSql);
+        this.statementType = statementType;
 
         if (this.statementType == StatementType.INSERT) {
             insertIntoSQL = originalSql.substring(0, originalSql.indexOf("VALUES") + 6);
@@ -216,11 +228,11 @@ public class PreparedStatementImpl extends StatementImpl implements PreparedStat
     @Override
     public void clearParameters() throws SQLException {
         checkClosed();
-        if (originalSql.contains("?")) {
-            this.parameters = new Object[sqlSegments.length - 1];
-        } else {
-            this.parameters = new Object[0];
-        }
+        Arrays.fill(parameters, null);
+    }
+
+    int getParametersCount() {
+        return parameters.length;
     }
 
     @Override
@@ -245,6 +257,7 @@ public class PreparedStatementImpl extends StatementImpl implements PreparedStat
     public void addBatch() throws SQLException {
         checkClosed();
         if (statementType == StatementType.INSERT) {
+            // adding values to the end of big INSERT statement.
             addBatch(compileSql(valueSegments));
         } else {
             addBatch(compileSql(sqlSegments));
@@ -256,7 +269,6 @@ public class PreparedStatementImpl extends StatementImpl implements PreparedStat
     public int[] executeBatch() throws SQLException {
         checkClosed();
         if (statementType == StatementType.INSERT && !batch.isEmpty()) {
-            List<Integer> results = new ArrayList<>();
             // write insert into as batch to avoid multiple requests
             StringBuilder sb = new StringBuilder();
             sb.append(insertIntoSQL).append(" ");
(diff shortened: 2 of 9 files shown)


jdbc-v2/src/main/java/com/clickhouse/jdbc/ConnectionImpl.java
```diff
@@ -359,7 +359,14 @@ public Statement createStatement(int resultSetType, int resultSetConcurrency, in
     @Override
     public PreparedStatement prepareStatement(String sql, int resultSetType, int resultSetConcurrency, int resultSetHoldability) throws SQLException {
         checkOpen();
-        return new PreparedStatementImpl(this, sql);
+
+        StatementImpl.StatementType statementType = StatementImpl.parseStatementType(sql);
+        if (statementType == StatementImpl.StatementType.INSERT) {
```
mzitnik, line 364: Should we have it as a feature flag?

jdbc-v2/src/main/java/com/clickhouse/jdbc/WriterStatementImpl.java
```diff
@@ -0,0 +1,457 @@
...
+    @Override
+    public int executeUpdate() throws SQLException {
+        return (int) this.executeLargeUpdate();
+    }
+
+    @Override
+    public long executeLargeUpdate() throws SQLException {
+        checkClosed();
+        int updateCount = 0;
+        InputStream in = new ByteArrayInputStream(out.toByteArray());
```
mzitnik, line 86: Do we plan to write the in-stream and not at once (single big batch)

jdbc-v2/src/main/java/com/clickhouse/jdbc/WriterStatementImpl.java
```diff
@@ -0,0 +1,457 @@
...
+    @Override
+    public void clearParameters() throws SQLException {
+        checkClosed();
+        writer.clearRow();
+    }
+
+    @Override
+    public boolean execute() throws SQLException {
+        executeLargeUpdate();
+        return true;
```
mzitnik, line 264: Should we check if `executeLargeUpdate`  return value is greater than 0 to return true.

jdbc-v2/src/main/java/com/clickhouse/jdbc/internal/JdbcUtils.java
```diff
@@ -285,6 +285,19 @@ public static String escapeQuotes(String str) {
                 .replace("\"", "\\\"");
     }
 
+    private static Pattern UNQUOTE_TABLE_NAME = Pattern.compile(
```
mzitnik, line 288: Did we test the regular expression overhead?

jdbc-v2/src/test/java/com/clickhouse/jdbc/PreparedStatementTest.java
```diff
@@ -564,20 +565,59 @@ void testClearParameters() throws Exception {
...
             ps.setString(2, "testName2");
             ps.setInt(3, 19);
             ps.execute();
         }
     }
+
+    @Test
+    void testBatchInsert() throws Exception {
+        String table = "test_batch";
+        Random rnd = new Random();
```
mzitnik, line 588: I would set the random with a seed and log the seed so if we have some issue, it will be easy to reproduce.

jdbc-v2/src/test/java/com/clickhouse/jdbc/PreparedStatementTest.java
```diff
@@ -564,20 +565,59 @@ void testClearParameters() throws Exception {
...
+            };
+
+            for (int i = 0; i < sql.length; i++) {
+                try (PreparedStatement stmt = conn.prepareStatement(String.format(sql[i], table))) {
+                    Assert.assertEquals(stmt.getClass(), impl[i]);
+                    for (int j = 0; j < 10; j++) {
+                        stmt.setInt(1, rnd.nextInt());
+                        stmt.setFloat(2, rnd.nextFloat());
+                        stmt.setInt(3, rnd.nextInt());
+                        stmt.addBatch();
```
mzitnik, line 612: We should also test with `clearParameters()`  and also extending the data types

jdbc-v2/src/test/java/com/clickhouse/jdbc/PreparedStatementTest.java
```diff
@@ -564,20 +565,59 @@ void testClearParameters() throws Exception {
...
+            try (Statement stmt = conn.createStatement()) {
+                stmt.execute("CREATE TABLE IF NOT EXISTS " + table +
+                        " ( v1 Int32, v2 Float32, v3 Int32) Engine MergeTree ORDER BY ()");
+            }
+
+            String[] sql = new String[]{
+                    "INSERT  INTO \n `%s` \nVALUES (?, multiply(?, 10), ?)", // only string possible
+                    "INSERT  INTO\n `%s` \nVALUES (?, ?, ?)", // row binary writer
+            };
+            Class<?>[] impl = new Class<?>[]{
```
mzitnik, line 600: Is it possible to add a JMH to test show check the differences in performance between two implementation (so we could explain why it is more beneficial to use the rowbinary )

jdbc-v2/src/main/java/com/clickhouse/jdbc/internal/JdbcUtils.java
chernser: This is not very big. But we need a normal parser instead of this pile of regexp's 
We need to have only a few runs through string to collect all information. Currently we do repeat searches over and over.

jdbc-v2/src/test/java/com/clickhouse/jdbc/PreparedStatementTest.java
chernser: added

jdbc-v2/src/test/java/com/clickhouse/jdbc/PreparedStatementTest.java
chernser: yes it is possible. With a feature flag it is possible.

jdbc-v2/src/test/java/com/clickhouse/jdbc/PreparedStatementTest.java
chernser: Will do.

jdbc-v2/src/main/java/com/clickhouse/jdbc/WriterStatementImpl.java
chernser: yes, it will be another change.

jdbc-v2/src/main/java/com/clickhouse/jdbc/WriterStatementImpl.java
chernser: good point - we need to check if update returns a result. In case of `WriterStatementImpl` it should be always `false`

jdbc-v2/src/main/java/com/clickhouse/jdbc/internal/StatementParser.java
```diff
@@ -0,0 +1,265 @@
+package com.clickhouse.jdbc.internal;
```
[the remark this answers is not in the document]
mzitnik: Let's see if the ClickHouse server has a parser testing so we could use the same SQL queries

jdbc-v2/src/main/java/com/clickhouse/jdbc/PreparedStatementImpl.java
```diff
@@ -254,7 +266,7 @@ public void addBatch() throws SQLException {
     @Override
     public int[] executeBatch() throws SQLException {
         checkClosed();
-        if (statementType == StatementType.INSERT && !batch.isEmpty()) {
+        if (statementType == StatementParser.StatementType.INSERT && !batch.isEmpty()) {
             // write insert into as batch to avoid multiple requests
             StringBuilder sb = new StringBuilder();
             sb.append(insertIntoSQL).append(" ");
```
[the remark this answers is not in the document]
chernser: Do you mean composing SQL while setting parameters? 
yes we can. But this is not part of this PR.

jdbc-v2/src/main/java/com/clickhouse/jdbc/PreparedStatementImpl.java
```diff
@@ -57,26 +59,36 @@ public class PreparedStatementImpl extends StatementImpl implements PreparedStat
...
         this.originalSql = sql.trim();
         //Split the sql string into an array of strings around question mark tokens
-        this.sqlSegments = splitStatement(originalSql);
-        this.statementType = parseStatementType(originalSql);
+        this.sqlSegments = parsedStatement.getSqlSegments();
+        this.statementType = parsedStatement.getType();
 
-        if (this.statementType == StatementType.INSERT) {
+        if (this.statementType == StatementParser.StatementType.INSERT) {
             insertIntoSQL = originalSql.substring(0, originalSql.indexOf("VALUES") + 6);
```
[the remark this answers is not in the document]
chernser: Yes, this is the problem. But with parser. This PR should not fix it.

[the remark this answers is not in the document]
mzitnik: Can you open an issues for the parser implementation

mzitnik (approved):

Issue #2025, opened 2024-12-16
Tableschema registration for a table with materialized view requires the registered POJO to include those columns


kprateek777:
### Steps to reproduce
1. Create a table that has some columns that get MATERIALIZED e.g. the table has a Map named Attributes with an entry <m1, value> and __m1__ is materialized behind the scenes
```     
`__m1__` LowCardinality(String) MATERIALIZED Attributes['m1'],
```
2. client.getTableSchema(tableName) gets the schema including the materialized columns which mandates the POJO to include even these columns. This means 
```
client.register(MyPOjo.class, schema);
```
If I don't have them in the POJO then on registration I get the exception 
```
java.lang.IllegalArgumentException: No serializer found for column '__m1__'. Did you forget to register it?
```
3. I should not be adding them to the POJO in the first place
4. I tried getting a schema using a SELECT query that did not have the MATERIALIZED columns which then registered fine but the client does not register the table name only the columns. Afterwards during the call to client.insert () it tries to find the table name to find out the hasDefaults and throws the following exception
```
            boolean hasDefaults = (Boolean)this.tableSchemaHasDefaults.get(tableName);

.
..
.
.
java.lang.NullPointerException: Cannot invoke "java.lang.Boolean.booleanValue()" because the return value of "java.util.Map.get(Object)" is null
        at com.clickhouse.client.api.Client.insert(Client.java:1227)
        at com.clickhouse.client.api.Client.insert(Client.java:1186)
```
5. There is code that prevents the schema from a SELECT query to have a tableName. All of these conditions are a weird loop. 

#### Environment
* Client version: client v2 0.7.1-patch1 or 0.7.1
* Language version: java 17
* OS: Red Hat Enterprise Linux release 8.9 (Ootpa)

#### ClickHouse server
* ClickHouse Server version:
* ClickHouse Server non-default settings, if any:
* `CREATE TABLE` statements for tables involved:
```
CREATE TABLE me.me_v0
(
    `Id` UInt64 CODEC(Delta(8), ZSTD(1)),
    `Name` LowCardinality(String) CODEC(ZSTD(1)),
    `Attributes` Map(LowCardinality(String), String) CODEC(ZSTD(1)),
    `__m1__` LowCardinality(String) MATERIALIZED Attributes['m1'],
    INDEX idx_attr_key mapKeys(Attributes) TYPE bloom_filter(0.01) GRANULARITY 1,
[...]

PR #2320, opened 2025-04-21
Issue 2118

### LINKED ISSUE #2118: client.insert(...) fails if registered class doesn't match all existing columns
### Describe the bug
Attempting an `insert` fails if any column from the target table is not defined in the registered POJO.

### Steps to reproduce
1. Register a POJO associated with any given table. The table should contain at least one additional column that is **not** defined in the POJO.
2. Attempt to call `client.insert(...)` for one of these POJO instances.
3. Receive an `IllegalArgumentException` (`No serializer found for column 'example'. Did you forget to register it?`).

From my understanding, the problem comes from the fact that `client.insert(...)` assumes the resulting `INSERT` command should include values for all existing columns (which should not be the case), throwing an exception in line 1350 of `com.clickhouse.client.api.Client` when this is not the case.

### Expected behaviour
`client.insert(...)` should ideally work even if the registered class is 'missing' some columns from the mapped table. While this can be partially solved by just adding said missing column into the registered class, this particular behavior can be problematic in certain cases

For example, when adding a new column to an existing table, this would theoretically break the mapping of the version running in production - which would not have the corresponding attribute in the POJO class.

### Error log
```
java.lang.IllegalArgumentException: No serializer found for column 'example'. Did you forget to register it?
	at com.clickhouse.client.api.Client.insert(Client.java:1350)
```

Thanks in advance for reading this!

mzitnik:
This pull request introduces support for handling default values in ClickHouse columns, including `MATERIALIZED`, `ALIAS`, and `EPHEMERAL` types. It adds new functionality to skip these columns during data insertion and ensures proper parsing and testing of default expressions. The most important changes are grouped below:

### Enhancements to `ClickHouseColumn` class:

* Added a new `DefaultValue` enum to represent different default types (`DEFAULT`, `MATERIALIZED`, `EPHEMERAL`, `ALIAS`) and associated fields `defaultValue` and `defaultExpression` to the `ClickHouseColumn` class.
* Introduced getter and setter methods for `defaultValue` and `defaultExpression` in the `ClickHouseColumn` class.

### Changes to data insertion logic:

* Updated the `insert` method in `Client.java` to skip columns with default values other than `DEFAULT` during data insertion.
* Modified the `commitRow` method in `RowBinaryFormatWriter.java` to similarly skip columns with `MATERIALIZED`, `ALIAS`, or other non-default values.

### Schema parsing improvements:

* Enhanced `TableSchemaParser` to parse `default_type` and `default_expression` properties from table schema and set them in `ClickHouseColumn`.

### New test cases:

* Added a test case `testWriterWithMaterialize` to validate insertion behavior with columns having various default types (`MATERIALIZED`, `ALIAS`, `EPHEMERAL`).
* Introduced `insertSimplePOJOsWithMaterializeColumn` test to verify POJO insertion with `MATERIALIZED` and `ALIAS` columns.

### Additional utility class:

* Added `SimplePOJO` class with a method to generate table creation SQL for testing columns with default values and default expressions.## Summary

Closes #2118 #2025 
## Checklist
- [X] Closes #2118 #2025

### DIFF
--- a/clickhouse-data/src/main/java/com/clickhouse/data/ClickHouseColumn.java
+++ b/clickhouse-data/src/main/java/com/clickhouse/data/ClickHouseColumn.java
@@ -106,6 +106,21 @@ public final class ClickHouseColumn implements Serializable {
     private Map<Class<?>, Integer> mapKeyToVariantOrdNumMap;
     private Map<Class<?>, Integer> mapValueToVariantOrdNumMap;
 
+    public enum DefaultValue {
+        DEFAULT("Default"),
+        MATERIALIZED("MATERIALIZED"),
+        EPHEMERAL("EPHEMERAL"),
+        ALIAS("ALIAS");
+
+        public final String defaultValue;
+
+        DefaultValue(String defaultValue) {
+            this.defaultValue = defaultValue;
+        }
+    }
+
+    private DefaultValue defaultValue;
+    private String defaultExpression;
 
     private static ClickHouseColumn update(ClickHouseColumn column) {
         column.enumConstants = ClickHouseEnum.EMPTY;
@@ -852,6 +867,14 @@ public final class ClickHouseColumn implements Serializable {
     public void setHasDefault(boolean hasDefault) {
         this.hasDefault = hasDefault;
     }
+
+    public void setDefaultValue(DefaultValue defaultValue) { this.defaultValue = defaultValue; }
+
+    public DefaultValue getDefaultValue() { return defaultValue; }
+
+    public void setDefaultExpression(String defaultExpression) { this.defaultExpression = defaultExpression; }
+
+    public String getDefaultExpression() { return defaultExpression; }
 
     public boolean isLowCardinality() {
         return !lowCardinalityDisabled && lowCardinality;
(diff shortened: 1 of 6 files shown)


clickhouse-data/src/main/java/com/clickhouse/data/ClickHouseColumn.java
```diff
@@ -106,6 +106,21 @@ public final class ClickHouseColumn implements Serializable {
     private Map<Class<?>, Integer> mapKeyToVariantOrdNumMap;
     private Map<Class<?>, Integer> mapValueToVariantOrdNumMap;
 
+    public enum DefaultValue {
+        DEFAULT("Default"),
+        MATERIALIZED("MATERIALIZED"),
```
chernser, line 111: Should it be like Default - "Materialized" ?

clickhouse-data/src/main/java/com/clickhouse/data/ClickHouseColumn.java
```diff
@@ -106,6 +106,21 @@ public final class ClickHouseColumn implements Serializable {
...
+        EPHEMERAL("EPHEMERAL"),
+        ALIAS("ALIAS");
+
+        public final String defaultValue;
```
chernser, line 115: May be lets rename it to `kind` or `type` just to avoid too many `defaultValue` in one place?

clickhouse-data/src/main/java/com/clickhouse/data/ClickHouseColumn.java
mzitnik: According to https://clickhouse.com/docs/sql-reference/statements/create/table#default_values all are `DefaultValue`

chernser (approved):

clickhouse-data/src/main/java/com/clickhouse/data/ClickHouseColumn.java
mzitnik: See my last comment, I prefer to stick with ClickHouse documentation language

Issue #2322, opened 2025-04-22
Is it correct for the supportsBatchUpdates() method of the DatabaseMetaData class to return false?


yuelongsun:
I am using JDBC driver version 0.8.4 in combination with Spring's JdbcTemplate to perform write operations.
In the batchUpdate method of JdbcTemplate, there is the following code. Through debugging, I found that since the method com.clickhouse.jdbc.metadata.DatabaseMetaData.supportsBatchUpdates() returns false, batch writing cannot be executed. As a result, the else block is executed, and the writes are performed one by one (row by row).

if (JdbcUtils.supportsBatchUpdates(ps.getConnection())) {
	for (int i = 0; i < batchSize; i++) {
		pss.setValues(ps, i);
		if (ipss != null && ipss.isBatchExhausted(i)) {
			break;
		}
		ps.addBatch();
	}
	return ps.executeBatch();
}
else {
	List<Integer> rowsAffected = new ArrayList<>();
	for (int i = 0; i < batchSize; i++) {
		pss.setValues(ps, i);
		if (ipss != null && ipss.isBatchExhausted(i)) {
			break;
		}
		rowsAffected.add(ps.executeUpdate());
	}
	int[] rowsAffectedArray = new int[rowsAffected.size()];
	for (int i = 0; i < rowsAffectedArray.length; i++) {
		rowsAffectedArray[i] = rowsAffected.get(i);
	}
	return rowsAffectedArray;
}

yuelongsun:
![Image](https://github.com/user-attachments/assets/ad8876dc-4676-4c7d-9ddb-e6b28db67407)

yuelongsun:
This is method batchUpdate of JdbcTemplate : 

public int[] batchUpdate(String sql, final BatchPreparedStatementSetter pss) throws DataAccessException {
		if (logger.isDebugEnabled()) {
			logger.debug("Executing SQL batch update [" + sql + "]");
		}

		int[] result = execute(sql, (PreparedStatementCallback<int[]>) ps -> {
			try {
				int batchSize = pss.getBatchSize();
				InterruptibleBatchPreparedStatementSetter ipss =
						(pss instanceof InterruptibleBatchPreparedStatementSetter ?
						(InterruptibleBatchPreparedStatementSetter) pss : null);
				if (JdbcUtils.supportsBatchUpdates(ps.getConnection())) {
					for (int i = 0; i < batchSize; i++) {
						pss.setValues(ps, i);
[...]

PR #2326, opened 2025-04-22
Quick Update to DatabaseMetaDataImpl.java

Paultagoras:
## Summary
* Switch `supportsBatchUpdates()` from `false` to `true`

Closes https://github.com/ClickHouse/clickhouse-java/issues/2322
## Checklist
- [x] Closes #<issue ref>


chernser (approved):

jdbc-v2/src/main/java/com/clickhouse/jdbc/metadata/DatabaseMetaDataImpl.java
```diff
@@ -1087,62 +1087,52 @@ public boolean supportsResultSetConcurrency(int type, int concurrency) throws SQ
 
     @Override
     public boolean ownUpdatesAreVisible(int type) throws SQLException {
-        // TODO: should be true
```
mshustov, line 1090: what does this TODO mean?

chernser: it means that we need to review what we return here. name of the method tells nothing about what method should do unfortunately

PR #2337, opened 2025-04-24
Solves issue-2191

### LINKED ISSUE #2191: In JDBC V1 version it had the option to disable an extra query that is make to the server
In JDBC V1 version it had the option to disable an extra query (`select currentUser() user, timezone() timezone, version()...`) that is made to the server by setting server & time zone values.
In the JDBC new implementation `this.client.loadServerInfo(); ` is triggered at `ConnectionImpl` 
It would be great if we had a similar mechanism as well  

### Code example for JDBC V1 implementation 
```java
properties.put("server_version", "24.12");
properties.put("server_time_zone", "UTC");
```

mzitnik:
## Summary
If the client wants to disable the retrieval of the server version & server timezone (extra query when init of JDBC Connection)
Example (how to disable)
```
properties.put(ClientConfigProperties.SERVER_TIMEZONE.getKey(), "GMT");
properties.put(ClientConfigProperties.SERVER_VERSION.getKey(), "1.0.0");
```
Closes #2191
## Checklist
- [X] Closes #2191

### DIFF
--- a/client-v2/src/main/java/com/clickhouse/client/api/Client.java
+++ b/client-v2/src/main/java/com/clickhouse/client/api/Client.java
@@ -192,16 +192,24 @@ public class Client implements AutoCloseable {
      *
      */
     public void loadServerInfo() {
-        try (QueryResponse response = this.query("SELECT currentUser() AS user, timezone() AS timezone, version() AS version LIMIT 1").get()) {
-            try (ClickHouseBinaryFormatReader reader = this.newBinaryFormatReader(response)) {
-                if (reader.next() != null) {
-                    this.configuration.put(ClientConfigProperties.USER.getKey(), reader.getString("user"));
-                    this.configuration.put(ClientConfigProperties.SERVER_TIMEZONE.getKey(), reader.getString("timezone"));
-                    serverVersion = reader.getString("version");
+        // only if 2 properties are set disable retrieval from server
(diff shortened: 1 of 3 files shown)


client-v2/src/main/java/com/clickhouse/client/api/Client.java
```diff
@@ -192,16 +192,24 @@ private Client(Set<String> endpoints, Map<String,String> configuration, boolean
...
      */
     public void loadServerInfo() {
-        try (QueryResponse response = this.query("SELECT currentUser() AS user, timezone() AS timezone, version() AS version LIMIT 1").get()) {
-            try (ClickHouseBinaryFormatReader reader = this.newBinaryFormatReader(response)) {
-                if (reader.next() != null) {
-                    this.configuration.put(ClientConfigProperties.USER.getKey(), reader.getString("user"));
-                    this.configuration.put(ClientConfigProperties.SERVER_TIMEZONE.getKey(), reader.getString("timezone"));
-                    serverVersion = reader.getString("version");
+        // only if 2 properties are set disable retrieval from server
+        if (!this.configuration.containsKey(ClientConfigProperties.SERVER_TIMEZONE.getKey()) && !this.configuration.containsKey(ClientConfigProperties.SERVER_VERSION.getKey())) {
```
chernser, line 196: I think. in the future it will be a bug when more then timezone is get with server info. 
So I'd just have an explicit flag documenting that all required configuration should be set.  In the builder test that all required configuration is set

chernser (approved):

Issue #2375, opened 2025-05-19
[jdbc-v2] ResultSet#findColumn always returns 0


gravity182:
## Description

The `ResultSet#findColumn` method returns 0 for any column.

### Steps to reproduce
1. Obtain ResultSet
2. Call `findColumn` for either of the columns' labels.

### Expected Behaviour

The method should return an actual column's index.

### Code Example

Let's assume it's the flow inside some jdbc result mapper:
```java
ResultSet _rs = ...

int _idx_uuid = _rs.findColumn("uuid"); // returns 0
String uuid = _rs.getString(_idx_uuid) // throws NoSuchColumnException
```

The code above throws `NoSuchColumnException` since `getString` expects the index parameter to start from 1. Given that `findColumn` returns 0, it'll always throw an Exception.

#### Environment
* Client version: jdbc-v2-0.8.5
* Language version: Java 17
* OS: Windows 10

#### ClickHouse Server
* ClickHouse Server version: 25.4

PR #2376, opened 2025-05-19
Fix `ResultSet#findColumn` - return actual column's index instead of 0

gravity182:
## Summary
Now `findColumn` returns actual column's index.

Closes https://github.com/ClickHouse/clickhouse-java/issues/2375.

The reason this method returns 0 is the following: this method obtains an instance of `ClickHouseColumn` to return its index. However, its private field `columnIndex` always equals 0 for any write/read path. This field is set in the constructor initially and never updated to reflect the actual column's index - https://github.com/gravity182/clickhouse-java/blob/a3f80b6476c6832a0f8e09c88b65f55e86b4e270/clickhouse-data/src/main/java/com/clickhouse/data/ClickHouseColumn.java#L670.

Interestingly, this method worked correctly in jdbc-v1. See the implementation - https://github.com/gravity182/clickhouse-java/blob/3ce1b09161e15bfd13695d7fb7bdd198b4672e00/clickhouse-jdbc/src/main/java/com/clickhouse/jdbc/ClickHouseResultSet.java#L219-L225. As you can see, it simply counted from 1 until the required column is found.

Having said that, we can proceed with this solution fixing only the `findColumn` method or fix this problem on the deeper level, setting `ClickHouseColumn#columnIndex` field to a correct column's index. There probably might be other methods affected by this problem as well.

## Checklist
- [x] Closes https://github.com/ClickHouse/clickhouse-java/issues/2375

### DIFF
--- a/client-v2/src/main/java/com/clickhouse/client/api/metadata/TableSchema.java
+++ b/client-v2/src/main/java/com/clickhouse/client/api/metadata/TableSchema.java
@@ -104,10 +104,23 @@ public class TableSchema {
         return indexToName(index - 1);
     }
 
+    /**
+     * Takes column name and return corresponding index (starting from 1).
+     * Equals to {@code nameToIndex(name) + 1}.
+     *
+     * @param name - column name
+     * @return - column index starting from 1
+     */
     public int nameToColumnIndex(String name) {
         return nameToIndex(name) + 1;
     }
 
+    /**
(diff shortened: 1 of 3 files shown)


chernser (approved):

chernser:
Thank you, @gravity182! 
Excellent PR!

PR #2382, opened 2025-05-22
Set custom query id using ThreadLocal

skauppin:
## Summary

Older JDBC driver had the possibility to set a custom query ID using `com.clickhouse.client.ClickHouseRequestManager`. Seems like the latest version doesn't have a way to do the same. We have a use case where we need to check whether an database query (maybe an insert also) is already running and not start it again if it is. This may happen e.g. if the operation times out on the client side.

How to now set a custom query ID
```java
Client.setThreadQueryId("my-query-id");

// execute a query

Client.removeThreadQueryId();
```

This seems to set the query Id as we want, even in the case that we are using Spring JdbcTemplate for executing db operations. `com.clickhouse.client.api.query.QuerySettings` has a setter and getter for a query ID, but we did not find any way to actually set it that way.

### DIFF
--- a/client-v2/src/main/java/com/clickhouse/client/api/Client.java
+++ b/client-v2/src/main/java/com/clickhouse/client/api/Client.java
@@ -114,6 +114,23 @@ import static java.time.temporal.ChronoUnit.SECONDS;
  */
 public class Client implements AutoCloseable {
 
+    private static ThreadLocal<String> customQueryIdValue = new ThreadLocal<>();
+
+    /**
+     * Set custom query ID for the current thread. This query ID will be used
+     * for any query triggered from the same thread.
+     */
+    public static void setThreadQueryId(String customQueryId) {
+         customQueryIdValue.set(customQueryId);
+    }
+
+    /**
+     * Remove custom query ID for the current thread.
+     */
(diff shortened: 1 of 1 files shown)


chernser:
Good day, @skauppin ! 

Do you need client to check if query is running? (It would be great if you describe your need in an issue) 

Here is an example how it supposed to be used: 
```java 
 settings = new InsertSettings()
                .setDeduplicationToken(RandomStringUtils.randomAlphabetic(36))
                .setQueryId(String.valueOf(UUID.randomUUID()));
           InsertResponse response = client.insert(tableName, simplePOJOs, settings).get(EXECUTE_CMD_TIMEOUT, TimeUnit.SECONDS);

```

It is up to application how to implement checking queryID.

As for JDBC - it should be implemented in JDBC code. But we need kind of feature description to understand how to implement.

Thanks!

chernser:
@skauppin 
I appreciate the example and will look into it this week. 
I agree we need to support the Spring and will try to find suitable way for this.

chernser:
created an issue https://github.com/ClickHouse/clickhouse-java/issues/2424
Will be handled later

PR #2465, opened 2025-06-23
Fix a small bug with server metrics parsing + a few code cleanups.

avshenuk:
## Summary
- Fix bug with incorrect usage of Integer where Long is expected.
Metrics returned from the server can overflow Integer e.g. when elapsed time is higher than 2 seconds (it's in nanoseconds).
- Use generic name instead of specific metric name.
- Update misleading Javadocs for client side metrics.
Stopwatch metrics always convert nanoseconds to milliseconds when getLong() is called on them.


chernser:
@avshenuk thank you for the contribution!

chernser (approved):

Issue #2414, opened 2025-06-04
[JDBC] StatementImpl


chernser:
## Description

There are some issues that should be addressed in `StatementImpl`: 
- add more test for uncovered methods
- test `close()` in concurrent case and check all child resources are properly closed. Same for `cancel()`
- `enquoteLiteral(String val)` needs review 

Implement next methods:
- `isClosed()`
- `setLargeMaxRows(long max)`
- `getLargeMaxRows()` 
- `enquoteIdentifier`
- `isSimpleIdentifier(String identifier)` 
- `enquoteNCharLiteral(String val)`

PR #2471, opened 2025-06-25
[jdbc-v2] Fix statement impl

chernser:
## Summary

Closes https://github.com/ClickHouse/clickhouse-java/issues/2414
## Checklist
- [ ] Closes #<issue ref>

### DIFF
--- a/client-v2/src/main/java/com/clickhouse/client/api/sql/SQLUtils.java
+++ b/client-v2/src/main/java/com/clickhouse/client/api/sql/SQLUtils.java
@@ -1,5 +1,8 @@
 package com.clickhouse.client.api.sql;
 
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+
 public class SQLUtils {
     /**
      * Escapes and quotes a string literal for use in SQL queries.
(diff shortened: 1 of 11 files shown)


client-v2/src/main/java/com/clickhouse/client/api/sql/SQLUtils.java
```diff
@@ -0,0 +1,131 @@
...
+     * - Is empty
+     * - Contains any non-alphanumeric characters except underscore
+     * - Starts with a digit
+     * - Is a reserved keyword (not implemented in this basic version)
+     * 
+     * @param identifier the identifier to check
+     * @return true if the identifier needs to be quoted, false otherwise
+     */
+    private static boolean needsQuoting(String identifier) {
+        if (identifier.isEmpty()) {
```
mzitnik, line 69: Should we check here if the identifier is null?

mzitnik:
We get this [56.4% Coverage on New Code](https://sonarcloud.io/component_measures?id=ClickHouse_clickhouse-java&pullRequest=2471&metric=new_coverage&view=list) (required ≥ 80%)
Looking into `SQLUtils`, it, it seems like zero new code coverage. 
Let's take a deep look into the code from a coverage perspective

client-v2/src/main/java/com/clickhouse/client/api/sql/SQLUtils.java
chernser: Yes, I've added IllegalArgumentException

chernser:
@mzitnik I've fixed the coverage report 
New coverage is 62% of a new code - but not all code is new.

PR #2484, opened 2025-07-02
Database names with dashes

### LINKED ISSUE #2463: Dash in database name is not supported by V2 client
It looks like parsing of JDBC URL does not support dash (hyphen) in database name. This results in `Code: 81. DB::Exception: Database test does not exist.` when trying to connect to database named `test-dash`.

Found in 0.8.4 but this part of code in `master` is the same.

RegEx at https://github.com/ClickHouse/clickhouse-java/blob/b2692a32f8de5f96b4a4ec2319057df0514c0c58/jdbc-v2/src/main/java/com/clickhouse/jdbc/internal/JdbcConfiguration.java#L129 shows the following when testing:
```
>>> import re
>>> re.match('(https?:)?\\/\\/([\\w\\.\\-]+|\\[[0-9a-fA-F:]+\\]):?([\\d]*)(?:\\/([\\w]+))?\\/?\\??(.*)$', 'https://host-name:8123/test-dash?param=pampam').groups()
('https:', 'host-name', '8123', 'test', '-dash?param=pampam')
```
(Tested in python as it was a quick test but I assume `\w` is the same)
Note the 4th group only has `test` and `-dash` is already in 5th group where params should reside.

Adding `-` to characters set seems to fix parsing:
```
> re.match('(https?:)?\\/\\/([\\w\\.\\-]+|\\[[0-9a-fA-F:]+\\]):?([\\d]*)(?:\\/([\\w-]+))?\\/?\\??(.*)$', 'https://host-name:8123/test-dash?param=pampam').groups()
('https:', 'host-name', '8123', 'test-dash', 'param=pampam')
```
Here the 4th group contains full DB name `test-dash`

There was already a similar issue long ago: #251 but that was for client V1.

enqueue:
## Summary

Try to support database names with dashes. See https://github.com/ClickHouse/clickhouse-java/issues/2463

Closes https://github.com/ClickHouse/clickhouse-java/issues/2463

## Checklist
- [ ✅  ] Closes #2463
- [ ✅ ] Unit and integration tests covering the common scenarios were added

### DIFF
--- a/jdbc-v2/src/main/java/com/clickhouse/jdbc/internal/JdbcConfiguration.java
+++ b/jdbc-v2/src/main/java/com/clickhouse/jdbc/internal/JdbcConfiguration.java
@@ -6,6 +6,9 @@ import com.clickhouse.jdbc.Driver;
 import com.google.common.collect.ImmutableMap;
 
 import java.net.URI;
+import java.net.URISyntaxException;
+import java.net.URLDecoder;
+import java.nio.charset.StandardCharsets;
 import java.sql.DriverPropertyInfo;
 import java.sql.SQLException;
 import java.util.Comparator;
@@ -13,18 +16,20 @@ import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.Properties;
-import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 import java.util.stream.Collectors;
 
 public class JdbcConfiguration {
-    private static final org.slf4j.Logger log = org.slf4j.LoggerFactory.getLogger(JdbcConfiguration.class);
-    public static final String PREFIX_CLICKHOUSE = "jdbc:clickhouse:";
-    public static final String PREFIX_CLICKHOUSE_SHORT = "jdbc:ch:";
 
-    public static final String USE_SSL_PROP = "ssl";
+    private static final String PREFIX_CLICKHOUSE = "jdbc:clickhouse:";
+    private static final String PREFIX_CLICKHOUSE_SHORT = "jdbc:ch:";
+    static final String USE_SSL_PROP = "ssl";
 
-    final boolean disableFrameworkDetection;
+    private static final String PARSE_URL_CONN_URL_PROP = "connection_url";
+    private static final Pattern PATTERN_HTTP_TOKEN = Pattern.compile(
+        "[A-Za-z0-9!#$%&'*+\\.\\^_`\\|~-]+");
+
+    private final boolean disableFrameworkDetection;
 
     final Map<String, String> clientProperties;
     public Map<String, String> getClientProperties() {
@@ -70,12 +75,38 @@ public class JdbcConfiguration {
         }
 
         this.connectionUrl = createConnectionURL(tmpConnectionUrl, useSSL);
-        this.isIgnoreUnsupportedRequests= Boolean.parseBoolean(getDriverProperty(DriverProperties.IGNORE_UNSUPPORTED_VALUES.getKey(), "false"));
+        this.isIgnoreUnsupportedRequests = Boolean.parseBoolean(getDriverProperty(DriverProperties.IGNORE_UNSUPPORTED_VALUES.getKey(), "false"));
     }
 
-    public static boolean acceptsURL(String url) {
-        // TODO: should be also checked for http/https
-        return url.startsWith(PREFIX_CLICKHOUSE) || url.startsWith(PREFIX_CLICKHOUSE_SHORT);
+    /**
+     * This method (only) checks if this driver is probably responsible for the
+     * connection as given in {@code url}, no further sanity checks are
+     * performed.
+     *
+     * @param url
+     *            the JDBC connection URL
+     * @return {@link true} if ClickHouse JDBC driver is responsible for
+     *         connection, {@code false} else
+     * @throws SQLException
+     *             if there is a technical error parsing the {@code url}
+     */
+    public static boolean acceptsURL(String url) throws SQLException {
+        if (url == null) {
+            throw new SQLException("URL is null");
+        }
+        if (!url.startsWith(PREFIX_CLICKHOUSE)
+            && !url.startsWith(PREFIX_CLICKHOUSE_SHORT))
+        {
+            return false;
+        }
+        try {
+            URI uri = new URI(url);
+            // make sure uri is used
+            return "jdbc".equals(uri.getScheme());
+        } catch (URISyntaxException urise) {
+            throw new SQLException(
+                "Not a valid URL '" + url + "'. ", urise);
+        }
     }
 
     public String getConnectionUrl() {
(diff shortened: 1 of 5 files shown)


jdbc-v2/src/main/java/com/clickhouse/jdbc/internal/JdbcConfiguration.java
```diff
@@ -126,7 +126,7 @@ private static String stripJDBCPrefix(String url) {
      *     <li>5 - query parameters as is (optional)</li>
      * </ul>
      */
-    private static final Pattern URL_REGEXP = Pattern.compile("(https?:)?\\/\\/([\\w\\.\\-]+|\\[[0-9a-fA-F:]+\\]):?([\\d]*)(?:\\/([\\w]+))?\\/?\\??(.*)$");
+    private static final Pattern URL_REGEXP = Pattern.compile("(https?:)?\\/\\/([\\w\\.\\-]+|\\[[0-9a-fA-F:]+\\]):?([\\d]*)(?:\\/([\\w-]+))?\\/?\\??(.*)$");
```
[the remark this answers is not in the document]
enqueue: No, this is common practice. The more backslashes the more irritation.

jdbc-v2/src/test/java/com/clickhouse/jdbc/ConnectionWithCustomDatabaseTest.java
```diff
@@ -0,0 +1,26 @@
...
+ * Use is subject to license terms.
+ */
+package com.clickhouse.jdbc;
+
+import java.sql.Connection;
+
+import org.testng.Assert;
+import org.testng.annotations.Test;
+
+public class ConnectionWithCustomDatabaseTest extends JdbcIntegrationTest {
```
chernser: Please move to the `com.clickhouse.jdbc.ConnectionTest`: 
1. add `if (isCloud()) { return } ` to skip on the cloud 
2. create database with a dash in name (may be use UUID) and other symbols 
3. pass database name thru properties to `getJdbcConnection(props)` 

Thanks!

chernser: This test might not be needed if configuration test check everything.

jdbc-v2/src/test/java/com/clickhouse/jdbc/internal/JdbcConfigurationTest.java
```diff
@@ -49,6 +51,7 @@ public static Object[][] testConnectionUrlDataProvider() {
                 {"jdbc:clickhouse://localhost:8443/default?param1=value1&custom_header1=val1,val2,val3", "http://localhost:8443", defaultProps, withListParams},
                 {"jdbc:clickhouse://localhost:8443/default?custom_header1=\"role 1,3,4\",'val2',val3&param1=value1", "http://localhost:8443", defaultProps, withListParamsQuotes},
                 {"jdbc:clickhouse://localhost:8443/clickhouse?ssl=true", "https://localhost:8443", defaultProps, useDatabaseSSLParams},
+                {"jdbc:clickhouse://localhost:8443/with-dash", "http://localhost:8443", defaultProps, dashParams},
```
chernser: Please add more test for other symbols. 
Actually even unicode works: 
```
clickhouse :) create database `☺`

CREATE DATABASE `☺`

Query id: 1406603a-70aa-432b-a7de-7cd0380efcc6

Connecting to localhost:9000 as user default.
Connected to ClickHouse server version 25.2.2.

Ok.

```
But this is fine to check only ASCII printable characters like `_` or `$`.

chernser:
@enqueue thank you for the contribution! 
I think the new test class can be removed - anyway you have test in configuration section. 
I will appreciate if you test other symbols in configuration test as well. (But not required) 

Thanks!

chernser:
@enqueue 
Unit tests failed: 
```
Error:  Failures: 
Error:    JdbcConfigurationTest.testConfigurationProperties:73 expected [default1] but found [clickhouse?client_name=test_application&database=default1]
Error:    JdbcConfigurationTest.testConnectionUrl:24 Maps do not have the same size:3 != 5
Error:    JdbcConfigurationTest.testConnectionUrl:23 expected [https://localhost:8443] but found [http://localhost:8443]
Error:    JdbcConfigurationTest.testConnectionUrl:24 Maps do not have the same size:3 != 5
Error:    JdbcConfigurationTest.testConnectionUrl:24 Maps do not have the same size:3 != 5
Error:    JdbcConfigurationTest.testConnectionUrl:24 Maps do not have the same size:2 != 3
Error:    JdbcConfigurationTest.testConnectionUrl:24 Maps do not have the same size:3 != 5

```
Would you please fix? 

I'd also would like to know if this tests failing for your local build? Please let me know if it is not failing - I want to make sure tests are working in all environments. Thanks!

jdbc-v2/src/test/java/com/clickhouse/jdbc/internal/JdbcConfigurationTest.java
enqueue: Thanks for your help. I am going to adjust this PR and add a couple of more test cases.

enqueue:
Hello @chernser I am still working on this PR, and am going to enhance implementation and tests. As always, some easy looking issue leads you into a rabbit hole 🤷‍♂️ .

I _think_ I was able to get properly encoded HTTP header values work.

```bash
[felix@fmueller2021 jdbc-v2]$ echo "SELECT 1" |  curl 'http://localhost:8123/' -H "X-ClickHouse-Database*: UTF-8''%F0%9F%98%8B" --data-binary @-
1
```

Are additional ("phantasy") JDBC URL parameters actually taken into account, i.e. should they end up as query parameters or headers in HTTP call? There is a test for parsing them into `JdbcConfiguration` but I could not find any use of them further down the road.

Is it intended behavior that JDBC URL allows invalid URLs, e.g. non URL-encoded query parameters? This would be different to other popular JDBC drivers, e.g. [MySQL](https://dev.mysql.com/doc/connector-j/en/connector-j-reference-jdbc-url-format.html), [PostgreSQL](https://jdbc.postgresql.org/documentation/use/#connecting-to-the-database)

Furthermore, it looks to me  that JdbcConfiguration does not support multiple hosts via JDBC URL, while the old implementation does. Is this intentional? If not, perhaps a follow-up issue / PR would be good.

chernser:
@enqueue I've merged. Thank you! 
Yes, some simple issues may lead to a rabbit hole but is very important to understand how deep is it to plan next steps. It is fine to do a partial fix and comment with what else should be done if fix will take very long time.

My opinion on the JDBC URL: it is not HTTP one and should not follow HTTP rules. It is very inconvenient to observe encoded values in JDBC connect string ("URL"). 
Besides old drive did not require encoded value and had very relaxed rules - so we left values as is and encode them later.

chernser:
@enqueue I've fixed https://github.com/ClickHouse/clickhouse-java/pull/2486

enqueue:
@chernser Thanks, I tried to limit changes to fixing issue at hand.

:warning: Note that query parameters are currently expected to be URL encoded. To check this simply add an entry to `JdbcConfigurationTest#VALID_URLs` like so:

```java
new JdbcConfigurationTestData("jdbc:clickhouse://localhost?key1=not enc/öd',e&d")
    .withAdditionalExpectedClientProperties(
        Map.of(
            "key1", "not enc/öd',e&d"
        ))
```

The test will fail, because general URI format according to [RFC 2396](https://www.rfc-editor.org/rfc/rfc2396.html) is violated (also see [JDK JavaDoc URI](https://docs.oracle.com/javase/8/docs/api/java/net/URI.html)).

If we want to allow non-encoded values we will have to establish proprietary escaping rules for some characters ("&", "=" come to mind).

chernser:
@enqueue 
Thank you for pointing this out!

URL encoded parameter seems a correct way. At least everyone expects it and other JDBC drivers have this rule. 
We will highlight this change so who affected will encode parameters.  

Using special escaping rules would be hard to adopt.

PR #2485, opened 2025-07-07
Fix JDK8 compile 1

enqueue:


chernser:
@enqueue Thank you! 

However I've fixed the issue. 
few more comments: 
- `IllegalStateException` is about state but we do not have state, so `ClientException` is more suitable here. `IllegalArgumentException` will not work here because exception do not directly linked to any of the arguments that user may specify. 
- Using regexp is too expensive for just detecting blank string. using `trim().isEmpty()` will be faster.

Issue #1870, opened 2024-10-17
[client-v2][discussion] Load-balancing on the client side.  


chernser:
### Topic 
Client-v2 implementation may connect to a single target host today. Client uses Apache HTTP client which has built-in connection pool. This set of properties is enough to handle many use-cases, because we assume that ClickHouse cluster is behind a load-balancer. 
Client-v1 has load-balancing on the client side and can handle failover to a backup node. This mechanism is complex because should track many moving parts. 

Handling load balancing on a client side has some challenges:
- keep track of liveness of all nodes.
- load distribution requires very fine connection control. 
- fair load distribution would need to know load of each pod.

External HTTP load balancer would work better:
- all known load-balancers have liveness check - no need to do it on client.
- proxy knows how many requests were sent to each node - so it is much easier to handle
- having a centralized proxy helps to control traffic to a DB. For example, if some application start acting bad - it can be easily disconnected from cluster.

Proxy would become a single point of failure, but it is lightweight and easy to restart than swarm of pods.

This issue is for the discussion. Please share your thoughts about  pros, cons for both approaches. Thanks!

ashwinsri1:
Hey,
I had a few uncertainties regarding the load balancer(LB) solution:

- If I provide the LB DNS as a connection host and create a connection pool with it, wouldn't it connect to only one of the nodes of the cluster at the time of creation of pool?
- Since the client will be making the connection to the clickhouse cluster, some visibility on the connection pool for each of the nodes is better as I can see the issues/performances of each of the thread from the client logs instead of going to the LB dashboard/logs.

Keeping in mind your point of the client side challenges, I am not at all against using external LB but I need some certainty that when the client will be connecting to the CH cluster, all the resources will be used effectively.

huddedar34:
Hi @chernser ,
We have a similar usecase where we have single shard and 3 replicas. Clickhouse is hosted in 3 separate VMs. We are okay to use external LB. But based on the docs I understand that when we use a DNS it will resolve to one of the nodes and create a connection pool with that node ([Ref](https://github.com/ClickHouse/clickhouse-java/blob/main/client-v2/src/main/java/com/clickhouse/client/api/Client.java#L234)). We want to utilise all the replicas for processing. How can we achieve this using clickhouse java client v2? Do we have any plan to support this? Any short-term solution you can provide?  Thanks in advance!

chernser:
Good day. @ashwinsri1 ! 
Thank you for the great question! 

[...]

PR #2488, opened 2025-07-08
[jdbc-v2] Set default ports, parse multiple endpoints

enqueue:
### Default port 

`Client` endpoint URLs need to have a port set (see [Client](https://github.com/ClickHouse/clickhouse-java/blob/b7ce67cbd875406d5d3b13f73e70fa9948d10c42/client-v2/src/main/java/com/clickhouse/client/api/Client.java#L310)). The `JdbcConfiguration` will use the default ports depending on `ssl` parameter / http(s) prefix.

### Multiple endpoints in JDBC URL

I know that only one endpoint will currently be used, and there is an ongoing discussion in #1870 whether we want to make it the driver's responsibility to maintain liveness states, distribute calls, keep a pool of connections etc. But the current documentation explains that it is possible to give multiple endpoints. Without this PR this would lead to broken connection. With this PR it is more likely that a successful connection can be established. Alternatively "fix" documentation and throw `SQLException` right away when we detect multiple endpoints in JDBC URL. This would make everything easier 😄 . See #2489 which is an alternative to this PR.

## Checklist
- [  ✅ ] Unit and integration tests covering the common scenarios were added

### DIFF
--- a/jdbc-v2/src/main/java/com/clickhouse/jdbc/internal/JdbcConfiguration.java
+++ b/jdbc-v2/src/main/java/com/clickhouse/jdbc/internal/JdbcConfiguration.java
@@ -2,6 +2,7 @@ package com.clickhouse.jdbc.internal;
 
 import com.clickhouse.client.api.Client;
 import com.clickhouse.client.api.ClientConfigProperties;
+import com.clickhouse.client.api.http.ClickHouseHttpProto;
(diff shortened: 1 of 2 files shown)


enqueue:
This is not needed, since [multiple endpoints should not be supported via JDBC URL](https://github.com/ClickHouse/clickhouse-java/pull/2489#issuecomment-3049542720).

PR #2501, opened 2025-07-22
Instant / DateTime64 via parameter

### LINKED ISSUE #2456: Passing an Instant as a query parameter in client-v2 fails during parsing
## Description

The API for query parameters with an `Instant` generates a query which can not be parsed by the server.

### Steps to reproduce

See example below.

```
com.clickhouse.client.api.ServerException: Code: 457. DB::Exception: Value 2025-06-16T11:54:26.413588562Z cannot be parsed as DateTime64 for query parameter 'year_start' because it isn't parsed completely: only 29 of 30 bytes was parsed: 2025-06-16T11:54:26.413588562. (BAD_QUERY_PARAMETER) (version 25.3.3.42 (official build))
```

### Expected Behaviour

It should work smoothly.

### Code Example

```java
  @Test
  void testParamsWithInstant() throws Exception {
    final Client client = new Client.Builder()
        .addEndpoint(Protocol.HTTP, "localhost", 8123, false)
        .setUsername("default")
        .setPassword("")
        .build();
    final Map<String, Object> params = Map.of("year_start", Instant.now());
    // This query should not throw an exception
    client.query("SELECT * FROM system.tables WHERE metadata_modification_time < {year_start:DateTime64}", params).get();
  }

```
Test is passing when serializing the parameter to a String, e.g. `"2025-06-16 11:54:26.413588562"`.

A related query with an array also fails with the following parsing error:

```
com.clickhouse.client.api.ServerException: Code: 130. DB::Exception: Cannot read array from text, expected comma or end of array, found '-': value [2025-06-16T12:03:29.844112439Z] cannot be parsed as Array(DateTime64) for query parameter 'year_start'. (CANNOT_READ_ARRAY_FROM_TEXT) (version 25.3.3.42 (official build)) 
```

The error above was done with:

```
    final Map<String, Object> params = Map.of("year_start", List.of(Instant.now()));
    // This query should not throw an exception
    client.query("SELECT * FROM system.tables WHERE metadata_modification_time in {year_start:Array(DateTime64)}", params).get();

```

#### Client Configuration
See above.

#### Environment
* [ ] Cloud
* Client version:
* Language version:
* OS: 

#### ClickHouse Server
* ClickHouse Server version: 25.3.3.42

enqueue:
## Summary

This PR adds support for supplying a `java.time.Instant` as query parameter to a query. The driver will take care to translate the Java instant into a format ClikcHouse server will understand. In some use cases, user will have to perform the mapping themselves.

This is only the start into a great journey. As outlined in #2456 , clients will also expect this to work with [Array](https://clickhouse.com/docs/sql-reference/data-types/array), probably [Tuple](https://clickhouse.com/docs/sql-reference/data-types/tuple) etc. as well. Nice candidates with time zone madness are also `LocalDateTime` and friends.

This feels very similar to requirements from JDBC, where Java objects have to be serialized somehow (see [PreparedStatementImpl.encodeObject()](https://github.com/ClickHouse/clickhouse-java/blob/2f3b74127c77b173948415424d0effcba92523e2/jdbc-v2/src/main/java/com/clickhouse/jdbc/PreparedStatementImpl.java#L744). We have to be cautious not to have different ways of serializing stuff for ClickHouse queries.

I did not know where to place the integration tests, really. I first started adding a method to `QueryTests`, but I reckoned that this class' primary purpose was to ensure the same behavior using different configuration options (compression etc.) via sub classes. I also considered `DataTypeTests`, but this class mainly deals with some complex, nested stuff. So I created a new test, `com.clickhouse.client.e2e.ParameterizedQueryTest`.  Please let me know if you see a better place for this integration test.

The `format` methods in `DataTypeUtils` create an empty String for `null` values. This might not always be the best default value for every ClickHouse data type. There are different options:

1. Throw NPE or IAE if object is null
2. Provide default String values per data type, either directly in `ClickHouseDataType` or here, in the `DataTypeUils` class.

Some more general remarks:

1. There are many Java warnings which are correct, e.g. about unused imports, unused fields etc. It would really be great if we could improve code quality a little bit (related #304).
2. I was struggling with Java 8 target. There are so many API improvements, language features, bug fixes we are missing out on. Do maintainers have a sound strategy for Java version targets?
3. Then, on the other hand, there is a dependency on `lombok` library. This is not very practical because you have to set up your IDE to pre-compile this stuff etc. Either stick to "old-school" Java, let IDE generate getters/setters for you or go current Java standard, [records](https://docs.oracle.com/en/java/javase/17/language/records.html).

Closes #2456 
## Checklist
- [ ✅ ] Closes #2456
- [ ✅ ] Unit and integration tests covering the common scenarios were added

### DIFF
--- a/client-v2/src/main/java/com/clickhouse/client/api/Client.java
+++ b/client-v2/src/main/java/com/clickhouse/client/api/Client.java
@@ -7,7 +7,6 @@ import com.clickhouse.client.api.data_formats.NativeFormatReader;
 import com.clickhouse.client.api.data_formats.RowBinaryFormatReader;
 import com.clickhouse.client.api.data_formats.RowBinaryWithNamesAndTypesFormatReader;
 import com.clickhouse.client.api.data_formats.RowBinaryWithNamesFormatReader;
-import com.clickhouse.client.api.data_formats.internal.AbstractBinaryFormatReader;
 import com.clickhouse.client.api.data_formats.internal.BinaryStreamReader;
 import com.clickhouse.client.api.data_formats.internal.MapBackedRecord;
 import com.clickhouse.client.api.data_formats.internal.ProcessParser;
@@ -37,7 +36,6 @@ import com.clickhouse.client.api.serde.POJOSerDe;
 import com.clickhouse.client.api.transport.Endpoint;
 import com.clickhouse.client.api.transport.HttpEndpoint;
 import com.clickhouse.client.config.ClickHouseClientOption;
-import com.clickhouse.config.ClickHouseOption;
 import com.clickhouse.data.ClickHouseColumn;
 import com.clickhouse.data.ClickHouseDataType;
 import com.clickhouse.data.ClickHouseFormat;
@@ -1575,7 +1573,9 @@ public class Client implements AutoCloseable {
         Supplier<QueryResponse> responseSupplier;
 
             if (queryParams != null) {
-                settings.setOption("statement_params", queryParams);
+                settings.setOption(
+                    HttpAPIClientHelper.KEY_STATEMENT_PARAMS,
+                    formatQueryParameters(queryParams));
             }
             final QuerySettings finalSettings = new QuerySettings(buildRequestSettings(settings.getAllSettings()));
             responseSupplier = () -> {
@@ -2099,5 +2099,13 @@ public class Client implements AutoCloseable {
         requestSettings.putAll(configuration);
         requestSettings.putAll(opSettings);
         return requestSettings;
+    }
+
+    private Map<String, String> formatQueryParameters(Map<String, Object> queryParams) {
+        HashMap<String, String> newMap = new HashMap<>(queryParams.size());
+        for (String key : queryParams.keySet()) {
+            newMap.put(key, DataTypeUtils.format(queryParams.get(key)));
+        }
+        return newMap;
     }
 }
--- a/client-v2/src/main/java/com/clickhouse/client/api/DataTypeUtils.java
+++ b/client-v2/src/main/java/com/clickhouse/client/api/DataTypeUtils.java
@@ -1,6 +1,12 @@
 package com.clickhouse.client.api;
 
+import java.time.Instant;
+import java.time.ZoneId;
 import java.time.format.DateTimeFormatter;
+import java.util.Arrays;
+import java.util.Objects;
+
+import com.clickhouse.data.ClickHouseDataType;
 
 public class DataTypeUtils {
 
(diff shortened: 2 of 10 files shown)


enqueue:
I added a test for unusual String parameters which fails for multi-line Strings and quoted Strings.

I had to disable the tests because I was not able to make this work.

I think this is related to query parameter parsing downstream, see issues https://github.com/ClickHouse/ClickHouse/issues/70240 and https://github.com/ClickHouse/ClickHouse/issues/69656

chernser:
Good day, @enqueue !
Thank you very much for the contribution - I will try review it asap 

Here are some answer to the general remarks : 
- Yes, we will do cleanup of warning. We could use spotless plugin to clean some of them or at list fail build. 
- Java 8 problem roots from some old library that are using our client. We will phase out java 8 soon. Meanwhile we may be start providing multi release builds for components where using newer java make more performance and development benefits. 
- `Lombok` is used in tests only and only in one place. We do not plan using such things in production code. Sorry for inconvenience - we will think about removing it later. 

> The format methods in DataTypeUtils create an empty String for null values. This might not always be the best default value for every ClickHouse data type.
If there are two many uncertainties  then it is safe to be protective and throw NPE. We should not guess. 

As for integration tests - we mostly test against real DB and just mark tests with `integration` group. I suggest to more `com.clickhouse.client.e2e.ParameterizedQueryTest` to `com.clickhouse.client` package.

client-v2/src/main/java/com/clickhouse/client/api/DataTypeUtils.java
```diff
@@ -19,4 +25,116 @@ public class DataTypeUtils {
...
+            case DateTime:
+            case DateTime32:
+                return String.valueOf(instant.getEpochSecond());
+            default:
+                return formatInstantDefault(instant);
+        }
+    }
+
+    private static String formatInstantDefault(Instant instant) {
+        String nanos = String.valueOf(instant.getNano());
```
chernser: Why not to use a `DateTimeFormatter` ?

client-v2/src/main/java/com/clickhouse/client/api/DataTypeUtils.java
```diff
@@ -19,4 +25,116 @@ public class DataTypeUtils {
...
+     * @return a suitable String representation of {@code object}, or the empty
+     *         String for {@code null} objects
+     */
+    public static String format(Object object, ClickHouseDataType dataTypeHint,
+        ZoneId timeZone)
+    {
+        if (object == null) {
+            return "";
+        }
+        if (object instanceof Instant) {
```
chernser: This is convenient but slow. 
As user should know type of the object they pass we need only `formatInstant` method.

client-v2/src/main/java/com/clickhouse/client/api/Client.java
```diff
@@ -2100,4 +2101,13 @@ private Map<String, Object> buildRequestSettings(Map<String, Object> opSettings)
         requestSettings.putAll(opSettings);
         return requestSettings;
     }
+
+    private Map<String, String> formatQueryParameters(Map<String, Object> queryParams) {
+        HashMap<String, String> newMap = new HashMap<>(queryParams.size());
+        for (String key : queryParams.keySet()) {
+            newMap.put(key, DataTypeUtils.format(queryParams.get(key)));
```
chernser: That is what I was always concerned: making classes too universal and smart. There are problems with this: 
-  code is responsible for too many things
- elevated expectations that class handles everything.


Please remove this part and let user pass formatted value explicitly. We will create a request build with different helping functionality and of different flavors. The core code should remain simple and lightweight.

chernser:
@enqueue please see my comments. 
Thanks!

client-v2/src/main/java/com/clickhouse/client/api/DataTypeUtils.java
enqueue: We will end up with many methods for different data types in the API, but OK...

client-v2/src/main/java/com/clickhouse/client/api/DataTypeUtils.java
enqueue: Good suggestion, thanks.

client-v2/src/main/java/com/clickhouse/client/api/Client.java
enqueue: Consequently, we should change `Map<String, Object> queryParams` to `Map<String, String> queryParams`, and remove the `String.valueOf(v)` part from `HttpAPIClientHelper`. Else users might think they could pass arbitrary Objects and the driver will take care of formatting, serialization.

client-v2/src/main/java/com/clickhouse/client/api/DataTypeUtils.java
chernser: I want that this logic will be in some other abstraction like query builder. 
``` 
    new QueryBuilder()
       .sql("select {param:DateTime64}") 
        .parameter("param", someInstant)
.execute(client); 
```
- `.parameter()` will do conversion like you have implemented with `instanceof` 
- core logic stays simple and only with essential functions 
- anyone can use own formatted values or rely on function

WDYT?

client-v2/src/main/java/com/clickhouse/client/api/Client.java
chernser: This is very true. We need to fix it. 
I will discuss with the team when we can make it as you suggested.

PR #2502, opened 2025-07-23
Fix native build, make shaded JAR work

enqueue:
## Summary

Make native compilation work again.

I noticed that some CI builds are broken because native image was not created. This was due to renaming of artifact suffix from `-shaded` to `-all`.

Furthermore, the examples using Main class etc. did not work because the slf4j API was not included in shaded jar.

I don't know if anyone every uses the driver like this, and v1 is deprecated. But it is still in here and there are still some users, so we should take care to to at least make CI happy :smile:

### DIFF
--- a/clickhouse-jdbc/pom.xml
+++ b/clickhouse-jdbc/pom.xml
@@ -400,7 +400,7 @@
                                 <excludes>
                                     <!-- provided libraries -->
                                     <exclude>io.micrometer:*</exclude>
-                                    <exclude>org.slf4j:slf4j-simple</exclude>
+                                    <exclude>org.slf4j:*</exclude>
                                 </excludes>
                             </artifactSet>
                             <relocations>


clickhouse-jdbc/pom.xml
```diff
@@ -400,7 +400,7 @@
                                 <excludes>
                                     <!-- provided libraries -->
                                     <exclude>io.micrometer:*</exclude>
-                                    <exclude>org.slf4j:*</exclude>
+                                    <exclude>org.slf4j:slf4j-simple</exclude>
```
chernser: please remove, we exclude it on purpose to not force our version of slf4j.
besides we should not embed any slf4j implementation. 

thanks!

enqueue: > we exclude it on purpose to not force our version of slf4j

Without slf4j API, there will be runtime error.

I am not very familiar with slf4j, and I hope that I will never be, but I thought that was the goal of the project: Library authors could say "we log using slf4j API", and clients could bring their own implementations which are compatible with API version, be it log4j, java.util.logging, reload4j,...

Personally, I would never use anything like this. This is a relict from old times, when every project used their own logging framework. For a JDBC driver nowadays there isn't any good reason to use anything but `java.util.logging`. 

> we should not embed any slf4j implementation

No implementation is embedded, only API.

chernser: I agree with you and we are planning to migrate to java logging API. 

Including just API is not an option because it become mandatory part of classpath. 
In most cases it is resolvable. We then can create a bundle for some tools or just move to java logging API.

enqueue: I removed it. This will leave uber jar and native executable unusable, but OK...

chernser (approved):

PR #2507, opened 2025-07-27
Adjust format of ResultSet.getString() for DateTime

### LINKED ISSUE #2408: Change of format during retrieval of DateTime and Int16 using jdbc-V2
## Description
I have encountered different problems retrieving data with new jdbc-v2 client. In particular, I encountered some issues when updating clickhouse-jdbc from release 0.7.2 (which still used jdbc-v1) to release 0.8.5 (which, if I get it right, uses by default jdbc-v2). 
More specifically, my issues are with: 

1. Retrieve of DateTime objects: with release 0.7.2, by doing a SELECT * from a table with a DateTime column, when trying to map the retrieved DateTime to a String, I get a String like "2025-05-30 12:36:22"; when I update clickhouse-jdbc to 0.8.5 release, I get a ZonedDateTime to String object, like "2025-05-30T12:36:22Z[Etc/UTC]" 
2. Retrieve of Int16 objects and map them into Short; similarly, I try to retrieve some Int16 values and map them into Short: in 0.7.2, I used  `((Short) map.get("int64_column")).shortValue()`, while with 0.8.5 release I have to do `((Integer) map.get("int64_column")).shortValue()`
 
I tried to use property `clickhouse.jdbc.v1=true` and, with the DateTime columns problem, I tried with properties: 
`properties.setProperty("date_time_output_format", "iso")`, `properties.setProperty("date_time_output_format", "simple")` and `properties.setProperty("server_time_zone", "UTC")` but nothing changed.

### Steps to reproduce
1. Update clickhouse-jdbc to 0.8.x 
2. Try to query on a table with a DateTime column and map the values into a String 
2.1. Try to query on a table with a Int16 column and map the values into a Short

### LINKED ISSUE #2448: Date32 return ZonedDateTime rather then LocalDate strings
## Description

Only `Date` clickhouse types are formatted and returned as local date strings. `Date32` is returned as LocalDateTime. See https://github.com/ClickHouse/clickhouse-java/blob/b46d249bc894709b9f68962917c56b021dee9652/client-v2/src/main/java/com/clickhouse/client/api/data_formats/internal/AbstractBinaryFormatReader.java#L334-L340

The fix for date was added in https://github.com/ClickHouse/clickhouse-java/pull/2062/files, not sure why Date32 was not added

As a user its surprising that `Date32`  come back as LocalDateTimes. The value returned by clickhouse is only the date, see https://fiddle.clickhouse.com/caed507d-8d72-48c5-92e2-fac7f5dff261

### Steps to reproduce
```
 connection.createStatement().use { stmt ->

                val query = """
                    SELECT 
                        toDate('1900-01-01') AS Date, 
                        toDate32('1900-01-01') AS Date32
                    """.trimIndent()

                stmt.executeQuery(query).use { rs ->
                    while (rs.next()) {
                        val date = rs.getString("Date")
                        val date32 = rs.getString("Date32")
                        println("date=$date, date32=$date32")
                    }
                }
            }
```

prints: `date=1970-01-01, date32=1900-01-01T00:00Z[UTC]`

### Expected Behaviour

Date32 columns are returned as LocalDates. `date=1970-01-01, date32=1900-01-01`

enqueue:
## Summary
This PR makes changes the format of `ResultSet.getString()` for Date32, DateTime, and DateTime32 columns to a more traditional format, i.e.

- YYYY-MM-dd for Date / Date32
- YYYY-MM-dd HH:mm:ss for DateTime / DateTime32

This PR does _not_ change any of the existing time zone magic.

Closes #2408 #2448

## Checklist
- [ :heavy_check_mark: ] Closes #2448
- [ :heavy_check_mark:  ] Unit and integration tests covering the common scenarios were added

### DIFF
--- a/client-v2/src/main/java/com/clickhouse/client/api/data_formats/internal/AbstractBinaryFormatReader.java
+++ b/client-v2/src/main/java/com/clickhouse/client/api/data_formats/internal/AbstractBinaryFormatReader.java
@@ -341,10 +341,16 @@ public abstract class AbstractBinaryFormatReader implements ClickHouseBinaryForm
         } else if (value instanceof ZonedDateTime) {
             ClickHouseDataType dataType = column.getDataType();
             ZonedDateTime zdt = (ZonedDateTime) value;
-            if (dataType == ClickHouseDataType.Date) {
-                return zdt.format(com.clickhouse.client.api.DataTypeUtils.DATE_FORMATTER);
+            switch (dataType) { // should not be null
+                case Date:
+                case Date32:
+                    return zdt.format(DataTypeUtils.DATE_FORMATTER);
+                case DateTime:
+                case DateTime32:
+                    return zdt.format(DataTypeUtils.DATETIME_FORMATTER);
+                default:
+                    return value.toString();
             }
-            return value.toString();
         } else if (value instanceof BinaryStreamReader.EnumValue) {
             return ((BinaryStreamReader.EnumValue)value).name;
         } else if (value instanceof Number ) {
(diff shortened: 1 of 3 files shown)


chernser:
Thank you for the contribution @enqueue! 
I will review today.

PR #2508, opened 2025-07-27
[jdbc] Fix prepared statement impl

### LINKED ISSUE #2391: [jdbc] Parsing statement containing `WITH` clause fails on client side
## Description
0.8.6 Patch broke parameter parsing in WITH block
I think it's related to this: https://github.com/ClickHouse/clickhouse-java/pull/2351

### Code Example
```java
    @Test
    void test_parameters() throws SQLException {
        String sql = """
            WITH
                toDateTime(?) AS target_time
            SELECT *
            FROM table
            """;
        assertEquals(1, new ConnectionImpl(host, properties)
            .prepareStatement(sql)
            .getParameterMetaData()
            .getParameterCount());
    } // pass
``` 

```java
    @Test
    void test_parameters() throws SQLException {
        String sql = """
            WITH
                toDateTime(?) AS target_time,
                select 1 as gr
            SELECT *
            FROM table
            """;
        assertEquals(1, new ConnectionImpl(host, properties)
            .prepareStatement(sql)
            .getParameterMetaData()
            .getParameterCount());
    } // 0.8.5 - pass, 0.8.6 - fail

#### Environment
* [ ] Cloud
* Client version: 0.8.6
* Language version: java
* OS: MacOs

#### ClickHouse Server
* ClickHouse Server version:

### PR DESCRIPTION
**Pull request by chernser**
## Summary

- Improves test coverage for the PreparedStatementImpl
- Adds missing check that all parameters are set before executing a query 
- Does simple type cast. Further work will be done later 

Closes https://github.com/ClickHouse/clickhouse-java/issues/2418
Closes #2391 
## Checklist
- [x] Closes #<issue ref>

### DIFF
--- a/client-v2/src/main/java/com/clickhouse/client/api/DataTypeUtils.java
+++ b/client-v2/src/main/java/com/clickhouse/client/api/DataTypeUtils.java
@@ -30,7 +30,9 @@ public class DataTypeUtils {
     /**
      * Formatter for the DateTime type with nanoseconds.
      */
-    public static DateTimeFormatter DATETIME_WITH_NANOS_FORMATTER = DateTimeFormatter.ofPattern("uuuu-MM-dd HH:mm:ss.nnnnnnnnn");
+    public static DateTimeFormatter DATETIME_WITH_NANOS_FORMATTER = new DateTimeFormatterBuilder().appendPattern("uuuu-MM-dd HH:mm:ss")
+            .appendFraction(ChronoField.NANO_OF_SECOND, 0, 9, true)
+            .toFormatter();
 
     private static final DateTimeFormatter INSTANT_FORMATTER = new DateTimeFormatterBuilder()
         .appendValue(ChronoField.INSTANT_SECONDS)
--- a/client-v2/src/main/java/com/clickhouse/client/api/DataTypeUtils.java
+++ b/client-v2/src/main/java/com/clickhouse/client/api/DataTypeUtils.java
@@ -20,17 +20,17 @@ public class DataTypeUtils {
     /**
      * Formatter for the DateTime type.
      */
-    public static DateTimeFormatter DATETIME_FORMATTER = DateTimeFormatter.ofPattern("yyyy-MM-dd HH:mm:ss");
+    public static DateTimeFormatter DATETIME_FORMATTER = DateTimeFormatter.ofPattern("uuuu-MM-dd HH:mm:ss");
 
     /**
      * Formatter for the Date type.
      */
-    public static DateTimeFormatter DATE_FORMATTER = DateTimeFormatter.ofPattern("yyyy-MM-dd");
+    public static DateTimeFormatter DATE_FORMATTER = DateTimeFormatter.ofPattern("uuuu-MM-dd");
 
     /**
      * Formatter for the DateTime type with nanoseconds.
      */
-    public static DateTimeFormatter DATETIME_WITH_NANOS_FORMATTER = DateTimeFormatter.ofPattern("yyyy-MM-dd HH:mm:ss.nnnnnnnnn");
+    public static DateTimeFormatter DATETIME_WITH_NANOS_FORMATTER = DateTimeFormatter.ofPattern("uuuu-MM-dd HH:mm:ss.nnnnnnnnn");
 
     private static final DateTimeFormatter INSTANT_FORMATTER = new DateTimeFormatterBuilder()
         .appendValue(ChronoField.INSTANT_SECONDS)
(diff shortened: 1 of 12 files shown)

### CODE REVIEW

File: client-v2/src/main/java/com/clickhouse/client/api/DataTypeUtils.java [hunk 1]
```diff
@@ -20,17 +20,19 @@ public class DataTypeUtils {
     /**
      * Formatter for the DateTime type.
      */
-    public static DateTimeFormatter DATETIME_FORMATTER = DateTimeFormatter.ofPattern("yyyy-MM-dd HH:mm:ss");
+    public static DateTimeFormatter DATETIME_FORMATTER = DateTimeFormatter.ofPattern("uuuu-MM-dd HH:mm:ss");
```
mzitnik, line 23: Do you have a link or explanation for doing that, since you do not see that so often (using the format with uuuu)

jdbc-v2/src/main/java/com/clickhouse/data/Tuple.java
```diff
@@ -37,4 +39,22 @@ public String toString() {
         }
         return output;
     }
+
+
+    @Override
+    public boolean equals(Object obj) {
+        if (this == obj) {
```
mzitnik, line 46: Let's check if obj is null

jdbc-v2/src/main/java/com/clickhouse/jdbc/PreparedStatementImpl.java
```diff
@@ -112,23 +119,27 @@ public PreparedStatementImpl(ConnectionImpl connection, String sql, ParsedPrepar
...
+    private String buildSQL() throws SQLException {
         StringBuilder compiledSql = new StringBuilder(originalSql);
         int posOffset = 0;
         int[] positions = parsedPreparedStatement.getParamPositions();
         for (int i = 0; i < argCount; i++) {
             int p = positions[i] + posOffset;
             String val = values[i];
-            compiledSql.replace(p, p+1, val == null ? "NULL" : val);
-            posOffset += val == null ? 0 : val.length() - 1;
+            if (val == null) {
```
mzitnik, line 129: A remark for before and after the string will be great here

jdbc-v2/src/test/java/com/clickhouse/jdbc/PreparedStatementTest.java
```diff
@@ -1156,5 +1260,160 @@ public void testWithInClause() throws Exception {
...
+                   }
+               }
+            }
+        }
+    }
+
+    @DataProvider(name = "testTypeCastsDP")
+    public static Object[][] testTypeCastsDP() {
+        return new Object[][] {
+                {100, ClickHouseDataType.Int8, ClickHouseDataType.Int8},
```
mzitnik, line 1291: {100, ClickHouseDataType.Int8, ClickHouseDataType.Int8} according to method signutare `testTypeCastsWithoutArgument(Object value, SQLType targetType, ClickHouseDataType expectedType)`
SQLType targetType -> ClickHouseDataType expectedType)
I'm missing something here

client-v2/src/main/java/com/clickhouse/client/api/DataTypeUtils.java
chernser: `y` if for year in era and `u` is for just year.

jdbc-v2/src/main/java/com/clickhouse/data/Tuple.java
chernser: ok. The null check is done on `obj instanceof Tuple`.

jdbc-v2/src/main/java/com/clickhouse/jdbc/PreparedStatementImpl.java
chernser: Previously we did not check if all parameters are set. This change just moved lines 122 and 123 below the check. 
`val` is null only when we have not set any value. NULL will be a "NULL" string.

jdbc-v2/src/test/java/com/clickhouse/jdbc/PreparedStatementTest.java
chernser: I've made `ClickHouseDataType` to implement `SQLType` what is valid according to the spec. 
I do always convert any type to `ClickHouseDataType`. In this case first `ClickHouseDataType.Int8` acts as some `SQLType` instance. The second `ClickHouseDataType.Int8` used to check if name of the type returned correctly. For that I use `toTypeName` function.

mzitnik (approved):

Issue #2462, opened 2025-06-19
[jdbc-v2] Exception reading table with JSON column


chernser:
## Description

There is an exception 
```
[HY000] Failed to read value for column test
Unsupported data type with tag 100
```

This probably relates to the type mapping in JDBC 

### Steps to reproduce
1. Create table with JSON column 
2. Try to read all its content 
3.
### Error Log or Exception StackTrace

```
```

### Code Example

```java


```

#### Client Configuration
```java

```

#### Environment
* [ ] Cloud
* Client version:
* Language version:
* OS: 

#### ClickHouse Server
* ClickHouse Server version:
* ClickHouse Server non-default settings, if any:
* `CREATE TABLE` statements for tables involved:
* Sample data for all these tables, use [clickhouse-obfuscator](https://github.com/ClickHouse/ClickHouse/blob/master/programs/obfuscator/Obfuscator.cpp#L42-L80) if necessary

chernser:
Is not reproducible - needs more tests.

vasilyevskiy:
How to reproduce:

```
--Json with schema
CREATE TABLE json_with_schema
(
    ingestDateTime DateTime64(6, 'UTC') DEFAULT now64(6),
    json_field JSON ( id Int32, name String, value Int64)
)
ENGINE = Memory;

INSERT INTO json_with_schema (ingestDateTime, json_field) VALUES
('2025-04-21 10:00:00.123456', '{"id": 1, "name": "Event1", "value": 100}'),
('2025-04-21 10:01:00.654321', '{"id": 2, "name": "Event2", "value": 200}');

SELECT * FROM json_with_schema;
```

100% reproduced in DataGrip

chernser:
Thank you, @vasilyevskiy ! 

I've not reproduced it after some fixes (they are not released yet) with the HEAD version of the client. 
But will re-test with you example. 

UPDATE: 
I've reproduced it on the latest build. Will fix.

chernser:
@vasilyevskiy 
There is a bug reading JSON values defined with path & type. It might need some time to fix. 
The workaround is to read JSON values as string 

```java 
        Properties readProperties = new Properties();
        readProperties.put(ClientConfigProperties.serverSetting("output_format_binary_write_json_as_string"), "1");
        try (Connection conn = getJdbcConnection(readProperties)) {
```

PR #2531, opened 2025-08-12
[client-v2] Adding JSON support with predefined paths 

chernser:
## Summary
- Adds handling JSON columns when some paths are predefined. In this case path type is encoded as usual for RowBinary instead of Dynamic.

Closes https://github.com/ClickHouse/clickhouse-java/issues/2462
## Checklist
- [ ] Closes #<issue ref>

### DIFF
--- a/clickhouse-data/src/main/java/com/clickhouse/data/ClickHouseColumn.java
+++ b/clickhouse-data/src/main/java/com/clickhouse/data/ClickHouseColumn.java
@@ -72,6 +72,7 @@ public final class ClickHouseColumn implements Serializable {
     private static final String KEYWORD_MAP = ClickHouseDataType.Map.name();
     private static final String KEYWORD_NESTED = ClickHouseDataType.Nested.name();
     private static final String KEYWORD_VARIANT = ClickHouseDataType.Variant.name();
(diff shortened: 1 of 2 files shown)


clickhouse-data/src/main/java/com/clickhouse/data/ClickHouseColumn.java
```diff
@@ -658,6 +676,54 @@ public static List<ClickHouseColumn> parse(String args) {
         return Collections.unmodifiableList(c);
     }
 
+    public static final String JSON_MAX_PATHS_PARAM = "max_dynamic_paths";
+    public static final String JSON_MAX_DYN_TYPES_PARAM = "max_dynamic_types";
+    public static final String JSON_SKIP_MARKER = "SKIP";
+
+    public static void parseJSONColumn(String args, List<ClickHouseColumn> nestedColumns, List<String> parameters) {
```
mshustov, line 683: Doesn't it deserve a dedicated set of tests?

chernser: This part is tested via integration tests. I think no need to duplicate same in unit. 
While testing I'm checking coverage to make sure integration tests validate the code.

Issue #2148, opened 2025-02-14
"v2 driver" log messages at "info" level


devurandom:
### Describe the bug

The v2 client spams "v2 driver" log messages at "info" level from: https://github.com/ClickHouse/clickhouse-java/blob/7b7749fb00c848d4ecbec74d54ac8dca8f12de4e/clickhouse-jdbc/src/main/java/com/clickhouse/jdbc/ClickHouseDriver.java#L83

### Steps to reproduce
1. Create a v2 client and make queries with it
2. Observe lots of `INFO [com.clickhouse.jdbc.ClickHouseDriver] v2 driver` log messages

### Expected behaviour

ClickHouse Java client stays silent on INFO level and higher, unless there is something particularly noteworthy.

#### Environment
* Client version: 0.8.0
* Language version: Java 11 / Clojure 1.12
* OS: Linux 6

#### ClickHouse server
* ClickHouse Server version: Cloud
* ClickHouse Server non-default settings, if any: N/A

chernser:
Good day, @devurandom ! 
It was done intentionally because we have changed the internal implementation of JDBC driver. We will remove it. 
Meanwhile please just turn off logger for this class. 
Thanks!

devurandom:
The issue persists with version 0.8.2 and 0.8.4.

chernser:
Good day, @devurandom! 
Sorry about that. We've planned to make overall logging cleanup so this problem will be solved.

PR #2533, opened 2025-08-12
[JDBC-v2] Removed too verbose logs. Lowered some logs level to debug

chernser:
## Summary
- Removes too verbose and unsafe logging 
- Lowers some logs level to `debug`.

Closes https://github.com/ClickHouse/clickhouse-java/issues/2148
## Checklist
- [x] Closes #<issue ref>


chernser:
@mzitnik should we return some specific logs? 
My assumption was that Driver init happens only once at application and has not much logic to debug. 
Please let me know what should be left. 
Thanks!

mzitnik:
In some cases we want to know if the driver was loaded and what version so it's good indication and it only happen once so we do not spam the log too much

chernser:
@mzitnik 
Yes, I understand that sometimes it is useful. But for application we are a library and they want minimal noise in logs from libraries.

Issue #2396, opened 2025-05-29
Couldn't use clickhouse-jdbc 0.8.x with liquibase


eldarst:
## Description
I'm using liquibase version 4.32.0, when liquibase trying to parse clickhouse column meta data, it falls when parsing DATA_TYPE field. It expected to be Integer, and in 0.7.x versions it used to return as Int, but in 0.8.x version it just return String.
### Steps to reproduce
1. Connect to Clickhouse using clickhouse-jdbc 0.8.x
2. Instantiate new Liquibase
3. liquibase.update
### Error Log or Exception StackTrace
[main] INFO com.clickhouse.jdbc.ClickHouseDriver - ClickHouse JDBC driver version: clickhouse-jdbc 0.8.6 (revision: 2d305b7)
[main] INFO com.clickhouse.jdbc.ClickHouseDriver - v2 driver
[main] INFO com.clickhouse.jdbc.Driver - ClickHouse JDBC driver version: 0.8.6
[main] INFO com.clickhouse.jdbc.ClickHouseDriver - ClickHouse JDBC driver version: clickhouse-jdbc 0.8.6 (revision: 2d305b7)
[main] INFO com.clickhouse.jdbc.ClickHouseDriver - v2 driver
[main] INFO com.clickhouse.jdbc.ClickHouseDriver - v2 driver
[main] INFO com.clickhouse.jdbc.metadata.DatabaseMetaDataImpl - getTables: catalog={}, schemaPattern={}, tableNamePattern={}, types={}
мая 29, 2025 4:21:48 PM liquibase.command
ERROR: Exception Details
ERROR: Exception Details

INFO: Logging exception.
ERROR: Exception Primary Class:  SQLException
ERROR: Exception Primary Reason:  Failed to convert String to java.lang.Integer
ERROR: Exception Primary Source:  ClickHouse 25.5.1.2782
мая 29, 2025 4:21:48 PM liquibase.command
INFO: Command execution complete
ERROR: Exception Primary Class:  SQLException

ERROR: Exception Primary Reason:  Failed to convert String to java.lang.Integer

ERROR: Exception Primary Source:  ClickHouse 25.5.1.2782

Exception in thread "main" liquibase.exception.CommandExecutionException: liquibase.exception.ChangeLogParseException: liquibase.exception.UnexpectedLiquibaseException: liquibase.exception.DatabaseException: java.sql.SQLException: Failed to convert String to java.lang.Integer
[...]

PR #2540, opened 2025-08-15
[JDBC-V2] Fix some metadata resultsets return inconsistent data on `getObject()` 

chernser:
## Summary

`DatabaseMetadata` has two methods `getColumns` and `getTypeInfo` which return a `ResultSet`. This result sets should contain ClickHouse data type name to an integer constant. Previously data was calculated on server but SQL became too hard to maintain. Then `MetadataResultSetImpl` was added to fix the issue to let override some columns. However it is not a good solution because need to do many compensation logic. 
This PR implements a new approach:
- Now value function can be added to a column. This lets reader to re-calculate value based on current row. It can be useful to do small transformations 
- Adds ability to ResultSetMetaDataImpl to "hide" last N columns. This lets to add needed information to a result set but hide it from user as they read `ResultSetMetadata` to find out columns.     
New approach do a few modification and what is important based on normal ResultSet implementation. 

Closes https://github.com/ClickHouse/clickhouse-java/issues/2396
## Checklist
- [ ] Closes #<issue ref>

### DIFF
--- a/clickhouse-data/src/main/java/com/clickhouse/data/ClickHouseColumn.java
+++ b/clickhouse-data/src/main/java/com/clickhouse/data/ClickHouseColumn.java
@@ -38,6 +38,7 @@ import com.clickhouse.data.value.array.ClickHouseShortArrayValue;
 import java.io.Serializable;
 import java.lang.reflect.Array;
 import java.math.BigInteger;
+import java.sql.SQLException;
 import java.time.OffsetDateTime;
 import java.util.ArrayList;
 import java.util.Arrays;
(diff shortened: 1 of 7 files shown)


mzitnik (approved):

jdbc-v2/src/main/java/com/clickhouse/jdbc/internal/DetachedResultSet.java
```diff
@@ -0,0 +1,1128 @@
...
+
+    private boolean wasNull;
+
+    private int row;
+
+    private final int lastRow;
+
+    private boolean closed;
+
+    private Map<String, Integer> columnMap;
```
mzitnik, line 66: Can we use ImmutableMap, or at least let's init with column count since we have this information

mzitnik:
Since there was a change in metadata retrieval, let's verify that we did not break metabase, tableau

PR #2549, opened 2025-08-20
[JDBC-V2] fix(`StatementImpl`/`PreparedStatementImpl`): prevent duplicate data insertion by clearing batch after execution for issue-2548

### LINKED ISSUE #2548: [jdbc-v2] Critical: executeBatch reuse causes duplicate data insertion after 0.8.6
A critical regression was introduced in version 0.8.6. The `PreparedStatement` can no longer be safely reused for multiple `executeBatch()` calls.

**Current Behavior:**
When a `PreparedStatement` is reused to call `executeBatch()` multiple times, **it re-inserts the data from previous batches**. This happens because the internal batch state is not cleared after execution, due to the removal of the `cleanBatch()` logic.

**Expected Behavior:**
The `executeBatch()` method should only insert the data that was added to the batch *since the last execution*. The `PreparedStatement` should be reusable without leaking state from previous operations.

**Impact:**
This causes severe data corruption by creating duplicate entries in the database and violates the standard contract of a `PreparedStatement`.

### Suggested Fix
The batch should be cleaned automatically after execution. A reliable approach, used by implementations like the MySQL connector, is to perform this cleanup in a `finally` block to guarantee it happens even if an exception is thrown.

Proposed code structure for the `executeBatch()` method:
```java
try {
    // ... original execute logic ...
} finally {
    cleanBatch(); // Crucial: Reset the internal batch state
}
```

Test:
```
    static void insertDateWithPreparedStatement(String url, Properties properties) throws SQLException {
        try (Connection conn = DriverManager.getConnection(url, properties)) {//Grab a connection using the jdbc DriverManager
            try (PreparedStatement pstmt = conn.prepareStatement("INSERT INTO " + TABLE_NAME + " VALUES(?, ?, ?, ?)")) {//Create a prepared statement
                pstmt.setDate(1, Date.valueOf("2025-01-01"));//Set the first parameter to '2025-01-01' (using java.sql.Date)
                pstmt.setInt(2, 1);//Set the second parameter to 1
                pstmt.setString(3, "Alice");//Set the third parameter to "Alice"
                pstmt.setObject(4, Collections.singletonMap("key1", "value1"));
                pstmt.addBatch();//Add the current parameters to the batch
                pstmt.executeBatch();//Execute the batch

                pstmt.setObject(1, ZonedDateTime.now());
                pstmt.setInt(2, 2);//Set the second parameter to 2
                pstmt.setString(3, "Bob");//Set the third parameter to "Bob"
                pstmt.setObject(4, Collections.singletonMap("key2", "value2"));
                pstmt.addBatch();//Add the current parameters to the batch

                pstmt.executeBatch();//Execute the batch
            }
        }
    }
```

potoo0:
## Summary
- Override `clearBatch` for `PreparedStatementImpl`
- Clears the internal batch state immediately after executeBatch() to prevent parameter leakage and enable safe statement reuse.

Closes #2548

## Checklist
- [x] Closes #2548

### DIFF
--- a/examples/jdbc/src/main/java/com/clickhouse/examples/jdbc/Basic.java
+++ b/examples/jdbc/src/main/java/com/clickhouse/examples/jdbc/Basic.java
@@ -59,11 +59,18 @@ public class Basic {
                 pstmt.setString(3, "Alice");//Set the third parameter to "Alice"
                 pstmt.setObject(4, Collections.singletonMap("key1", "value1"));
                 pstmt.addBatch();//Add the current parameters to the batch
+                pstmt.executeBatch();//Execute the batch
 
                 pstmt.setObject(1, ZonedDateTime.now());
                 pstmt.setInt(2, 2);//Set the second parameter to 2
                 pstmt.setString(3, "Bob");//Set the third parameter to "Bob"
                 pstmt.setObject(4, Collections.singletonMap("key2", "value2"));
+                pstmt.addBatch();//Add the current parameters to the batch
+
+                pstmt.setObject(1, ZonedDateTime.now());
+                pstmt.setInt(2, 2);//Set the second parameter to 2
+                pstmt.setString(3, "Chris");//Set the third parameter to "Chris"
+                pstmt.setObject(4, Collections.singletonMap("key3", "value3"));
                 pstmt.addBatch();//Add the current parameters to the batch
 
                 pstmt.executeBatch();//Execute the batch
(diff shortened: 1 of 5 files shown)


chernser:
@potoo0 please sign CLA

chernser:
This change requires tests. I see only example code change (what is should not be). 

Thanks!

potoo0:
> This change requires tests. I see only example code change (what is should not be).

@chernser 
Thanks for the review! I've added the tests for this change.

PR #2561, opened 2025-09-04
2324 use post query params (draft)

### LINKED ISSUE #2324: Query parameters can't be sent through POST body
### Use case
Currently, the query parameters are always sent as URI params, which can be problematic when there are numerous params (or long params)

### Describe the solution you'd like
I would like to be able to send the query params in the body of the query instead of the URI. 
The server already supports this as shown here https://github.com/ClickHouse/ClickHouse/issues/8842

The client should offer the option to send query parameters and query in a multipart body like shown in the linked issue

andris-rauda-sumsub:
## Summary
Try implementing sending statement parameters via multipart/form-data to mitigate issues with too long url.

PR is a work in progress.

Tried to implement passing query parameters as `Content-Type: multipart/form-data`, got some ugly mockup (because of a streaming support), that kind of does work, however, it turns out, that when calling it with many parameters from local integration test, it starts failing with a weird server error even sooner, than the original `application/x-www-form-urlencoded` variant:

`com.clickhouse.client.api.ServerException: Code: 1000. DB::Exception: <Unreadable error message> (transport error: 500)`

So, seems, for now it is unfeasible to pursue this.

Closes 
## Checklist
- [ ] Closes #2324

### DIFF
--- a/client-v2/src/test/java/com/clickhouse/client/ClientTests.java
+++ b/client-v2/src/test/java/com/clickhouse/client/ClientTests.java
@@ -216,7 +216,7 @@ public class ClientTests extends BaseIntegrationTest {
                     Assert.assertEquals(config.get(p.getKey()), p.getDefaultValue(), "Default value doesn't match");
                 }
             }
-            Assert.assertEquals(config.size(), 32); // to check everything is set. Increment when new added.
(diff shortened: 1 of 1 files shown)


chernser:
Good day, @andris-rauda-sumsub! 
Thank you for starting the contribution! 

This may be would be easier to implement after we have organize a separate transport layer with multiple http client support.

PR #2566, opened 2025-09-07
Convert values to java.time.Instant

lmartelli:
Make `ResultSetImpl.getObject()` handle `java.time.Instant`


chernser:
Good day, @lmartelli ! 

Issue #2598, opened 2025-09-19
Querying JSON field returns "Failed to read value for column data" or "Failed to read next row"


philip-dbvis:
## Description

When querying certain JSON column I get following errors. 

### Steps to reproduce
**1.**
Create a table with id and JSON as columns.
`CREATE TABLE test (id UInt64, data JSON) ENGINE = MergeTree ORDER BY id;
`
Insert data.

```
INSERT INTO test (id, data) VALUES
(
     1,
    '{"FailedToReadValueForColumnData":{"bar":[{"field1":"value1","field2":"value2","field3":"value3","field4":"value4"},{"field1":"value5","field2":"value6","field3":"value7","field4":"value8"}]}}'
);

INSERT INTO test (id, data) VALUES
(
     2,
    '{"Works":"bye"}'
);
INSERT INTO test (id, data) VALUES
(
     3,
    '{"FailedReadNextRow":{"hi":[{"hi":"byyyyyyyyye"}]}}'
);

```
**2.**
Query them 
```
SELECT * from test where id=1;
SELECT * from test where id=2;
SELECT * from test where id=3;
```
**3.**

You'll see that id 1 gives error with message "Failed to read value for column data", id 2 works, id 3 gives error with message "Failed to read next row"

### Error Log or Exception StackTrace

```
0 = {StackTraceElement@20265} "com.clickhouse.jdbc.internal.ExceptionUtils.toSqlState(ExceptionUtils.java:69)"
1 = {StackTraceElement@20266} "com.clickhouse.jdbc.internal.ExceptionUtils.toSqlState(ExceptionUtils.java:42)"
2 = {StackTraceElement@20267} "com.clickhouse.jdbc.ResultSetImpl.next(ResultSetImpl.java:118)"
```

### Expected Behaviour

To get the full rows back

#### Client Configuration
```java
Using the JDBC driver version 0.9.2
```

chernser:
@philip-dbvis thank you for the issue! 
We will fix it.

chernser:
@philip-dbvis 
do you have stack trace for these errors?

philip-dbvis:
> [@philip-dbvis](https://github.com/philip-dbvis) do you have stack trace for these errors?

I added it in the ticket. Is it not sufficient?

```
0 = {StackTraceElement@20265} "com.clickhouse.jdbc.internal.ExceptionUtils.toSqlState(ExceptionUtils.java:69)"
[...]

Issue #2593, opened 2025-09-17
Support for JSON data type on Dbeaver - Datagrip


shi-sis:
## Description
I'm using the driver 0.9.2 with DBeaver or Datagrip.
When trying to read a table with a JSON column I get the following error : 
SQL Error [HY000]: Failed to read value for column json_data

### Steps to reproduce
1. SELECT * FROM table_with_json_column

### Error Log or Exception StackTrace

eclipse.buildId=unknown
java.version=21.0.5
java.vendor=Eclipse Adoptium
BootLoader constants: OS=win32, ARCH=x86_64, WS=win32, NL=en
Framework arguments:  -eclipse.keyring C:\Users\shimeur\AppData\Roaming\DBeaverData\secure\secure_storage
Command-line arguments:  -os win32 -ws win32 -arch x86_64 -eclipse.keyring C:\Users\shimeur\AppData\Roaming\DBeaverData\secure\secure_storage

org.jkiss.dbeaver.model
Error
Wed Sep 17 09:21:20 CEST 2025
Failed to read value for column json_data

com.clickhouse.client.api.ClientException: Failed to read value for column json_data
	at com.clickhouse.client.api.data_formats.internal.BinaryStreamReader.readValue(BinaryStreamReader.java:259)
	at com.clickhouse.client.api.data_formats.internal.BinaryStreamReader.readValue(BinaryStreamReader.java:90)
	at com.clickhouse.client.api.data_formats.internal.AbstractBinaryFormatReader.readRecord(AbstractBinaryFormatReader.java:179)
	at com.clickhouse.client.api.data_formats.internal.AbstractBinaryFormatReader.next(AbstractBinaryFormatReader.java:249)
	at com.clickhouse.jdbc.ResultSetImpl.next(ResultSetImpl.java:107)
	at org.jkiss.dbeaver.model.impl.jdbc.exec.JDBCResultSetImpl.next(JDBCResultSetImpl.java:282)
	at org.jkiss.dbeaver.model.impl.jdbc.exec.JDBCResultSetImpl.nextRow(JDBCResultSetImpl.java:190)
	at org.jkiss.dbeaver.model.impl.jdbc.struct.JDBCTable.readData(JDBCTable.java:210)
	at org.jkiss.dbeaver.ui.controls.resultset.ResultSetJobDataRead.lambda$0(ResultSetJobDataRead.java:128)
[...]

Issue #2102, opened 2025-01-22
java.sql.SQLException: java.io.IOException: Failed to read boolean value, expect 0 (false) or 1 (true) but we got: 67


sunqiaozz:
When you read clickhouse data using jdbc, the next() function of the ResultSet reports an error when you are halfway through reading ClickHouse data，java.sql.SQLException: java.io.IOException: Failed to read boolean value, expect 0 (false) or 1 (true) but we got: 67

chernser:
Good day, @sunqiaozz ! 
Would you please provide more details about the issue? 
Stacktrace? version? What is query?

sunqiaozz:
Hi,brother

sunqiaozz:
Hi,brother
When I use the JDBC driver to connect to ClickHouse, I get the Connection, and then execute the SQL statement to get the ResultSet, and call the next() method to get the data, this error will occur after reading for some time.

```
ResultSet rs = executeQuery(sql);
while (rs.next()) {}
```



I tried to read ClickHouse's JDBC source code and found that it was **currentRow = rowCursor.next()** in the **next()** method of the **ClickHouseResultSet** class; An error occurred in the code
```
@Override
public boolean next() throws SQLException {
	ensureOpen();

	lastReadColumn = 0;
	boolean hasNext = true;
	if (hasNext()) {
		try {
			currentRow = rowCursor.next();
		} catch (UncheckedIOException e) {
			throw SqlExceptionUtils.handle(e);
		}
		rowNumber++;
	} else {
		currentRow = null;
		hasNext = false;
	}
	return hasNext;
}
```

Then I tried using the latest JDBC driver and it still didn't work

sunqiaozz:
Hi,brother
When I use the JDBC driver to connect to ClickHouse, I get the Connection, and then execute the SQL statement to get the ResultSet, and call the next() method to get the data, this error will occur after reading for some time.

```
ResultSet rs = executeQuery(sql);
while (rs.next()) {}
```
I tried to read ClickHouse's JDBC source code and found that it was **currentRow = rowCursor.next()** in the **next()** method of the **ClickHouseResultSet** class; An error occurred in the code

```
@Override
public boolean next() throws SQLException {
	ensureOpen();

	lastReadColumn = 0;
	boolean hasNext = true;
	if (hasNext()) {
		try {
			currentRow = rowCursor.next();
		} catch (UncheckedIOException e) {
			throw SqlExceptionUtils.handle(e);
		}
[...]

PR #2601, opened 2025-09-25
[Client-V2, JDBC] Fix problem reading JSON nested arrays 

chernser:
## Summary
- Fixes issues reading data from JSON columns with nested arrays of object or other multi-level nested objects

Closes https://github.com/ClickHouse/clickhouse-java/issues/2598
Closes https://github.com/ClickHouse/clickhouse-java/issues/2593
Closes https://github.com/ClickHouse/clickhouse-java/issues/2102

## Checklist
- [x] Closes #<issue ref>


client-v2/src/main/java/com/clickhouse/client/api/data_formats/internal/BinaryStreamReader.java
```diff
@@ -1149,73 +1145,176 @@ public byte[] allocate(int size) {
...
+                byte intervalKind = readByte();
+                type = ClickHouseDataType.intervalKind2Type.get(intervalKind);
+                if (type == null) {
+                    throw new ClientException("Unsupported interval kind: " + intervalKind);
+                }
+                return ClickHouseColumn.of("v", type, false, 0, 0);
             }
-            return ClickHouseColumn.of("v", type, false, 0, 0);
+            case JSON: {
+                byte serializationVersion = readByte();
```
mzitnik, line 1253: `serializationVersion` not in use? Seems to be redundant, or any other plans?

client-v2/src/main/java/com/clickhouse/client/api/data_formats/internal/BinaryStreamReader.java
```diff
@@ -1149,73 +1145,176 @@ public byte[] allocate(int size) {
...
+                return ClickHouseColumn.of("v", "LowCardinality(" + column.getOriginalTypeName() + ")");
+            }
+            case Map: {
+                ClickHouseColumn keyInfo = readDynamicData();
+                ClickHouseColumn valueInfo = readDynamicData();
+                return ClickHouseColumn.of("v", "Map(" + keyInfo.getOriginalTypeName() + "," + valueInfo.getOriginalTypeName() + ")");
+            }
+            case Nested: {
+                int size = readVarInt(input);
+                StringBuilder nested = new StringBuilder();
```
mzitnik: Since we have estimated size can we init with `StringBuilder` with size

jdbc-v2/src/test/java/com/clickhouse/jdbc/DataTypeTests.java
```diff
@@ -1960,4 +1961,48 @@ public void testGeoMultiPolygon() throws Exception {
...
+
+        try (Connection conn = getJdbcConnection(); Statement stmt = conn.createStatement()) {
+            final String sql = "INSERT INTO test_jdbc_json_read (data) VALUES ('%s'), ('{}')";
+            stmt.executeUpdate(String.format(sql, json));
+
+            try (ResultSet rs = stmt.executeQuery("SELECT * FROM test_jdbc_json_read")) {
+
+                assertTrue(rs.next());
+                Object jsonObj = rs.getObject(1);
+                assertTrue(rs.next());
```
mzitnik, line 1992: Should we test the content?

mzitnik (changes requested):
Also we did a lot of work related to other types do we have a good test coverage for that?

client-v2/src/main/java/com/clickhouse/client/api/data_formats/internal/BinaryStreamReader.java
chernser: I need to figure out. but seems not in use.

jdbc-v2/src/test/java/com/clickhouse/jdbc/DataTypeTests.java
```diff
@@ -617,7 +618,7 @@ public void testTimeTypes() throws SQLException {
         Properties createProperties = new Properties();
         createProperties.put(ClientConfigProperties.serverSetting("allow_experimental_time_time64_type"), "1");
         runQuery("CREATE TABLE test_time64 (order Int8, "
-                + "time Time('UTC'), time64 Time64(9) "
+                + "time Time, time64 Time64(9) "
```
chernser, line 625: This  is the fix for falling tests on latest.

chernser:
> Also we did a lot of work related to other types do we have a good test coverage for that?

We had tests for other types. These changes are more code rearrangement to optimize multiple `if-else` 
I will add more tests. 
I have another initiative to somehow unify datatype tests for client itself and jdbc so we can use same set of datasets and queries to test against different classes.

client-v2/src/main/java/com/clickhouse/client/api/data_formats/internal/BinaryStreamReader.java
chernser: it will be a guessing. Default size of StringBuilder is 16 chars. 
So I think we may define `int DEFAULT_STRING_BUILDER_SIZE = 100` should be good. 
Here is an example of tuple definition with total length of 103 chars/bytes: 
```
Tuple(productId Int32, name String, price Float32, added Date, removed Date, storeId UUID, flags UInt8)
```

May be we need to make it configurable or may be add some metric first.

jdbc-v2/src/test/java/com/clickhouse/jdbc/DataTypeTests.java
chernser: Yes. 
I was relying on tests in the client, but here is another story. 
Will add checks.

PR #2603, opened 2025-09-29
feat: add bom

XenoAmess:
## Summary

add bom

Closes 
## Checklist
- [ ] Closes #<issue ref>


chernser:
@XenoAmess
Thank you for contributing! 
However there are next reasons why no BOM is needed: 
* soon we will removed many subprojects
* all libraries and versions are declared in pom.xml in the root folder
* libraries should not have pom because they do not define dependencies for your project. Spring uses BOM because it consists of many different components that can be mixed together.

PR #2579, opened 2025-09-10
[JDBC-V2] Fix Parser Issues

chernser:
## Summary

This PR introduces option to select old JavaCC parser that is proven to work and has more relaxed rules.  Parameter parsing code is ported from older version as most stable logic. And this PR will make old parser as choice by default so it will work almost like JDBC V1. 

And: 
- Fixes CTE issue 
- Fixes issue with using keyword as alias
- Fixes problem when query falsely detected as one without result set 
- Fixes issue with logging 
- Fixes using TIMESTAMP in queries

Closes https://github.com/ClickHouse/clickhouse-java/issues/2574
Closes https://github.com/ClickHouse/clickhouse-java/issues/2568
Closes https://github.com/ClickHouse/clickhouse-java/issues/2537
Closes https://github.com/ClickHouse/clickhouse-java/issues/2595
Closes https://github.com/ClickHouse/clickhouse-java/issues/2617

Closes https://github.com/ClickHouse/clickhouse-java/issues/2569
Closes https://github.com/ClickHouse/clickhouse-java/issues/2570
Closes https://github.com/ClickHouse/clickhouse-java/issues/2571
Closes https://github.com/ClickHouse/clickhouse-java/issues/2572
Closes https://github.com/ClickHouse/clickhouse-java/issues/2573
Closes https://github.com/ClickHouse/clickhouse-java/issues/2609

## Checklist
Delete items not relevant to your PR:
- [x] Closes #<issue ref>
- [x] Unit and integration tests covering the common scenarios were added
- [ ] A human-readable description of the changes was provided to include in CHANGELOG
- [ ] For significant changes, documentation in https://github.com/ClickHouse/clickhouse-docs was updated with further explanations or tutorials

PR #2611, opened 2025-10-09
fix(jdbc-v2): Correctly parse mixed-order CTEs in SqlParser

AVMusorin:
## Summary

The grammar for CTEs was too restrictive. It required a specific order for different CTE syntax styles.

The grammar expected all CTEs to be defined before all CSEs (Common Scalar Expression).

This commit refactors the `ctes` grammar rule to be more flexible. It introduces a new, more generic `cte` rule that can represent either CTE style. The main `ctes` rule is now defined as a list of these generic `cte` elements, allowing them to appear in any order.

## Checklist


chernser:
Good day, @AVMusorin ! 

Actually CTEs were defined incorrectly in the grammar - they were included twice. 
We have fixing it in the https://github.com/ClickHouse/clickhouse-java/pull/2579

PR #2612, opened 2025-10-11
Expose several methods as public

mzitnik:
## Summary
Exposes several methods in SerializerUtils to the public so serialization logic can be reused.

## Checklist
- [ ] Closes #<issue ref>

### DIFF
--- a/client-v2/src/main/java/com/clickhouse/client/api/data_formats/internal/SerializerUtils.java
+++ b/client-v2/src/main/java/com/clickhouse/client/api/data_formats/internal/SerializerUtils.java
@@ -411,6 +411,9 @@ public class SerializerUtils {
         }
     }
 
+    /**
+     DO NOT USE - part of internal API that will be changed
+     */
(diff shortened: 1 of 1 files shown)


client-v2/src/main/java/com/clickhouse/client/api/data_formats/internal/SerializerUtils.java
```diff
@@ -437,7 +437,7 @@ public static void serializeArrayData(OutputStream stream, Object value, ClickHo
         }
     }
 
-    private static void serializeTupleData(OutputStream stream, Object value, ClickHouseColumn column) throws IOException {
+    public static void serializeTupleData(OutputStream stream, Object value, ClickHouseColumn column) throws IOException {
```
chernser, line 446: This Class will be redesigned and API will be changed. 
Even it is in `internal` package please put a short javadoc for each exposed method 
```java 
/** 
   DO NOT USE - part of internal API that will be changed
*/

```

PR #2623, opened 2025-10-24
fix: allow duplicate column names in query responses

maxjustus:
Fixes client throwing errors like `Multiple entries with same key: number=1 and number=0` For queries with duplicate column names. IE: `select number, * from system.numbers limit 1;`. Aligns with `clickhouse client` which allows duplicate column names.

```
❯ clickhouse client --query "select number, * from system.numbers limit 1" --format JSONEachRow;                        08:33:45
{"number":0,"number":0}
```

### DIFF
--- a/client-v2/src/main/java/com/clickhouse/client/api/metadata/TableSchema.java
+++ b/client-v2/src/main/java/com/clickhouse/client/api/metadata/TableSchema.java
@@ -2,9 +2,9 @@ package com.clickhouse.client.api.metadata;
 
 import com.clickhouse.data.ClickHouseColumn;
 import com.google.common.collect.ImmutableList;
-import com.google.common.collect.ImmutableMap;
 
 import java.util.Collection;
+import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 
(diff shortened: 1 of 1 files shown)


client-v2/src/main/java/com/clickhouse/client/api/metadata/TableSchema.java
```diff
@@ -30,15 +30,17 @@ public TableSchema(String tableName, String query, String databaseName, Collecti
         this.databaseName = databaseName;
         this.query = query;
         this.columns = ImmutableList.copyOf(columns);
-        ImmutableMap.Builder<String, Integer> colIndexMapBuilder = ImmutableMap.builder();
+        Map<String, Integer> colIndexMapBuilder = new HashMap<>();
```
chernser, line 33: `HashMap` is much slower that `ImmutableMap` - that we have measured in our benchmark tests. 
`ImmutableMap` allows to leave last value.

chernser:
Good day, @maxjustus ! 

However we have fix for the problem in review https://github.com/ClickHouse/clickhouse-java/pull/2618

client-v2/src/main/java/com/clickhouse/client/api/metadata/TableSchema.java
maxjustus: got it!

PR #2627, opened 2025-10-29
[jdbc-v2] Convert Clickhouse `Array(T)` to Java `T[]` instead of `Object[]`

### LINKED ISSUE #2457: JDBC-V2: ResultSet inferred type for arrays is always `Object[]`
## Description

When retrieving a column, whose type is array, the inferred type is always `Object[]`; for example, it should be `Long[]` for `Array(Int64)`.

### Steps to reproduce

See code below.

See failing test below.

### Expected Behaviour

Test should pass.

### Code Example

```java
  @Test
  void required_testJdbcWithArray() throws Exception {
    final String url = "jdbc:ch://localhost:8123";

    final Properties info = new Properties();
    info.put("user", "default");
    info.put("password", "");

    try (final Connection conn = DriverManager.getConnection(url, info)) {
      conn.prepareStatement("CREATE TABLE test_array_table"
          + " (id Int64, array_field Array(Int64)) ENGINE = MergeTree() ORDER BY id").execute();
      conn.prepareStatement("INSERT INTO test_array_table VALUES (1, [1,2,3])").execute();
      try (final Statement tableStatement = conn.createStatement();
      final ResultSet resultSet = tableStatement.executeQuery("SELECT `array_field` FROM `test_array_table`")) {
        resultSet.next();
        final Array columnValue = resultSet.getArray(1);
        // The 2 assertions below should pass: instead we retrieve Object[]
        Assertions.assertThat(columnValue.getArray()).isExactlyInstanceOf(Long[].class);
        Assertions.assertThat(JDBCType.valueOf(columnValue.getBaseType())).isEqualTo(JDBCType.BIGINT);
      }
    }
  }

```

#### Client Configuration
See above.

#### Environment
* [ ] Cloud
* Client version:
* Language version:
* OS: 

#### ClickHouse Server
* ClickHouse Server version: 25.3.3.42

dolfinus:
## Summary

I've faced an issue when reading `Array(Decimal)` with clickhouse-jdbc 0.9.2 and Spark 3.5.7 - instead of converting values to Spark's `DecimalType`, it failed with exception:
```
Exception java.lang.ClassCastException: class [Ljava.lang.Object; cannot be cast to class [Ljava.math.BigDecimal; ([Ljava.lang.Object; and [Ljava.math.BigDecimal; are in module java.base of loader 'bootstrap')
```

This is because `JdbcUtils.convert(value, java.math.BigDecimal.class, column)` returns `Object[]`, and Java doesn't allow casts like those:
```java
(java.math.BigDecimal[]) new Object[]{};
```

Even if each item of array is actually a `BigDecimal`.

Now this method returns `BigDecimal[]` for these arguments. This requires some reflection usage to construct `T[]` with T known only in runtime.

I haven't run integration tests on this yet.

## Checklist
- [X] Closes #2457

### DIFF
--- a/jdbc-v2/src/main/java/com/clickhouse/jdbc/internal/JdbcUtils.java
+++ b/jdbc-v2/src/main/java/com/clickhouse/jdbc/internal/JdbcUtils.java
@@ -217,6 +217,27 @@ public class JdbcUtils {
         return DATA_TYPE_CLASS_MAP.get(clickhouseType);
     }
 
+    private static Class<?> unwrapPrimitiveType(Class<?> type) {
+        if (type == int.class) {
+            return Integer.class;
+        } else if (type == long.class) {
+            return Long.class;
+        } else if (type == boolean.class) {
+            return Boolean.class;
+        } else if (type == float.class) {
+            return Float.class;
+        } else if (type == double.class) {
+            return Double.class;
+        } else if (type == char.class) {
+            return Character.class;
+        } else if (type == byte.class) {
+            return Byte.class;
+        } else if (type == short.class) {
+            return Short.class;
+        }
+        return type;
+    }
+
     public static Object convert(Object value, Class<?> type) throws SQLException {
         return convert(value, type, null);
     }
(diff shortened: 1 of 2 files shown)


chernser:
Good day, @dolfinus ! 
Thank you for contributing! 

However `getObject()` by default should return `Array` object for all arrays. 
Defined in [8.9.3 JDBC Types Mapped to Java Object Types](https://docs.oracle.com/javase/1.5.0/docs/guide/jdbc/getstart/mapping.html)

There is a `java.sql.ResultSet#getObject(int, java.util.Map<java.lang.String,java.lang.Class<?>>)` that should be used in case user wants stripped array.

dolfinus:
`master` branch already contains clause:
```java
else if (type != java.sql.Array.class && value instanceof List<?>
```

The only thing changed there is that generic Object[] replaced with T[].
If PR title is misleading, I can update it.

chernser:
@dolfinus 
I'll review this week.

chernser (approved):

dolfinus:
Thanks!

chernser:
@dolfinus 
Please take a look https://github.com/ClickHouse/clickhouse-java/pull/2667 
This is related to the PR - tests for nested arrays were failing.

Issue #2527, opened 2025-08-07
Extraneous input 'ip' expecting


z0mb1ek:
Hi, i use latest client 0.9.1 and get trouble with word `ip`, is it reserved? I cannot find

```
val sqlParser: SqlParser = SqlParser()
    println(sqlParser.parsePreparedStatement("CREATE TABLE IF NOT EXISTS statistics.vast_event_v2\n" +
            "(\n" +
            "    ip             UInt32\n" +
            ") ENGINE = ReplacingMergeTree()"))
```
and get error:

```
[...]

PR #2632, opened 2025-10-30
Pre 0.9.3

chernser:
## Summary
Updates version and changelog 

Closes: https://github.com/ClickHouse/clickhouse-java/issues/2527

## Checklist
- [ ] Closes #<issue ref>

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,7 @@ To use it set `com.clickhouse.jdbc.DriverProperties#SQL_PARSER` to `ANTLR4`.
   - https://github.com/ClickHouse/clickhouse-java/issues/2572
   - https://github.com/ClickHouse/clickhouse-java/issues/2573
   - https://github.com/ClickHouse/clickhouse-java/issues/2609
+  - https://github.com/ClickHouse/clickhouse-java/issues/2527
 
 - [repo] New artifact `clickhouse-jdbc-all` added to address issue when maven package qualifiers may not be used. 
 This artifact should is a copy of `clickhouse-jdbc:all` but should be used only when required. (https://github.com/ClickHouse/clickhouse-java/issues/2625) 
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -1,3 +1,51 @@
+## 0.9.3
+
+### Important Changes 
+- [jdbc-v2] SQL parser from v1 is ported to v2 to address multiple issues with SQL parsing. The new parser is
```
mshustov:
```suggestion
- [jdbc-v2] SQL parser from v1 is ported to v2 to address multiple issues with SQL parsing. The ANTLR-based parser is 
```

mshustov: It would be beneficial to describe the difference between the two.

CHANGELOG.md
```diff
@@ -1,3 +1,51 @@
+## 0.9.3
+
+### Important Changes 
+- [jdbc-v2] SQL parser from v1 is ported to v2 to address multiple issues with SQL parsing. The new parser is 
+still an option and will be developed further. To use it set `com.clickhouse.jdbc.DriverProperties#SQL_PARSER` to `ANTLR4`.
```
mshustov: Don't we use both parsers for different purposes: to extract information and parse parameters? What level of granularity do we provide for configuring the parser per use case?

CHANGELOG.md
```diff
@@ -1,3 +1,51 @@
...
+  - https://github.com/ClickHouse/clickhouse-java/issues/2617
+  - https://github.com/ClickHouse/clickhouse-java/issues/2569
+  - https://github.com/ClickHouse/clickhouse-java/issues/2570
+  - https://github.com/ClickHouse/clickhouse-java/issues/2571
+  - https://github.com/ClickHouse/clickhouse-java/issues/2572
+  - https://github.com/ClickHouse/clickhouse-java/issues/2573
+  - https://github.com/ClickHouse/clickhouse-java/issues/2609
+
+- [repo] New artifact `clickhouse-jdbc-all` added to address issue when maven package qualifiers may not be used. 
+This artifact should is a copy of `clickhouse-jdbc:all` but should be used only when required. (https://github.com/ClickHouse/clickhouse-java/issues/2625)
```
laeg: > This artifact should is a copy
I think the should is extra here

> but should be used only when required.
We should explain in what scenarios this may be required

CHANGELOG.md
chernser: Parameters are parsed by separate function. There is an option to parse parameters by ANTLR4 but it requires its grammar to be full. Default option is to use Javacc to extract information and another function to parse parameters.

PR #2646, opened 2025-11-05
[Data-Type] Added QBit type constant. Not an implementation.

chernser:
## Summary
Adds `QBit` constant to `ClickhouseDataType` enum. 
There is a separate issue https://github.com/ClickHouse/clickhouse-java/issues/2610 to add new data type support. 

## Checklist
- [ ] Closes #<issue ref>


chernser:
There is a test that fails each time it detects a new type. 
Adding type is a big task - so we need to create an issue. 
Then we can add enum constant to indicate that we are aware of the new type.

For QBit we have issue #2610 
So I think we can add constant to unblock tests.

PR #2660, opened 2025-11-17
[client-v2] log duration in ISO-8601

janeklb:
Motivation:

The log messages produced by Client.java (v2) include a duration, but since it is in nanoseconds it is hard to read / interpret.

for example
```
Query request failed after attempts: 1 - Duration: 2784541
```

and `2784541` is `2ms`

Modification:
Log the ISO-8601 format of a duration (via `Duration#toString()`) when logging durations in Client.java

so above would be
```
Query request failed after attempts: 1 - Duration: PT0.002784541S
```

### DIFF
--- a/client-v2/src/main/java/com/clickhouse/client/api/Client.java
+++ b/client-v2/src/main/java/com/clickhouse/client/api/Client.java
@@ -1480,7 +1480,7 @@ public class Client implements AutoCloseable {
 
                     // Check response
                     if (httpResponse.getCode() == HttpStatus.SC_SERVICE_UNAVAILABLE) {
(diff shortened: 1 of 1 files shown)


janeklb:
I see there is a `StopWatch.java` -- i could use this instead if you think it would be better (although that seems to be meant specifically for metrics as it is
```java
public class StopWatch implements Metric {
```

chernser:
Thank you, @janeklb ! 
The PR looks good.

chernser:
As for `StopWatch` I would wait. It was added to cover very simple case. 
We need to review metrics parts and build a complete observability solution.

PR #2663, opened 2025-11-20
[jdbc-v2] Support multi-dot notation for database names

### LINKED ISSUE #2650: [jdbc-v2] Support dot notation without quotes
### Description 

There is potentially a feature in ClickHouse when 'a.b.c' where 'a.b' is a DB name become valid without quotes. 
Today JDBC driver accepts ony 'a.b' where `a` is DB and `b` is a table.

devdavidkarlsson:
## Description
Fixes #2650 

This PR adds support for multi-dot notation in database names without requiring quotes, enabling table references like `a.b.c` where `a.b` is the database name and `c` is the table name.

## Changes
- **Grammar**: Updated ANTLR `databaseIdentifier` rule to support multiple dot-separated identifiers: `identifier (DOT identifier)*`
- **Parser Logic**: Added `extractTableName()` helper method in `SqlParserFacade` to properly unquote each identifier part separately and join them correctly
- **Tests**: Added `testMultiDotNotation()` test method with 3 test cases:
  - Unquoted multi-dot: `SELECT * FROM a.b.c WHERE id = ?`
  - Quoted identifiers: ``SELECT * FROM `db.part1`.`table` WHERE id = ?``
  - INSERT statement: `INSERT INTO a.b.c (col1, col2) VALUES (?, ?)`

## Testing
- ✅ All 367 existing tests continue to pass
- ✅ New test validates parsing of multi-dot database names
- ✅ Verified against ClickHouse server with databases containing dots in their names

## Example Usage
```sql
-- Database with dot in name (must be quoted when creating)
CREATE DATABASE `test.db`;
CREATE TABLE `test.db`.users (id UInt32, name String) ENGINE = MergeTree ORDER BY id;

-- Can now be referenced in JDBC with proper parsing
SELECT * FROM `test.db`.users WHERE id = ?;
INSERT INTO `test.db`.users VALUES (?, ?);
```

The JDBC driver now correctly parses these statements and extracts the full table name `test.db.users` for metadata operations.


jdbc-v2/src/main/antlr4/com/clickhouse/jdbc/internal/parser/antlr4/ClickHouseParser.g4
```diff
@@ -1172,7 +1172,7 @@ tableArgExpr
 // Databases
 
 databaseIdentifier
-    : identifier
+    : identifier (DOT identifier)*
```
chernser, line 1176: we need similar fix in the JavaCC grammar.

jdbc-v2/src/main/java/com/clickhouse/jdbc/internal/SqlParserFacade.java
```diff
@@ -265,18 +265,51 @@ public void enterAssignmentValuesList(ClickHouseParser.AssignmentValuesListConte
             }
 
 
+            private String extractTableName(ClickHouseParser.TableIdentifierContext tableId) {
```
chernser: I think we can make this in parser.

chernser:
Good day, @devdavidkarlsson ! 

I've left comments. 
Build fails with: 

```
Error:  Tests run: 1182, Failures: 1, Errors: 0, Skipped: 0, Time elapsed: 3.501 s <<< FAILURE! -- in TestSuite
Error:  com.clickhouse.jdbc.internal.JavaCCParserTest.testMultiDotNotation -- Time elapsed: 0.010 s <<< FAILURE!
java.lang.AssertionError: expected [false] but found [true]
	at org.testng.Assert.fail(Assert.java:110)
	at org.testng.Assert.failNotEquals(Assert.java:1413)
	at org.testng.Assert.assertFalse(Assert.java:78)
	at org.testng.Assert.assertFalse(Assert.java:88)
	at com.clickhouse.jdbc.internal.BaseSqlParserFacadeTest.testMultiDotNotation(BaseSqlParserFacadeTest.java:169)
```

devdavidkarlsson:
@chernser new attempt: I added the JavaCC (.jj) and got rid of extractTableName, tests should be green now.
Thank you 🙏

jdbc-v2/src/main/java/com/clickhouse/jdbc/internal/SqlParserFacade.java
```diff
@@ -248,6 +249,56 @@ public void enterColumnExprPrecedence3(ClickHouseParser.ColumnExprPrecedence3Con
...
+                    } else if (inQuote && ch == quoteChar) {
+                        // Check for escaped quote (doubled quote)
+                        if (i + 1 < rawTableId.length() && rawTableId.charAt(i + 1) == quoteChar) {
+                            currentPart.append(ch).append(ch);
+                            i++; // Skip the next quote
+                        } else {
+                            inQuote = false;
+                            currentPart.append(ch);
+                        }
+                    } else if (!inQuote && ch == '.') {
```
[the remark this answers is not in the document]
devdavidkarlsson: suggested correction (ch == '.' instead of ch == '.') is identical to what is already in the code.

chernser:
Good day, @devdavidkarlsson ! 
Will review today.

jdbc-v2/src/main/java/com/clickhouse/jdbc/internal/SqlParserFacade.java
```diff
@@ -248,6 +249,56 @@ public void enterColumnExprPrecedence3(ClickHouseParser.ColumnExprPrecedence3Con
                 super.enterColumnExprPrecedence3(ctx);
             }
 
+            private String unquoteTableIdentifier(String rawTableId) {
```
chernser: strange - I thought I've left the comment that it should be actually handle by parser. 
Because one of the points to have parser is to avoid writing this kind of methods.

jdbc-v2/src/main/javacc/ClickHouseSqlParser.jj
```diff
@@ -903,18 +903,32 @@ void nestedIdentifier(): {} {
...
         (LOOKAHEAD(2)
             <LPAREN> { token_source.addCustomKeywordPosition(ClickHouseSqlStatement.KEYWORD_TABLE_COLUMNS_START, token); }
             anyExprList()
             <RPAREN> { token_source.addCustomKeywordPosition(ClickHouseSqlStatement.KEYWORD_TABLE_COLUMNS_END, token); }
         )?
     )
     {
-        if (record && t != null && token_source.table == null) {
-            token_source.table = ClickHouseSqlUtils.unescape(t.image);
+        if (record && token_source.table == null && parts.size() > 0) {
```
chernser: why parser cannot do it for us?

jdbc-v2/src/main/java/com/clickhouse/jdbc/internal/SqlParserFacade.java
devdavidkarlsson: Thanks. The logic you commented on has been moved into the parsers themselves. In particular, the unquoteTableIdentifier helper was removed, and both the JavaCC and ANTLR grammars were updated to handle quoted identifiers directly (commit 95cce2ae).

In the current code, the listener just consumes the parser structure and applies unescape() to individual identifier tokens, following the same pattern used in enterUseStmt (i.e. getText() + utility), which I hope aligns with the existing style.

jdbc-v2/src/main/javacc/ClickHouseSqlParser.jj
devdavidkarlsson: Refactored JavaCC to use grammar structure. That is: instead of collecting parts into a List with post-processing, added databasePart/tablePart helper rules (I think consistent with existing *Part rules like formatPart, compressionPart). The grammar now determines what's database vs table via LOOKAHEAD.

chernser:
Good day, @devdavidkarlsson ! 

I appreciate your effort! 

Code now looks much more simple - what is great! 
I've run tests and if they pass - will merge.

chernser:
@devdavidkarlsson 
I see there is a failure because we need to fix one test. I hopefully do it today and will merge the PR.

Issue #1545, opened 2024-02-17
"getResultSet not implemented" error when mapping Array(Nullable(UInt32)) after updating from 0.3.2-patch11


gaabs:
### Describe the bug
Upon trying to update from `0.3.2-patch11` to `0.6.0-patch1:http`, some behavior changes were noticed regarding the mapping of `Array(Nullable(UInt32))` data type when using JDBI. The previous code used to work correctly, but it started to throw a `getResultSet not implemented` after updating the clickhouse dependency. It seems to have started at the `0.4.0` version.

Upon further investigation, it was observed that `Array(Nullable(UInt32))` is now being mapped to an array of `UnsignedInteger`, instead of to `Long`. That can be seen inspecting [JDBI's `buildArray` method](https://github.com/jdbi/jdbi/blob/v3.44.1/core/src/main/java/org/jdbi/v3/core/array/ArrayColumnMapper.java#L56). That ends up trying to use the resultSet, which [is not implemented in the clickhouse dependency](https://github.com/ClickHouse/clickhouse-java/blob/v0.6.0-patch1/clickhouse-jdbc/src/main/java/com/clickhouse/jdbc/ClickHouseArray.java#L73). 

It was also noticed that`Nullable(UInt32)` is still being mapped to `Long`, the same as before.

Is that behavior change expected?

### Steps to reproduce
1. Insert `Array(Nullable(UInt32))` data into some clickhouse table
2. Retrieve `Array(Nullable(UInt32))` data using JDBI

### Expected behavior

`Array(Nullable(UInt32))` should be mapped to `Long[]`.

### Code example

> build.gradle with used dependencies
```groovy
plugins {
    id("io.micronaut.application") version "4.2.1"
}

version = "0.1"
group = "com.example"

repositories {
    mavenCentral()
}

ext {
    micronautVersion = '4.2.3'
}

dependencies {
    annotationProcessor "org.projectlombok:lombok:1.18.30"
    compileOnly "org.projectlombok:lombok:1.18.30"

    implementation "org.jdbi:jdbi3-sqlobject:3.44.0"
    implementation "com.clickhouse:clickhouse-jdbc:0.6.0-patch1:http"
    implementation 'org.apache.httpcomponents.client5:httpclient5:5.3.1'

    runtimeOnly "io.micronaut.sql:micronaut-jdbc-hikari"
    runtimeOnly "org.yaml:snakeyaml"
}

java {
[...]

Issue #1438, opened 2023-08-31
getResultSet not implemented


adrevant:
### Describe the bug
I'm trying to send some data from a couple of tables in my Clickhouse database to big query by using Airbyte but I'm having this error:

```
java.sql.SQLFeatureNotSupportedException: getResultSet not implemented
```

that is shown because indeed clickhouse-jdbc didn't implement the function `getResultSet` in [ClickHouseArray.java](https://github.com/ClickHouse/clickhouse-java/blob/main/clickhouse-jdbc/src/main/java/com/clickhouse/jdbc/ClickHouseArray.java). 

### Error log
```
INFO i.a.w.g.ReplicationWorkerHelper(getReplicationOutput):313 - failures: [ {
  "failureOrigin" : "source",
  "failureType" : "system_error",
  "internalMessage" : "java.lang.RuntimeException: java.lang.RuntimeException: java.sql.SQLFeatureNotSupportedException: getResultSet not implemented",
  "externalMessage" : "Something went wrong in the connector. See the logs for more details.",
  "metadata" : {
    "attemptNumber" : 2,
    "jobId" : 5708,
    "from_trace_message" : true,
    "connector_command" : "read"
  },
```

#### Environment
* Client version: 23.5.3.24
* Language version: 11.0.19
* OS: Debian GNU/Linux 11 (bullseye)

Is there a plan for implementing it?

zhicwu:
Hi @brendavarguez, thanks for reporting the issue. Will come up with a PR to fix this in weekend.

Meanwhile, if you have control of the code, it's better to use `ResultSet.getObject()` to retrieve the value. Alternatively, you may play with the `typeMappings` option to see if you can find a different workaround.

mzitnik:
@zhicwu what is the status of this PR?

uttamdeb:
Hey, team. Recently faced the same issue, hence not eleborating any further. Any timeline on this, since it has been here for a while?

PR #2668, opened 2025-11-25
Jdbc impl get result set in array

chernser:
## Summary
- Adds implementation for `Array.getResultSet()` and `Array.getResultSet(index, count` to improve GUI tools UX. 
- No implementation of `com.clickhouse.jdbc.types.Array#getArray(long, int, java.util.Map<java.lang.String,java.lang.Class<?>>)` because it involves more complex mapping and not widely used. Will do later.
- Adds converters for base types and implements new `ArrayresultSet` with it. This way data conversion relies on value class knowledge and much faster than series of `instanceof` calls. 

Closes https://github.com/ClickHouse/clickhouse-java/issues/1545
Closes https://github.com/ClickHouse/clickhouse-java/issues/2683
Closes https://github.com/ClickHouse/clickhouse-java/issues/1438

## Checklist
- [x] Closes #<issue ref>

### DIFF
--- a/jdbc-v2/src/main/java/com/clickhouse/jdbc/types/ArrayResultSet.java
+++ b/jdbc-v2/src/main/java/com/clickhouse/jdbc/types/ArrayResultSet.java
@@ -39,7 +39,7 @@ public class ArrayResultSet implements ResultSet {
 
     private final Object array;
     private final int length;
-    private Integer pos;
+    private int pos;
     private boolean closed;
     private ResultSetMetaDataImpl metadata;
 
@@ -184,10 +184,10 @@ public class ArrayResultSet implements ResultSet {
     @Override
     public byte getByte(int columnIndex) throws SQLException {
         if (columnIndex == 1) {
-            if (pos < Byte.MAX_VALUE) {
+            if (pos + 1 < Byte.MAX_VALUE) {
                 return (byte) getRow();
             } else {
-                throw new SQLException("INDEX column value too big and cannot be get as byte");
+                throw new SQLException("INDEX column value too big and cannot be retrieved as byte");
             }
         }
         return ((Number) getValueAsObject(columnIndex, Byte.class, 0)).byteValue();
@@ -196,10 +196,10 @@ public class ArrayResultSet implements ResultSet {
     @Override
     public short getShort(int columnIndex) throws SQLException {
         if (columnIndex == 1) {
-            if (pos < Short.MAX_VALUE) {
+            if (pos + 1 < Short.MAX_VALUE) {
                 return (short) getRow();
             } else {
-                throw new SQLException("INDEX column value too big and cannot be get as short");
+                throw new SQLException("INDEX column value too big and cannot be retrieved as short");
             }
         }
         return ((Number) getValueAsObject(columnIndex, Short.class, 0)).shortValue();


client-v2/src/main/java/com/clickhouse/client/api/data_formats/internal/ValueConverters.java
```diff
@@ -0,0 +1,251 @@
...
+    public BigInteger convertNumberToBigInteger(Object value) {
+        return BigInteger.valueOf(((Number) value).longValue());
+    }
+
+    public BigDecimal convertNumberToBigDecimal(Object value) {
+        return BigDecimal.valueOf(((Number) value).doubleValue());
+    }
+
+    // Date & Time converters
+    public java.sql.Date conveSqlDateToSqlDate(Object value) {
```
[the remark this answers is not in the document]
chernser: This is your auto-complete :-)

client-v2/src/main/java/com/clickhouse/client/api/data_formats/internal/ValueConverters.java
```diff
@@ -0,0 +1,251 @@
...
+public final class ValueConverters {
+
+
+    // <source type, <target type, converter>>
+    private final Map<Class<?>, Map<Class<?>, Function<Object, Object>>> classConverters;
+
+    // <target type, converter>
+    private final ImmutableMap<Class<?>, Function<Object, Object>> numberConverters;
+    
+    public ValueConverters() {
```
mzitnik, line 25: Looking into the Java driver source code, it appears that we initialize `ValueConverters` for every `ArrayResultSet`, and I believe this can be overhead due to recreating the map repeatedly. Can we init it once?

client-v2/src/main/java/com/clickhouse/client/api/data_formats/internal/ValueConverters.java
```diff
@@ -0,0 +1,251 @@
...
+    // String to any
+    public String convertStringToString(Object value) {
+        return (String) value;
+    }
+
+    public byte[] convertStringToBytes(Object value) {
+        return ((String) value).getBytes();
+    }
+
+    public boolean convertStringToBoolean(Object value) {
```
mzitnik, line 139: `Parse` methods throw Exceptions. Can you point me to the location where we are handling it?

jdbc-v2/src/test/java/com/clickhouse/jdbc/DataTypeTests.java
```diff
@@ -835,11 +835,11 @@ public void testFloatTypes() throws SQLException {
             try (Statement stmt = conn.createStatement()) {
                 try (ResultSet rs = stmt.executeQuery("SELECT * FROM test_floats ORDER BY order")) {
                     assertTrue(rs.next());
-                    assertEquals(rs.getObject("float32"), -3.402823E38d);
+                    assertEquals(rs.getObject("float32"), -3.402823E38f);
```
mzitnik, line 838: What was the changed  `-3.402823E38f` vs `-3.402823E38d`

mzitnik:
ValueConverters, according to the coverage test, have zero coverage. Is it real?
```
com.clickhouse.client.api.data_formats.internal.ValueConverters | 0.00% |   | 109
-- | -- | -- | --
```

chernser:
@mzitnik coverage report is incorrect because tests for this class really in jdbc 
`ValueConverters` should be covered by 95% according to IDE run

jdbc-v2/src/test/java/com/clickhouse/jdbc/DataTypeTests.java
chernser: previously columns of `Float32` were returned as `Double` what is not correct. 
I've found it while testing converters for primitive types and fixed it. 
Here I've changed expected value from `Double` to `Float`

client-v2/src/main/java/com/clickhouse/client/api/data_formats/internal/ValueConverters.java
chernser: Good catch! Yes I will create it as static field. In the future we will pass it as parameter may be.

client-v2/src/main/java/com/clickhouse/client/api/data_formats/internal/ValueConverters.java
chernser: Got it! They will be thrown till user code - so I've added wrapping try-catch in `com.clickhouse.jdbc.types.ArrayResultSet#getValueAsObject` where converters are called.

jdbc-v2/src/test/java/com/clickhouse/jdbc/types/ArrayResultSetTest.java
```diff
@@ -0,0 +1,561 @@
...
+        ArrayResultSet rs = new ArrayResultSet(array, ClickHouseColumn.parse("v Array(Int32)").get(0));
+
+        rs.next();
+        assertEquals(rs.getInt(2), 0);
+        assertTrue(rs.wasNull());
+        assertTrue(rs.wasNull());
+
+        rs.next();
+        assertTrue(rs.wasNull());
+        assertEquals(rs.getInt(2), array[1]);
```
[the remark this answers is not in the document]
chernser: this is to check flag between transition of states.

mzitnik (approved):
let's open an issue to test it with AirByte @serebrserg

chernser:
Thank you, @mzitnik ! 
The issue is https://github.com/ClickHouse/clickhouse-java/issues/2686

PR #2713, opened 2026-01-15
[client-v2] Send parameters in request body 

chernser:
## Summary
- implements sending parameters in body request (only for statement parameters) 
- adds new configuration property `USE_HTTP_FORM_REQUEST_FOR_QUERY` 

Closes https://github.com/ClickHouse/clickhouse-java/issues/2324
## Checklist
- [ ] Closes #<issue ref>

---

> [!NOTE]
> Introduces optional multipart form requests for statement parameters and aligns HTTP/compression handling.
> 
> - Adds config `client.http.use_form_request_for_query` and builder `useHttpFormDataForQuery(...)`
> - `Client.query(...)`: when enabled and params present, sends multipart; otherwise uses standard POST; inserts/query paths updated to new helper APIs
> - `HttpAPIClientHelper`: new `executeRequest(..., String)` and `executeMultiPartRequest(...)`; refactors URI/context/header builders; wraps entities without passing `LZ4Factory` per-call; disables client compression for multipart
> - `CompressedEntity`/`LZ4Entity`: correct unknown content length for compressed requests and improve stream lifecycle
> - Tests: adjust default config size, add `testMultiPartRequest` integration test
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit b7eba3698f173853bf3ec375d66fd875aed56bde. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/client-v2/src/main/java/com/clickhouse/client/api/internal/HttpAPIClientHelper.java
+++ b/client-v2/src/main/java/com/clickhouse/client/api/internal/HttpAPIClientHelper.java
@@ -465,9 +465,6 @@ public class HttpAPIClientHelper {
     public ClassicHttpResponse executeRequest(Endpoint server, Map<String, Object> requestConfig,
                                               String body) throws Exception {
 
-        if (requestConfig == null) {
-            requestConfig = Collections.emptyMap();
-        }
         final URI uri = createRequestURI(server, requestConfig, true);
         final HttpPost req = createPostRequest(uri, requestConfig);
         final String contentEncoding = req.containsHeader(HttpHeaders.CONTENT_ENCODING) ? req.getHeader(HttpHeaders.CONTENT_ENCODING).getValue() : null;
@@ -480,9 +477,6 @@ public class HttpAPIClientHelper {
 
     public ClassicHttpResponse executeMultiPartRequest(Endpoint server, Map<String, Object> requestConfig, String sqlQuery) throws Exception {
 
-        if (requestConfig == null) {
-            requestConfig = Collections.emptyMap();
-        }
         requestConfig.put(ClientConfigProperties.COMPRESS_CLIENT_REQUEST.getKey(), false);
 
         final URI uri = createRequestURI(server, requestConfig, false);
@@ -505,9 +499,6 @@ public class HttpAPIClientHelper {
     public ClassicHttpResponse executeRequest(Endpoint server, Map<String, Object> requestConfig,
                                               IOCallback<OutputStream> writeCallback) throws Exception {
 
-        if (requestConfig == null) {
-            requestConfig = Collections.emptyMap();
-        }
         final URI uri = createRequestURI(server, requestConfig, true);
         final HttpPost req = createPostRequest(uri, requestConfig);
         String contentEncoding = req.containsHeader(HttpHeaders.CONTENT_ENCODING) ? req.getHeader(HttpHeaders.CONTENT_ENCODING).getValue() : null;


client-v2/src/main/java/com/clickhouse/client/api/ClientConfigProperties.java
```diff
@@ -182,6 +182,13 @@ public Object parseValue(String value) {
      * SNI SSL parameter that will be set for each outbound SSL socket.
      */
     SSL_SOCKET_SNI("ssl_socket_sni", String.class,""),
+
+    /**
+     * If parameters should be sent in request body.
+     * Note: work only with HTTP Compression
+     */
+    USE_HTTP_FORM_REQUEST_FOR_QUERY("client.http.use_form_request_for_query", Boolean.class, "false"),
```
kavirajk: suggestion: Can we name the config `HTTP_SEND_PARAMS_IN_BODY` instead? Not sure if there are any standard way to name these in java client. But just looking at it, made me confuse. Mainly `FOR_QUERY` is too generic and no where it says about `parameters`?.

or `USE_HTTP_FORM_REQUEST_FOR_QUERY_PARAMS` but I felt that is bit too mouthful. Given it's a `boolean` we can avoid `USE_` prefix I think.

client-v2/src/main/java/com/clickhouse/client/api/internal/CompressedEntity.java
```diff
@@ -75,7 +75,8 @@ public void close() throws IOException {
 
     @Override
     public long getContentLength() {
-        return httpEntity.getContentLength();
+        // compressed request length is unknown event if it is a byte[]
```
kavirajk: I think it should be `even` not `event`?

client-v2/src/main/java/com/clickhouse/client/api/ClientConfigProperties.java
```diff
@@ -182,6 +182,13 @@ public Object parseValue(String value) {
      * SNI SSL parameter that will be set for each outbound SSL socket.
      */
     SSL_SOCKET_SNI("ssl_socket_sni", String.class,""),
+
+    /**
+     * If parameters should be sent in request body.
+     * Note: work only with HTTP Compression
```
kavirajk: does it works with both HTTP compression and block level compression?

client-v2/src/test/java/com/clickhouse/client/HttpTransportTests.java
```diff
@@ -1185,6 +1188,44 @@ public void testEndpointUrlPathIsPreserved() throws Exception {
...
+        // compress request
+        try (Client client = newClient()
+                .compressClientRequest(true)
+                .setOption(ClientConfigProperties.USE_HTTP_FORM_REQUEST_FOR_QUERY.getKey(), "true")
+                .useHttpCompression(true).build()) {
+            client.queryAll("SELECT database, name FROM system.tables WHERE name IN {table_names:Array(String)}",
+                    params);
+            fail("exception expected");
+        } catch (Exception e) {
+            e.printStackTrace();
```
kavirajk: Not sure if this is used for debugging. and should be replaced with throw instead of print?

chernser:
Thank you, @kavirajk ! 

I appreciate your review!

client-v2/src/main/java/com/clickhouse/client/api/ClientConfigProperties.java
chernser: It works for server side compression only. I will rephrase the comment. 
It doesn't work with client side compression because requires to compress each parameter what would be expensive.

client-v2/src/main/java/com/clickhouse/client/api/ClientConfigProperties.java
chernser: Sounds good! 
I  struggle to find good names sometimes.

client-v2/src/test/java/com/clickhouse/client/HttpTransportTests.java
chernser: yes, in general it should be logged or thrown. Using `printStackTrace` is not good. Even in tests we have `org.testng.Assert#fail(java.lang.String, java.lang.Throwable)`.
I've changed code so this test branch is positive now.

client-v2/src/main/java/com/clickhouse/client/api/internal/HttpAPIClientHelper.java
```diff
@@ -410,51 +418,117 @@ public Exception readError(ClassicHttpResponse httpResponse) {
...
+
+        return doPostRequest(requestConfig, req);
+    }
+
+    public ClassicHttpResponse executeMultiPartRequest(Endpoint server, Map<String, Object> requestConfig, String sqlQuery) throws Exception {
+
+        if (requestConfig == null) {
+            requestConfig = Collections.emptyMap();
+        }
+        requestConfig.put(ClientConfigProperties.COMPRESS_CLIENT_REQUEST.getKey(), false);
```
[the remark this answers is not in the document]
chernser: Agree. Will fix.

PR #2740, opened 2026-02-09
Accept 1/0 for boolean configuration values

### LINKED ISSUE #2730: [jdbc-v2, client-v2] Accept `1` and `0` for boolean properties
### Description  

There is a number of boolean properties in the driver and the client. Currently they may be inconsistent and accept `1` along with `true`. For better user experience we need it consistent and documented.

kailash360:
## Summary

This PR standardizes boolean parsing across modules so configuration values accept both `true`/`false` and `1`/`0` (with `null`/unrecognized values treated as false). This improves compatibility with tools and environments that only emit numeric booleans.

Closes 
## Checklist
- [x] Closes #2730 
- [x] For significant changes, documentation in https://github.com/ClickHouse/clickhouse-docs was updated with further explanations or tutorials

---

> [!NOTE]
> **Medium Risk**
> Behavior changes for boolean parsing in config and value conversion paths could alter outcomes for previously non-standard string values, though the change is small and well-covered by new tests.
> 
> **Overview**
> Standardizes boolean parsing across `clickhouse-data`, `client-v2`, and `jdbc-v2` by introducing `ClickHouseUtils.parseBoolean(String)` (supports `true`/`false` and `1`/`0`, defaults to `false` for null/unknown) and routing existing boolean parsing through it.
> 
> Updates client config parsing (`ClientConfigProperties.parseValue`), string-to-boolean conversions in serialization/value conversion helpers (`SerializerUtils`, `ValueConverters`), and JDBC driver flags (`JdbcConfiguration`) to use the new parser, and adds targeted unit tests covering numeric and textual boolean inputs.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 3128606ee63c0b707810266ca6fb542bca506896. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/clickhouse-data/src/main/java/com/clickhouse/data/ClickHouseUtils.java
+++ b/clickhouse-data/src/main/java/com/clickhouse/data/ClickHouseUtils.java
@@ -85,6 +85,25 @@ public final class ClickHouseUtils {
     public static final String VARIABLE_PREFIX = "{{";
     public static final String VARIABLE_SUFFIX = "}}";
 
+    /**
+     * Parses a boolean value from string. Accepts "true"/"false" and "1"/"0".
+     *
+     * @param value string value
+     * @return parsed boolean, or {@code false} if value is null or unrecognized
+     */
+    public static boolean parseBoolean(String value) {
+        if (value == null) {
+            return false;
+        }
+        if ("1".equals(value)) {
+            return true;
+        }
+        if ("0".equals(value)) {
+            return false;
+        }
+        return Boolean.parseBoolean(value);
+    }
+
     private static <T> T findFirstService(Class<? extends T> serviceInterface) {
         ClickHouseChecker.nonNull(serviceInterface, "serviceInterface");
 
--- a/clickhouse-data/src/test/java/com/clickhouse/data/ClickHouseUtilsTest.java
+++ b/clickhouse-data/src/test/java/com/clickhouse/data/ClickHouseUtilsTest.java
@@ -21,6 +21,20 @@ import org.testng.annotations.Test;
 
 public class ClickHouseUtilsTest {
     @Test(groups = { "unit" })
+    public void testParseBoolean() {
+        Assert.assertFalse(ClickHouseUtils.parseBoolean(null));
+
+        Assert.assertTrue(ClickHouseUtils.parseBoolean("1"));
+        Assert.assertFalse(ClickHouseUtils.parseBoolean("0"));
+
+        Assert.assertTrue(ClickHouseUtils.parseBoolean("true"));
+        Assert.assertFalse(ClickHouseUtils.parseBoolean("false"));
+
+        // unrecognized values should be parsed as false
+        Assert.assertFalse(ClickHouseUtils.parseBoolean("yes"));
+    }
+
+    @Test(groups = { "unit" })
     public void testCreateTempFile() throws IOException {
         File f = ClickHouseUtils.createTempFile(null, null);
         Assert.assertNotNull(f);
--- a/client-v2/src/main/java/com/clickhouse/client/api/ClientConfigProperties.java
+++ b/client-v2/src/main/java/com/clickhouse/client/api/ClientConfigProperties.java
@@ -9,6 +9,7 @@ import org.slf4j.LoggerFactory;
 
 import java.util.ArrayList;
 import java.util.Arrays;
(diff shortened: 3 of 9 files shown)


clickhouse-data/src/test/java/com/clickhouse/data/ClickHouseUtilsTest.java
```diff
@@ -20,6 +20,20 @@
...
+    public void testParseBoolean() {
+        Assert.assertFalse(ClickHouseUtils.parseBoolean(null));
+
+        Assert.assertTrue(ClickHouseUtils.parseBoolean("1"));
+        Assert.assertFalse(ClickHouseUtils.parseBoolean("0"));
+
+        Assert.assertTrue(ClickHouseUtils.parseBoolean("true"));
+        Assert.assertFalse(ClickHouseUtils.parseBoolean("false"));
+
+        // unrecognized values should be parsed as false
```
mshustov, line 33: @chernser shouldn't we throw an exception for unrecognized values?

kailash360: If there is an unrecognized value, then we should also follow the behavior pattern of other configs in this case.

For example, the `default_query_settings` config accepts a set of settings, else `null` by default. Now, if someone sets a random setting that is not used by Clickhouse then does it throw an error, or simply ignores that? 

Similarly, `jdbc_sql_parser` accepts 1 of 3 enum constants. Does it throw an error if any other value is passed?

If we throw an exception only for `boolean` types and simply ignore for other types, then it will create inconsistent behaviour. Please let me know your thoughts on this.

chernser:
@kailash360 
thank you for the contribution - I will review PR this week.

chernser:
@kailash360 
There are some compilation errors 
```
Error:  /home/runner/work/clickhouse-java/clickhouse-java/client-v2/src/test/java/com/clickhouse/client/api/ClientConfigPropertiesTest.java:[24,1] class, interface, or enum expected
Error:  /home/runner/work/clickhouse-java/clickhouse-java/client-v2/src/test/java/com/clickhouse/client/api/ClientConfigPropertiesTest.java:[27,1] class, interface, or enum expected
Error:  /home/runner/work/clickhouse-java/clickhouse-java/client-v2/src/test/java/com/clickhouse/client/api/ClientConfigPropertiesTest.java:[28,1] class, interface, or enum expected
Error:  /home/runner/work/clickhouse-java/clickhouse-java/client-v2/src/test/java/com/clickhouse/client/api/ClientConfigPropertiesTest.java:[30,1] class, interface, or enum expected
``` 
We currently still supporting java 8.

clickhouse-data/src/main/java/com/clickhouse/data/ClickHouseUtils.java
```diff
@@ -85,6 +85,25 @@ public final class ClickHouseUtils {
     public static final String VARIABLE_PREFIX = "{{";
     public static final String VARIABLE_SUFFIX = "}}";
 
+    /**
+     * Parses a boolean value from string. Accepts "true"/"false" and "1"/"0".
+     *
+     * @param value string value
+     * @return parsed boolean, or {@code false} if value is null or unrecognized
+     */
+    public static boolean parseBoolean(String value) {
```
chernser, line 94: Please move this method to client-v2 
I suggest utilizing ValueConverters.

chernser: PS: clickhouse-data is an old package that we may deprecate so V2 should minimize dependencies on it. Besides clickhouse-data relates to client-v1 and this method is not used there.

clickhouse-data/src/test/java/com/clickhouse/data/ClickHouseUtilsTest.java
chernser: @kailash360 both mentioned settings are not validated but should. 
`jdbc_sql_parser` will throw exception later while creating connection. That is not correct and we will fix it. 
Most Client configurations validated in the Builder.

Please add throwing exception if boolean value is not what we expect. 

Thanks!

chernser (changes requested):
Please apply suggested changes. 
Thanks!

Issue #2347, opened 2025-05-01
[client-v2] Support returning raw http header


chernser:
### Description

There are some cases when we need to get all http from a server response.
Currently only a few decoded.

PR #2744, opened 2026-02-13
Expose X-ClickHouse-Server-Display-Name header via client-v2 API

synhershko:
Extract the server display name from HTTP response headers and surface it through OperationMetrics, QueryResponse, InsertResponse, and CommandResponse, following the existing queryId pattern.

Closes https://github.com/ClickHouse/clickhouse-java/issues/2347

---

> [!NOTE]
> **Low Risk**
> Additive API surface that only reads and exposes a small, whitelisted set of response headers; minimal impact on request/response flow aside from extra header collection.
> 
> **Overview**
> Client-v2 now captures a **whitelisted set of ClickHouse HTTP response headers** and attaches them to operation results.
> 
> `Client` passes collected headers into `QueryResponse` and `InsertResponse`, while `QueryResponse`, `InsertResponse`, and `CommandResponse` expose `getServerDisplayName()` (from `X-ClickHouse-Server-Display-Name`) plus `getResponseHeaders()` for the full collected map. `HttpAPIClientHelper` adds `collectResponseHeaders()` with a fixed whitelist to avoid exposing arbitrary headers.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 89328758f6069cf09a94766876b9abb3aadea871. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/client-v2/src/main/java/com/clickhouse/client/api/Client.java
+++ b/client-v2/src/main/java/com/clickhouse/client/api/Client.java
@@ -1338,9 +1338,7 @@ public class Client implements AutoCloseable {
                     String queryId =  HttpAPIClientHelper.getHeaderVal(httpResponse.getFirstHeader(ClickHouseHttpProto.HEADER_QUERY_ID), requestSettings.getQueryId(), String::valueOf);
                     metrics.operationComplete();
                     metrics.setQueryId(queryId);
-                    metrics.setServerDisplayName(HttpAPIClientHelper.getHeaderVal(
-                            httpResponse.getFirstHeader(ClickHouseHttpProto.HEADER_SRV_DISPLAY_NAME), null));
-                    return new InsertResponse(metrics);
+                    return new InsertResponse(metrics, HttpAPIClientHelper.collectResponseHeaders(httpResponse));
                 } catch (Exception e) {
                     String msg = requestExMsg("Insert", (i + 1), durationSince(startTime).toMillis(), requestSettings.getQueryId());
                     lastException = httpClientHelper.wrapException(msg, e, requestSettings.getQueryId());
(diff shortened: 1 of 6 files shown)


client-v2/src/main/java/com/clickhouse/client/api/Client.java
```diff
@@ -1338,6 +1338,8 @@ public CompletableFuture<InsertResponse> insert(String tableName, List<?> data,
                     String queryId =  HttpAPIClientHelper.getHeaderVal(httpResponse.getFirstHeader(ClickHouseHttpProto.HEADER_QUERY_ID), requestSettings.getQueryId(), String::valueOf);
                     metrics.operationComplete();
                     metrics.setQueryId(queryId);
+                    metrics.setServerDisplayName(HttpAPIClientHelper.getHeaderVal(
```
chernser: @synhershko 
Metrics is not a good place for headers: 
- header is not a metrics
- we will rework metrics and removing header from it will be problematic
- native protocol may have another way to pass this information. 

I suggest: 
- create `Map<String, String> collectResponseHeaders(HttpResponse)` in `HttpAPIClientHelper`
- add `Map<String, String> responseHeaders` to insert and query response classes 
- it can be a getters for significant headers or a getter method returning all headers. 

Thank you!

chernser (changes requested):

chernser:
Good day, @synhershko ! 
Please see my comment. 

Thanks!

client-v2/src/main/java/com/clickhouse/client/api/Client.java
synhershko: Thanks for the feedback! I followed the queryId pattern (also being retrieved from headers). Should I move both to `responseHeaders`?

chernser: This is fine. 
`queryId` should stay in metrics to keep API unchanged. 

But we need it among this response headers. So list of headers should be:
- `X-ClickHouse-Query-Id`
- `X-ClickHouse-Summary` (if summary changed and we do not support it yet - user will be able to get information)
- `X-ClickHouse-Server-Display-Name` 
- `X-ClickHouse-Database` 
- `X-ClickHouse-User` 

We are using whitelisting for headers for two reasons: 
- Security to not leak some sensitive header like `X-ClickHouse-Key` 
- API to expose only guaranteed headers.  


Thank you so much for helping with this change!

synhershko:
@chernser updated

chernser:
Thank you @synhershko !

PR #2756, opened 2026-02-24
Fix hasValue to handle missing columns gracefully

### LINKED ISSUE #2755: [client-v2] hasValue on GenericRecord does not work as intended
## Description

The `boolean hasValue(int)` and `boolean hasValue(string)` methods do not work as intended. If the value does not exist, an exception is thrown rather than `false`. 

### Steps to reproduce

1. Submit a basic query like (`SELECT 1 + 1 as result`)

2a. Run `hasValue("missing")`
2b. Run `hasValue(100)`
 
For 2a
```
Method threw 'com.clickhouse.client.api.metadata.NoSuchColumnException' exception.
```

For 2b
```
Method threw 'java.lang.ArrayIndexOutOfBoundsException' exception.
```

### Expected Behaviour

For both 2a and 2b, the expected behavior is for the method to return `false` if the value does not exist. 

### Notes

For 2a, see `TableSchema` - Either we should catch in the caller and return False, or return -1 as an indicator that the value is missing. 

```java
    public int nameToIndex(String name) {
        Integer index = colIndex.get(name);
        if (index == null) {
            throw new NoSuchColumnException("Result has no column with name '" + name + "'");
        }
        return index;
    }
```

For 2b, see `AbstractBinaryFormatReader` - We should just check the size of the array and see if the parameter is within bounds rather than checking for a null. 

```java
    public boolean hasValue(int colIndex) {
        return currentRecord[colIndex - 1] != null;
    }
```

@mzitnik

utafrali:
## Issue
The `hasValue()` method on GenericRecord would throw exceptions when checking for non-existent columns:
- `hasValue(int colIndex)` threw `ArrayIndexOutOfBoundsException` for out-of-range indices
- `hasValue(String colName)` threw `NoSuchColumnException` for unknown column names

## Solution
Add defensive checks to return `false` instead of throwing:
- Add bounds validation in the int-based method to check column index is within valid range (1-based, between 1 and record length)
- Wrap the name-to-index lookup in try-catch to handle `NoSuchColumnException`

This makes the API more intuitive and allows callers to safely check for column existence without exception handling.

Fixes #2755

### DIFF
--- a/client-v2/src/main/java/com/clickhouse/client/api/data_formats/internal/AbstractBinaryFormatReader.java
+++ b/client-v2/src/main/java/com/clickhouse/client/api/data_formats/internal/AbstractBinaryFormatReader.java
@@ -569,12 +569,19 @@ public abstract class AbstractBinaryFormatReader implements ClickHouseBinaryForm
 
     @Override
     public boolean hasValue(int colIndex) {
+        if (colIndex < 1 || colIndex > currentRecord.length) {
+            return false;
+        }
         return currentRecord[colIndex - 1] != null;
     }
 
     @Override
     public boolean hasValue(String colName) {
-        return hasValue(schema.nameToColumnIndex(colName));
+        try {
+            return hasValue(schema.nameToColumnIndex(colName));
+        } catch (NoSuchColumnException e) {
+            return false;
+        }
     }
 
     @Override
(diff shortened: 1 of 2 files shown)


chernser:
@utafrali please accept CLA
Thanks!

utafrali:
Closing this PR — I'm unable to sign the CLA at this time. Thank you for the review\!

PR #2758, opened 2026-02-25
[client-v2] fixed hasValue to not throw exception

chernser:
## Summary
-Fixes issue with hasValue when it throws exception for unknown column

Closes https://github.com/ClickHouse/clickhouse-java/issues/2755

## Checklist
- [x] Closes #<issue ref>

---

> [!NOTE]
> **Low Risk**
> Small, localized behavior change to column lookup/`hasValue()` with added integration coverage; minimal impact outside unknown-column checks.
> 
> **Overview**
> Fixes `hasValue()` to **return `false` instead of throwing** when callers reference an unknown column name or an out-of-range column index.
> 
> This adds non-throwing lookup helpers to `TableSchema` (`findColumnIndex`, `findColumnName`) and updates both the binary reader (`AbstractBinaryFormatReader`) and `MapBackedRecord` to use them. An integration test is added to cover both binary-reader and `queryAll` record paths with missing/nullable columns.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit c0324b852101766edec2ed82ddaf975e3c2d4c53. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/client-v2/src/main/java/com/clickhouse/client/api/data_formats/internal/AbstractBinaryFormatReader.java
+++ b/client-v2/src/main/java/com/clickhouse/client/api/data_formats/internal/AbstractBinaryFormatReader.java
@@ -577,11 +577,7 @@ public abstract class AbstractBinaryFormatReader implements ClickHouseBinaryForm
 
     @Override
     public boolean hasValue(String colName) {
-        try {
-            return hasValue(schema.nameToColumnIndex(colName));
-        } catch (NoSuchColumnException e) {
-            return false;
-        }
+        return hasValue(schema.findColumnIndex(colName));
     }
 
     @Override
--- a/client-v2/src/main/java/com/clickhouse/client/api/data_formats/internal/MapBackedRecord.java
+++ b/client-v2/src/main/java/com/clickhouse/client/api/data_formats/internal/MapBackedRecord.java
@@ -12,6 +12,7 @@ import com.clickhouse.data.value.ClickHouseGeoMultiPolygonValue;
 import com.clickhouse.data.value.ClickHouseGeoPointValue;
 import com.clickhouse.data.value.ClickHouseGeoPolygonValue;
 import com.clickhouse.data.value.ClickHouseGeoRingValue;
+import com.google.common.collect.ImmutableList;
 
 import java.math.BigDecimal;
 import java.math.BigInteger;
@@ -311,11 +312,8 @@ public class MapBackedRecord implements GenericRecord {
 
     @Override
     public boolean hasValue(int colIndex) {
-        try {
-            return hasValue(schema.columnIndexToName(colIndex));
-        } catch (NoSuchColumnException e) {
-            return false;
-        }
+        String columnName = schema.findColumnName(colIndex);
+        return columnName != null && hasValue(columnName);
     }
 
     @Override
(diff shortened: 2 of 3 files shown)


client-v2/src/main/java/com/clickhouse/client/api/data_formats/internal/AbstractBinaryFormatReader.java
```diff
@@ -569,12 +569,19 @@ public Object[] getObjectArray(String colName) {
...
+        if (colIndex < 1 || colIndex > currentRecord.length) {
+            return false;
+        }
         return currentRecord[colIndex - 1] != null;
     }
 
     @Override
     public boolean hasValue(String colName) {
-        return hasValue(schema.nameToColumnIndex(colName));
+        try {
```
mzitnik: maybe we can add `TableSchema` hasColumn instead, using try catch

client-v2/src/main/java/com/clickhouse/client/api/data_formats/internal/MapBackedRecord.java
```diff
@@ -310,7 +311,11 @@ public String[] getStringArray(String colName) {
 
     @Override
     public boolean hasValue(int colIndex) {
-        return hasValue(schema.columnIndexToName(colIndex));
+        try {
```
mzitnik: maybe we can add TableSchema hasColumn instead, using try catch

client-v2/src/main/java/com/clickhouse/client/api/data_formats/internal/AbstractBinaryFormatReader.java
chernser: if we have `hasColumn` then it will be two methods call. 
we need something like `nameToColumnIndexSafe()`, WDYT? 
But this I think overloads interface a bit. 
returning `-1` by `schema.nameToColumnIndex()` also looks not a java way. 

Idea behind exception is that application code should know what number of columns are in the result:
- we let application access schema to check dynamically 
- or application knows it upfront 

so checking value for missing column look more like a bug in application.

mzitnik: Using an exception has a little bit performance penalty.
Two methods call, are you worried about performance?
In general, we need to specify the expected behavior of our API. I prefer not use exceptions if we can avoid

chernser: I expect reading missing column is be a very specific case.
Ok, I'll implement another safe method to return -1 from schema on unknown column

chernser: but I agree that logic based on exception is bad

client-v2/src/main/java/com/clickhouse/client/api/data_formats/internal/MapBackedRecord.java
chernser: done.

PR #2771, opened 2026-03-02
Renamed NO_THROW_ON_UNKNOWN_CONFIG to IGNORE_UNKNOWN_CONFIG_KEY 

chernser:
## Summary

- Renames `NO_THROW_ON_UNKNOWN_CONFIG` to `IGNORE_UNKNOWN_CONFIG_KEY` to avoid reversed logic 
- Updates logic 
- Fixes issue with not checking value but only presence of the property

## Checklist
- [ ] Closes #<issue ref>

---

> [!NOTE]
> **Medium Risk**
> Changes the public config key used to ignore unknown properties and alters the check to depend on the property’s boolean value, which may break existing setups relying on the old key or truthy values like `1`. Scope is small but affects connection/config parsing behavior.
> 
> **Overview**
> Renames the opt-out flag for unknown/unmapped configuration from `NO_THROW_ON_UNKNOWN_CONFIG` to `IGNORE_UNKNOWN_CONFIG_KEY`.
> 
> Updates `ClientConfigProperties.parseConfigMap` so unknown config keys are ignored *only when* `IGNORE_UNKNOWN_CONFIG_KEY` is set to a boolean-true value; otherwise it still throws `ClientMisconfigurationException`. Integration tests in both `client-v2` and `jdbc-v2` are updated to use the new key/value.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 4310a95708746294e2de62eddf7dda8cb3b1b099. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/jdbc-v2/src/test/java/com/clickhouse/jdbc/DriverTest.java
+++ b/jdbc-v2/src/test/java/com/clickhouse/jdbc/DriverTest.java
@@ -194,6 +194,6 @@ public class DriverTest extends JdbcIntegrationTest {
         }
 
         // next should not throw exception
-        driver.connect(getEndpointString() + "?unknown_setting=1&" + ClientConfigProperties.NO_THROW_ON_UNKNOWN_CONFIG + "=1", new Properties()).close();
+        driver.connect(getEndpointString() + "?unknown_setting=1&" + ClientConfigProperties.IGNORE_UNKNOWN_CONFIG_KEY + "=true", new Properties()).close();
     }
 }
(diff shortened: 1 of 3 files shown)


jdbc-v2/src/test/java/com/clickhouse/jdbc/DriverTest.java
```diff
@@ -194,6 +194,6 @@ public void testUnknownSettings() throws Exception {
         }
 
         // next should not throw exception
-        driver.connect(getEndpointString() + "?unknown_setting=1&" + ClientConfigProperties.NO_THROW_ON_UNKNOWN_CONFIG + "=1", new Properties()).close();
+        driver.connect(getEndpointString() + "?unknown_setting=1&" + ClientConfigProperties.IGNORE_UNKNOWN_CONFIG_KEY + "=true", new Properties()).close();
```
mzitnik, line 197: @chernser should we also validate that this accepts `1` alongside `true`

chernser: there is another issue and we will fix it in next release. I will put in the documentation what value is accepted.

PR #2777, opened 2026-03-05
Add getSchema() to Records for schema access with empty results

PierceAndy:
## Summary

Add `getSchema()` method to `Records` class to enable access to table schema metadata even when query results are empty.

Following the pattern from PR #2004, this change exposes the `TableSchema` from the underlying `ClickHouseBinaryFormatReader`, making it consistent with `GenericRecord.getSchema()`.

**Use case:** When `queryRecords()` or `queryAll()` returns no rows, users can still inspect column metadata (names, types, etc.) without needing to iterate through records.

**Example:**
```java
Records records = client.queryRecords("SELECT * FROM table WHERE 1=0").get();
if (records.isEmpty()) {
    TableSchema schema = records.getSchema();
    schema.getColumns().forEach(col ->
        System.out.println(col.getColumnName() + ": " + col.getDataType())
    );
}
```

## Checklist

---

**CHANGELOG entry:**
```
* Add `Records.getSchema()` method to access table schema from query results, including empty result sets
```

---

> [!NOTE]
> **Low Risk**
> Low risk API addition that delegates to the existing binary reader schema with only new test coverage; minimal impact beyond adding a new method.
> 
> **Overview**
> Adds `Records.getSchema()` to return the underlying `TableSchema`, enabling schema inspection even when `Records.isEmpty()` is true.
> 
> Extends `QueryTests` with integration coverage for `getSchema()` on both empty and non-empty result sets, and asserts records’ per-row schema matches the collection schema.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit ab709af92d3ac4a380e9170796ef2e7bf65e67cb. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/client-v2/src/main/java/com/clickhouse/client/api/query/Records.java
+++ b/client-v2/src/main/java/com/clickhouse/client/api/query/Records.java
@@ -2,6 +2,7 @@ package com.clickhouse.client.api.query;
 
 import com.clickhouse.client.api.data_formats.ClickHouseBinaryFormatReader;
 import com.clickhouse.client.api.data_formats.internal.BinaryReaderBackedRecord;
+import com.clickhouse.client.api.metadata.TableSchema;
 import com.clickhouse.client.api.metrics.OperationMetrics;
 import com.clickhouse.client.api.metrics.ServerMetrics;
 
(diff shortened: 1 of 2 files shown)


chernser:
@PierceAndy, 

thank you for the contribution!

PR #2782, opened 2026-03-10
[jdbc-v2] Fix R2DBC Tests

chernser:
## Summary
- Fixes failing test in DataBaseMetadataTests
- Fixes R2DBC test

Closes  https://github.com/ClickHouse/clickhouse-java/issues/2783
## Checklist
- [ ] Closes #<issue ref>

---

> [!NOTE]
> **Medium Risk**
> Mostly test/CI changes, but it also tweaks R2DBC connection option parsing to accept `custom_http_params`, which can affect runtime connection URLs and HTTP settings behavior.
> 
> **Overview**
> Stabilizes integration tests across client/JDBC/R2DBC by consistently disabling async inserts (via `async_insert=0`/`custom_http_params`) in streaming inserts, transaction tests, proxy tests, and HTTP connection reuse tests.
> 
> Updates R2DBC URL option handling to allow passing `custom_http_params` through `ConnectionFactoryOptions`, and fixes a JDBC v2 metadata assertion by mapping `UInt128` to `Types.NUMERIC` (instead of `Types.OTHER`). Also narrows `ClickHouseTestClient.accept()` to only match the MySQL protocol to avoid unintended selection in other tests, and adjusts CI to install only JDK 21 using Corretto in the compile workflow.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit ecdbcb02ffa08b5a02431aa0cfbfc24652b2f161. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -50,7 +50,7 @@ jobs:
       - name: Install JDK 8 and Maven
         uses: actions/setup-java@v4
         with:
-          distribution: "corretto"
+          distribution: "temurin"
           java-version: 21
           cache: "maven"
       - name: Build and install libraries


.github/workflows/build.yml
```diff
@@ -50,10 +50,8 @@ jobs:
       - name: Install JDK 8 and Maven
         uses: actions/setup-java@v4
         with:
-          distribution: "temurin"
-          java-version: |
-            8
-            21
+          distribution: "corretto"
```
kurnoolsaketh: don't block on this, but why change the JVM?

clickhouse-client/src/test/java/com/clickhouse/client/ClickHouseTestClient.java
```diff
@@ -7,7 +7,7 @@ public class ClickHouseTestClient implements ClickHouseClient {
 
     @Override
     public boolean accept(ClickHouseProtocol protocol) {
-        return true;
+        return protocol == ClickHouseProtocol.MYSQL; // to avoid taking this client in other tests
```
kurnoolsaketh, line 10: can you explain this change? Again, i won't block on it but I have no context for this

.github/workflows/build.yml
chernser: It is left after experiment. I will revert it.

clickhouse-client/src/test/java/com/clickhouse/client/ClickHouseTestClient.java
chernser: Client V1 uses a ServiceLoader to discover available client implementations. 
So there is a test in clickhouse-client package that has NO implementations.
That is why ClickHouseTestClient exists. 
clickhouse-client exports test jar to provide core classes for integration tests and this stupid ClickHouseTestClient. 
Because ClickHouseTestClient looks like any other valid client implementation it is selected (randomly sometimes because no service order exist). 
Clients are selected by supported protocol. Previously this ClickHouseTestClient returned "true" for every protocol. 

Finally - this change is to make ClickHouseTestClient work for ServiceLoader test and not to be picked by production code in other tests.

PR #2790, opened 2026-03-13
fix off-by-one bug in ArrayResultSet.next()

sophiecuiy:
## Summary

- **Fix off-by-one bug in `ArrayResultSet.next()`** that caused it to return `true` one extra time past the end of the array, leading to a `"No current row"` `SQLException` on the phantom iteration.
- The guard condition `pos == length` was checked before incrementing `pos`, so when `pos` was at `length-1` (last valid index), it passed the guard, incremented to `length`, and returned `true`. Fix: change to `pos + 1 >= length`.
- Add regression tests (`testNextIterationCount`, `testNextEmptyArray`, `testNextSingleElement`) verifying the standard JDBC iteration pattern `while (rs.next()) { rs.getString(2); }` works correctly.

Discovered via Airbyte connector source-clickhouse (airbytehq/airbyte#61419).

## Checklist

- [X] Verify `testNextIterationCount` passes
- [X] Verify `testNextEmptyArray` passes
- [X] Verify `testNextSingleElement` passes
- [X] Verify existing `ArrayResultSetTest` tests still pass
- [X] Run full `jdbc-v2` test suite

### DIFF
--- a/jdbc-v2/src/main/java/com/clickhouse/jdbc/types/ArrayResultSet.java
+++ b/jdbc-v2/src/main/java/com/clickhouse/jdbc/types/ArrayResultSet.java
@@ -76,7 +76,7 @@ public class ArrayResultSet implements ResultSet {
 
     @Override
     public boolean next() throws SQLException {
-        if (pos == length || length == 0) {
+        if (pos + 1 >= length || length == 0) {
             return false;
         }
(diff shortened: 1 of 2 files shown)


chernser:
@sophiecuiy please sign CLA. 

Thanks!

PR #2815, opened 2026-03-31
[client-v2,jdbc-v2] Support Geometry Type

### LINKED ISSUE #2678: [client-v2] Add new `Geometry` type
## Description 
New type `Geometry` is released in `52.11` https://clickhouse.com/docs/whats-new/changelog#2511

Notes: 
In previous versions, the Geometry type was aliased to String, but now it is a full-featured type. [#83344](https://github.com/ClickHouse/ClickHouse/pull/83344)

Introduce Geometry type. Support reading WKB and WKT formats for it. In previous versions, the Geometry type was aliased to String, but now it is a full-featured type. [#83344](https://github.com/ClickHouse/ClickHouse/pull/83344)

https://clickhouse.com/docs/sql-reference/data-types/geo#geometry

chernser:
### Description 

ClickHouse supports many geometry types like Point, Ring, Polygon and others (https://clickhouse.com/docs/sql-reference/data-types/geo). There was an alias type `Geometry` that pointed to a `String` type. When this type is used it was replaced by `String`. This very well works with WKT functions (https://clickhouse.com/blog/state-of-geospatial-march-2026#ingesting-geospatial-data-via-wkt) that can parse different geometry notations. 

When `Variant` type was introduced it made possible to have better type alias so in `25.11` `Geometry` type became a real type not an alias. This doesn't break backward compatibility because `String` when table was created. However today cast to Geometry means different thing. 

`Variant` type of `Geometry` has a predefined list of types. Because of this we can pre-define column of such type and use it in serialization logic (client code uses `ClickHouseColumn` object to define how value should be serialized and deserialized).  Reading `Geometry` type is easy task because we client matches type ordinary number to java type. Here we should note that `Ring` and `LineString` are `Array(Point)` making them effectively same from data structure standpoint. Similar with `MultiLineString` and `Polygon` are defined as `Array(LineString)` and `Array(Ring)` respectively. This make type inference on a write operation complicated because task is to match array with specific nesting level to a type and use correct ordinary number. 

There should be a dedicated method to determine java array depth. Similar should exist for `List`. These only type client support to be written to geometry types. Because  we deal with limited number of array types it is more safe to use `instanceof Double[]` like statement instead of traversing whole array that can be quite big. In the future we will introduce some type hinting mechanism with annotations to make interface for type-safe. 

This PR:
- Adds new type to `ClickHouseDataType` enum 
- Adds `Geometry` `Variant` defined column to `ClickHouseColumn` and implement reading `Geometry` columns definitions.
- Adds method to determine array dimensions to `SerializationUtils` 
- Adds method to determine ordinary number of type for array or list of specific depth. 
- Adds tests for write and read. 

Closes: https://github.com/ClickHouse/clickhouse-java/issues/2678

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,8 @@
 
 - **[jdbc-v2]** Added `cluster_name` configuration property to specify a target cluster for statements like `KILL QUERY` that require an `ON CLUSTER` clause to execute across all nodes. (https://github.com/ClickHouse/clickhouse-java/issues/2837)
 
+- **[client-v2, jdbc-v2]** Added support for ClickHouse `Geometry` type for ClickHouse `25.11+`, where `Geometry` changed from a `String` alias to `Variant(Point, Ring, LineString, MultiLineString, Polygon, MultiPolygon)` (client still compatible with older versions). Includes client read/write handling and JDBC type mapping for retrieving and inserting geometry values. Current writes infer the target geometry variant from array nesting depth, so `Ring` vs `LineString` and `Polygon` vs `MultiLineString` are not yet distinguishable through the generic `Geometry` write path. (https://github.com/ClickHouse/clickhouse-java/pull/2815)
+
 ### Bug Fixes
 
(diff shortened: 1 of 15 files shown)


mshustov:
@claude[agent] review changes

chernser:
@claude review

chernser:
@claude review

kurnoolsaketh (approved):

PR #2857, opened 2026-05-20
jdbc-v2: swallow chunked-stream drain errors on ResultSet.close (fixes #2361)

### LINKED ISSUE #2361: Random exception while executing queries via JDBC driver
### Company or project name

_No response_

### Question

Hi,
I'm facing a problem that doesn't always occur and is therefore difficult to reproduce and resolve.
Here is some key information about the env:
- OS: Rocky Linux 9.2 (Blue Onyx)
- Language: Scala 2.13.13
- Main dependencies: 
`"com.clickhouse" % "clickhouse-jdbc" % "0.8.3"`
`"org.apache.spark" %% "spark-XXX" % "3.5.1"`

For some queries (and not always the same ones) this exception occurs:
```25/05/14 05:49:23 WARN JDBCRDD: Exception closing resultset
java.sql.SQLException: Failed to close response
	at com.clickhouse.jdbc.internal.ExceptionUtils.toSqlState(ExceptionUtils.java:66)
	at com.clickhouse.jdbc.internal.ExceptionUtils.toSqlState(ExceptionUtils.java:39)
	at com.clickhouse.jdbc.ResultSetImpl.close(ResultSetImpl.java:117)
	at org.apache.spark.sql.execution.datasources.jdbc.JDBCRDD.close$1(JDBCRDD.scala:195)
	at org.apache.spark.sql.execution.datasources.jdbc.JDBCRDD.$anonfun$compute$6(JDBCRDD.scala:225)
	at org.apache.spark.sql.execution.datasources.jdbc.JDBCRDD.$anonfun$compute$6$adapted(JDBCRDD.scala:225)
	at org.apache.spark.TaskContext$$anon$1.onTaskCompletion(TaskContext.scala:137)
	at org.apache.spark.TaskContextImpl.$anonfun$invokeTaskCompletionListeners$1(TaskContextImpl.scala:144)
	at org.apache.spark.TaskContextImpl.$anonfun$invokeTaskCompletionListeners$1$adapted(TaskContextImpl.scala:144)
	at org.apache.spark.TaskContextImpl.invokeListeners(TaskContextImpl.scala:199)
	at org.apache.spark.TaskContextImpl.invokeTaskCompletionListeners(TaskContextImpl.scala:144)
	at org.apache.spark.TaskContextImpl.markTaskCompleted(TaskContextImpl.scala:137)
	at org.apache.spark.TaskContext.runTaskWithListeners(TaskContext.scala:177)
	at org.apache.spark.scheduler.Task.run(Task.scala:141)
	at org.apache.spark.executor.Executor$TaskRunner.$anonfun$run$4(Executor.scala:620)
	at org.apache.spark.util.SparkErrorUtils.tryWithSafeFinally(SparkErrorUtils.scala:64)
	at org.apache.spark.util.SparkErrorUtils.tryWithSafeFinally$(SparkErrorUtils.scala:61)
	at org.apache.spark.util.Utils$.tryWithSafeFinally(Utils.scala:94)
	at org.apache.spark.executor.Executor$TaskRunner.run(Executor.scala:623)
	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1149)
	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:624)
	at java.lang.Thread.run(Thread.java:750)
Caused by: com.clickhouse.client.api.ClientException: Failed to close response
	at com.clickhouse.client.api.query.QueryResponse.close(QueryResponse.java:70)
	at com.clickhouse.jdbc.ResultSetImpl.close(ResultSetImpl.java:106)
	... 19 more
Caused by: org.apache.hc.core5.http.ConnectionClosedException: Premature end of chunk coded message body: closing chunk expected
	at org.apache.hc.core5.http.impl.io.ChunkedInputStream.getChunkSize(ChunkedInputStream.java:263)
	at org.apache.hc.core5.http.impl.io.ChunkedInputStream.nextChunk(ChunkedInputStream.java:222)
	at org.apache.hc.core5.http.impl.io.ChunkedInputStream.read(ChunkedInputStream.java:147)
	at org.apache.hc.core5.http.impl.io.ChunkedInputStream.close(ChunkedInputStream.java:314)
	at org.apache.hc.core5.io.Closer.close(Closer.java:48)
	at org.apache.hc.core5.http.impl.io.IncomingHttpEntity.close(IncomingHttpEntity.java:112)
	at org.apache.hc.core5.http.io.entity.HttpEntityWrapper.close(HttpEntityWrapper.java:120)
	at org.apache.hc.client5.http.impl.classic.ResponseEntityProxy.close(ResponseEntityProxy.java:180)
	at com.clickhouse.client.api.internal.LZ4Entity.close(LZ4Entity.java:104)
	at org.apache.hc.core5.io.Closer.close(Closer.java:48)
	at org.apache.hc.core5.http.message.BasicClassicHttpResponse.close(BasicClassicHttpResponse.java:93)
	at org.apache.hc.client5.http.impl.classic.CloseableHttpResponse.close(CloseableHttpResponse.java:200)
	at com.clickhouse.client.api.query.QueryResponse.close(QueryResponse.java:68)
	... 20 more
25/05/14 05:49:23 ERROR Executor: Exception in task 0.0 in stage 3684.0 (TID 119142)
java.sql.SQLException: Failed to read value for column date
	at com.clickhouse.jdbc.internal.ExceptionUtils.toSqlState(ExceptionUtils.java:66)
	at com.clickhouse.jdbc.internal.ExceptionUtils.toSqlState(ExceptionUtils.java:39)
	at com.clickhouse.jdbc.ResultSetImpl.next(ResultSetImpl.java:83)
	at org.apache.spark.sql.execution.datasources.jdbc.JdbcUtils$$anon$1.getNext(JdbcUtils.scala:354)
	at org.apache.spark.sql.execution.datasources.jdbc.JdbcUtils$$anon$1.getNext(JdbcUtils.scala:340)
	at org.apache.spark.util.NextIterator.hasNext(NextIterator.scala:73)
	at org.apache.spark.InterruptibleIterator.hasNext(InterruptibleIterator.scala:37)
	at org.apache.spark.util.CompletionIterator.hasNext(CompletionIterator.scala:31)
	at org.apache.spark.sql.catalyst.expressions.GeneratedClass$GeneratedIteratorForCodegenStage1.processNext(Unknown Source)
	at org.apache.spark.sql.execution.BufferedRowIterator.hasNext(BufferedRowIterator.java:43)
	at org.apache.spark.sql.execution.WholeStageCodegenEvaluatorFactory$WholeStageCodegenPartitionEvaluator$$anon$1.hasNext(WholeStageCodegenEvaluatorFactory.scala:43)
	at org.apache.spark.sql.execution.columnar.DefaultCachedBatchSerializer$$anon$1.next(InMemoryRelation.scala:88)
	at org.apache.spark.sql.execution.columnar.DefaultCachedBatchSerializer$$anon$1.next(InMemoryRelation.scala:80)
	at org.apache.spark.sql.execution.columnar.CachedRDDBuilder$$anon$2.next(InMemoryRelation.scala:290)
	at org.apache.spark.sql.execution.columnar.CachedRDDBuilder$$anon$2.next(InMemoryRelation.scala:287)
	at org.apache.spark.storage.memory.MemoryStore.putIterator(MemoryStore.scala:224)
	at org.apache.spark.storage.memory.MemoryStore.putIteratorAsValues(MemoryStore.scala:302)
	at org.apache.spark.storage.BlockManager.$anonfun$doPutIterator$1(BlockManager.scala:1597)
	at org.apache.spark.storage.BlockManager.org$apache$spark$storage$BlockManager$$doPut(BlockManager.scala:1524)
	at org.apache.spark.storage.BlockManager.doPutIterator(BlockManager.scala:1588)
	at org.apache.spark.storage.BlockManager.getOrElseUpdate(BlockManager.scala:1389)
	at org.apache.spark.storage.BlockManager.getOrElseUpdateRDDBlock(BlockManager.scala:1343)
	at org.apache.spark.rdd.RDD.getOrCompute(RDD.scala:379)
	at org.apache.spark.rdd.RDD.iterator(RDD.scala:329)
	at org.apache.spark.scheduler.ResultTask.runTask(ResultTask.scala:93)
	at org.apache.spark.TaskContext.runTaskWithListeners(TaskContext.scala:166)
	at org.apache.spark.scheduler.Task.run(Task.scala:141)
	at org.apache.spark.executor.Executor$TaskRunner.$anonfun$run$4(Executor.scala:620)
	at org.apache.spark.util.SparkErrorUtils.tryWithSafeFinally(SparkErrorUtils.scala:64)
	at org.apache.spark.util.SparkErrorUtils.tryWithSafeFinally$(SparkErrorUtils.scala:61)
	at org.apache.spark.util.Utils$.tryWithSafeFinally(Utils.scala:94)
	at org.apache.spark.executor.Executor$TaskRunner.run(Executor.scala:623)
	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1149)
	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:624)
	at java.lang.Thread.run(Thread.java:750)
Caused by: com.clickhouse.client.api.ClientException: Failed to read value for column date
	at com.clickhouse.client.api.data_formats.internal.BinaryStreamReader.readValue(BinaryStreamReader.java:241)
	at com.clickhouse.client.api.data_formats.internal.BinaryStreamReader.readValue(BinaryStreamReader.java:82)
	at com.clickhouse.client.api.data_formats.internal.AbstractBinaryFormatReader.readRecord(AbstractBinaryFormatReader.java:163)
	at com.clickhouse.client.api.data_formats.internal.AbstractBinaryFormatReader.readNextRecord(AbstractBinaryFormatReader.java:208)
	at com.clickhouse.client.api.data_formats.internal.AbstractBinaryFormatReader.next(AbstractBinaryFormatReader.java:229)
	at com.clickhouse.jdbc.ResultSetImpl.next(ResultSetImpl.java:81)
	... 32 more
Caused by: org.apache.hc.core5.http.ConnectionClosedException: Premature end of chunk coded message body: closing chunk expected
	at org.apache.hc.core5.http.impl.io.ChunkedInputStream.getChunkSize(ChunkedInputStream.java:263)
	at org.apache.hc.core5.http.impl.io.ChunkedInputStream.nextChunk(ChunkedInputStream.java:222)
	at org.apache.hc.core5.http.impl.io.ChunkedInputStream.read(ChunkedInputStream.java:183)
	at org.apache.hc.core5.http.io.EofSensorInputStream.read(EofSensorInputStream.java:135)
	at com.clickhouse.client.api.internal.ClickHouseLZ4InputStream.readFully(ClickHouseLZ4InputStream.java:81)
	at com.clickhouse.client.api.internal.ClickHouseLZ4InputStream.refill(ClickHouseLZ4InputStream.java:120)
	at com.clickhouse.client.api.internal.ClickHouseLZ4InputStream.read(ClickHouseLZ4InputStream.java:62)
	at com.clickhouse.client.api.data_formats.internal.BinaryStreamReader.readNBytes(BinaryStreamReader.java:541)
	at com.clickhouse.client.api.data_formats.internal.BinaryStreamReader.readShortLE(BinaryStreamReader.java:328)
	at com.clickhouse.client.api.data_formats.internal.BinaryStreamReader.readUnsignedShortLE(BinaryStreamReader.java:421)
	at com.clickhouse.client.api.data_formats.internal.BinaryStreamReader.readDate(BinaryStreamReader.java:883)
	at com.clickhouse.client.api.data_formats.internal.BinaryStreamReader.readDate(BinaryStreamReader.java:871)
	at com.clickhouse.client.api.data_formats.internal.BinaryStreamReader.readValue(BinaryStreamReader.java:173)
	... 37 more
25/05/14 05:49:23 WARN TaskSetManager: Lost task 0.0 in stage 3684.0 (TID 119142) (EEE-CCCC.BBB.AAA.ZZZ.XXXXXX.YYY executor driver): java.sql.SQLException: Failed to read value for column date
	at com.clickhouse.jdbc.internal.ExceptionUtils.toSqlState(ExceptionUtils.java:66)
	at com.clickhouse.jdbc.internal.ExceptionUtils.toSqlState(ExceptionUtils.java:39)
	at com.clickhouse.jdbc.ResultSetImpl.next(ResultSetImpl.java:83)
	at org.apache.spark.sql.execution.datasources.jdbc.JdbcUtils$$anon$1.getNext(JdbcUtils.scala:354)
	at org.apache.spark.sql.execution.datasources.jdbc.JdbcUtils$$anon$1.getNext(JdbcUtils.scala:340)
	at org.apache.spark.util.NextIterator.hasNext(NextIterator.scala:73)
	at org.apache.spark.InterruptibleIterator.hasNext(InterruptibleIterator.scala:37)
	at org.apache.spark.util.CompletionIterator.hasNext(CompletionIterator.scala:31)
	at org.apache.spark.sql.catalyst.expressions.GeneratedClass$GeneratedIteratorForCodegenStage1.processNext(Unknown Source)
	at org.apache.spark.sql.execution.BufferedRowIterator.hasNext(BufferedRowIterator.java:43)
	at org.apache.spark.sql.execution.WholeStageCodegenEvaluatorFactory$WholeStageCodegenPartitionEvaluator$$anon$1.hasNext(WholeStageCodegenEvaluatorFactory.scala:43)
	at org.apache.spark.sql.execution.columnar.DefaultCachedBatchSerializer$$anon$1.next(InMemoryRelation.scala:88)
	at org.apache.spark.sql.execution.columnar.DefaultCachedBatchSerializer$$anon$1.next(InMemoryRelation.scala:80)
	at org.apache.spark.sql.execution.columnar.CachedRDDBuilder$$anon$2.next(InMemoryRelation.scala:290)
	at org.apache.spark.sql.execution.columnar.CachedRDDBuilder$$anon$2.next(InMemoryRelation.scala:287)
	at org.apache.spark.storage.memory.MemoryStore.putIterator(MemoryStore.scala:224)
	at org.apache.spark.storage.memory.MemoryStore.putIteratorAsValues(MemoryStore.scala:302)
	at org.apache.spark.storage.BlockManager.$anonfun$doPutIterator$1(BlockManager.scala:1597)
	at org.apache.spark.storage.BlockManager.org$apache$spark$storage$BlockManager$$doPut(BlockManager.scala:1524)
	at org.apache.spark.storage.BlockManager.doPutIterator(BlockManager.scala:1588)
	at org.apache.spark.storage.BlockManager.getOrElseUpdate(BlockManager.scala:1389)
	at org.apache.spark.storage.BlockManager.getOrElseUpdateRDDBlock(BlockManager.scala:1343)
	at org.apache.spark.rdd.RDD.getOrCompute(RDD.scala:379)
	at org.apache.spark.rdd.RDD.iterator(RDD.scala:329)
	at org.apache.spark.scheduler.ResultTask.runTask(ResultTask.scala:93)
	at org.apache.spark.TaskContext.runTaskWithListeners(TaskContext.scala:166)
	at org.apache.spark.scheduler.Task.run(Task.scala:141)
	at org.apache.spark.executor.Executor$TaskRunner.$anonfun$run$4(Executor.scala:620)
	at org.apache.spark.util.SparkErrorUtils.tryWithSafeFinally(SparkErrorUtils.scala:64)
	at org.apache.spark.util.SparkErrorUtils.tryWithSafeFinally$(SparkErrorUtils.scala:61)
	at org.apache.spark.util.Utils$.tryWithSafeFinally(Utils.scala:94)
	at org.apache.spark.executor.Executor$TaskRunner.run(Executor.scala:623)
	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1149)
	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:624)
	at java.lang.Thread.run(Thread.java:750)
Caused by: com.clickhouse.client.api.ClientException: Failed to read value for column date
	at com.clickhouse.client.api.data_formats.internal.BinaryStreamReader.readValue(BinaryStreamReader.java:241)
	at com.clickhouse.client.api.data_formats.internal.BinaryStreamReader.readValue(BinaryStreamReader.java:82)
	at com.clickhouse.client.api.data_formats.internal.AbstractBinaryFormatReader.readRecord(AbstractBinaryFormatReader.java:163)
	at com.clickhouse.client.api.data_formats.internal.AbstractBinaryFormatReader.readNextRecord(AbstractBinaryFormatReader.java:208)
	at com.clickhouse.client.api.data_formats.internal.AbstractBinaryFormatReader.next(AbstractBinaryFormatReader.java:229)
	at com.clickhouse.jdbc.ResultSetImpl.next(ResultSetImpl.java:81)
	... 32 more
Caused by: org.apache.hc.core5.http.ConnectionClosedException: Premature end of chunk coded message body: closing chunk expected
	at org.apache.hc.core5.http.impl.io.ChunkedInputStream.getChunkSize(ChunkedInputStream.java:263)
	at org.apache.hc.core5.http.impl.io.ChunkedInputStream.nextChunk(ChunkedInputStream.java:222)
	at org.apache.hc.core5.http.impl.io.ChunkedInputStream.read(ChunkedInputStream.java:183)
	at org.apache.hc.core5.http.io.EofSensorInputStream.read(EofSensorInputStream.java:135)
	at com.clickhouse.client.api.internal.ClickHouseLZ4InputStream.readFully(ClickHouseLZ4InputStream.java:81)
	at com.clickhouse.client.api.internal.ClickHouseLZ4InputStream.refill(ClickHouseLZ4InputStream.java:120)
	at com.clickhouse.client.api.internal.ClickHouseLZ4InputStream.read(ClickHouseLZ4InputStream.java:62)
	at com.clickhouse.client.api.data_formats.internal.BinaryStreamReader.readNBytes(BinaryStreamReader.java:541)
	at com.clickhouse.client.api.data_formats.internal.BinaryStreamReader.readShortLE(BinaryStreamReader.java:328)
	at com.clickhouse.client.api.data_formats.internal.BinaryStreamReader.readUnsignedShortLE(BinaryStreamReader.java:421)
	at com.clickhouse.client.api.data_formats.internal.BinaryStreamReader.readDate(BinaryStreamReader.java:883)
	at com.clickhouse.client.api.data_formats.internal.BinaryStreamReader.readDate(BinaryStreamReader.java:871)
	at com.clickhouse.client.api.data_formats.internal.BinaryStreamReader.readValue(BinaryStreamReader.java:173)
	... 37 more
```
It is often sufficient to repeat the query but I would like a better solution to the problem.
At the moment I do not have any parameters set at the JDBC level, everything set to default.
Any raccomendation?

fm4v:
## Summary

Fixes #2361. `ResultSetImpl.close()` no longer rethrows `ConnectionClosedException: Premature end of chunk coded message body: closing chunk expected` from the chunked-response drain path. Iteration-time failures still surface through `next()`; this only fixes the close-cleanup noise.

## Why

When the server tears down the response connection mid-stream (canonically because `send_timeout` fires on the writer side and the terminating zero-length chunk is never written), Apache HC's `ChunkedInputStream.close()` drains remaining bytes and trips `Premature end of chunk coded message body`. With `compress=true` the close path goes through `FramedLZ4CompressorInputStream.close()` → `BoundedInputStream.close()` → `ChunkedInputStream.close()` and the same drain failure surfaces.

`ResultSetImpl.close()` previously rethrew this as a `SQLException`, which breaks well-behaved try-with-resources callers: they have already finished iterating, they cannot affect the server-side socket race, and the bug is surfacing in *cleanup* — they have no recourse other than wrapping every `rs.close()` (often implicit) in a try/catch. The reporter in #2361 ended up implementing client-side retry for the symptom; sqlancer / SQL fuzzers hit the same pattern.

If iteration genuinely fails (server closed mid-`next()`), that exception still surfaces — through `next()`, not through `close()`. The bare-drain-on-close case is informational only.

## What changed

1. `jdbc-v2/src/main/java/com/clickhouse/jdbc/ResultSetImpl.java`: gate the existing `e = re;` assignments in `close()` behind a new `isStreamDrainException` classifier. Drain-class exceptions are logged at debug and swallowed; all other close-time failures still propagate as before. Logic shape preserved.

2. `jdbc-v2/src/test/java/com/clickhouse/jdbc/ResultSetImplCloseTest.java`: six unit tests covering the classifier (bare message match, `closing chunk expected` variant, nested cause chain, class-name match without message, unrelated exceptions, null input). All pass.

3. `examples/jdbc/src/main/java/com/clickhouse/examples/jdbc/Issue2361Repro.java`: a self-contained `main` that reproduces the bug against a real ClickHouse server. Trigger recipe is in the class Javadoc: compression on, server `send_timeout=1`, buffered chunked response, slow client read. On a freshly-started server it trips at ~100% rate.

## Reproducer

```bash
java -DchUrl=jdbc:ch://localhost:8123 \
     com.clickhouse.examples.jdbc.Issue2361Repro 3
```

Before this PR (jdbc-v2 0.9.8 main):
```
iter 0: trips=1 elapsed=126840ms
iter 1: trips=2 elapsed=131245ms
...
FINAL: trips=2 / 3 iterations (67%)
```

After this PR:
```
iter 0: trips=0 elapsed=...
iter 1: trips=0 elapsed=...
FINAL: trips=0 / 3 iterations (0%)
```

## Test plan
- [x] `mvn -pl jdbc-v2 test -Dtest=ResultSetImplCloseTest` — 6/6 pass
- [x] Standalone reproducer demonstrates 100% → 0% trip rate before vs after
- [ ] Maintainer review for whether to also extend `isStreamDrainException` to other transient-cleanup exception classes (e.g. SSL `SSLException` on close, `SocketException: Connection reset` on close)

## Correctness — is the data the application consumed complete?

Yes, in the failure scenario this PR addresses. The relevant invariants are in upstream code already, not added here:

1. **`AbstractBinaryFormatReader.readRecord(...)`** catches `EOFException` only when `firstColumn=true` (between rows) and treats it as end-of-stream. EOF mid-row is rethrown as `IOException(recordReadExceptionMsg(...), e)`, which surfaces through `next()` — not through `close()`.

2. **`ChunkedInputStream.read(...)`** in Apache HC returns `-1` only when it has parsed the terminator zero-length chunk. If the TCP connection drops mid-chunk-header it throws `ConnectionClosedException: Premature end of chunk coded message body: closing chunk expected`; if the chunk header is corrupt it throws `MalformedChunkCodingException`. Truncation never quietly looks like clean EOF to the upper layer.

3. **Therefore the failure path covered by this PR** corresponds to: the lz4-framed stream's end-of-stream marker was seen, the reader fully consumed the payload, `next()` legitimately returned `false`, and only then did the chunked-stream drain in `close()` hit the missing terminator chunk. The application has the correct row set; the close-time error is an HTTP-framing artefact, not a data-integrity one.

Other truncation cases (mid-row, mid-frame-header, malformed chunk) all surface through `next()` before `close()` is reached — those paths are unchanged by this PR. The `isStreamDrainException` classifier is intentionally narrow (`Premature end of chunk` / `closing chunk expected` message, or `ConnectionClosedException` class match) so unrelated close-time failures still propagate as before.


fm4v:
## Local end-to-end validation against ClickHouse 26.5.1.854

Reproduced #2361 against the stock 0.9.8 jar and confirmed the fix on this branch using the `Issue2361Repro` example from this PR, plus a variant that also validates row content.

### Bug is real

Stock `clickhouse-jdbc-0.9.8-all.jar` against a server reachable on `:18124`, 3 iterations of the reproducer:

```
iter 0: trips=1 elapsed=38882ms
iter 1: trips=2 elapsed=347743ms
iter 2: trips=3 elapsed=37341ms

FINAL: trips=3 / 3 iterations (100%)

-- first failure stack (top frames) --
java.sql.SQLException: Premature end of chunk coded message body: closing chunk expected
    at com.clickhouse.jdbc.internal.ExceptionUtils.toSqlState(ExceptionUtils.java:75)
    at com.clickhouse.jdbc.internal.ExceptionUtils.toSqlState(ExceptionUtils.java:43)
    at com.clickhouse.jdbc.ResultSetImpl.close(ResultSetImpl.java:179)
    at com.clickhouse.examples.jdbc.Issue2361Repro.main(Issue2361Repro.java:92)
Caused by: com.clickhouse.client.internal.org.apache.hc.core5.http.ConnectionClosedException: Premature end of chunk coded message body: closing chunk expected
    at com.clickhouse.client.internal.org.apache.hc.core5.http.impl.io.ChunkedInputStream.getChunkSize(ChunkedInputStream.java:264)
    at com.clickhouse.client.internal.org.apache.hc.core5.http.impl.io.ChunkedInputStream.nextChunk(ChunkedInputStream.java:223)
    at com.clickhouse.client.internal.org.apache.hc.core5.http.impl.io.ChunkedInputStream.read(ChunkedInputStream.java:147)
    at com.clickhouse.client.internal.org.apache.hc.core5.http.impl.io.ChunkedInputStream.close(ChunkedInputStream.java:315)
    at com.clickhouse.client.internal.org.apache.hc.client5.http.impl.classic.ResponseEntityProxy.streamClosed(ResponseEntityProxy.java:133)
    at com.clickhouse.client.internal.org.apache.hc.core5.http.io.EofSensorInputStream.checkClose(EofSensorInputStream.java:231)
    at com.clickhouse.client.internal.org.apache.hc.core5.http.io.EofSensorInputStream.close(EofSensorInputStream.java:175)
    at com.clickhouse.client.internal.org.apache.commons.io.IOUtils.close(IOUtils.java:449)
    at com.clickhouse.client.internal.org.apache.commons.io.input.ProxyInputStream.close(ProxyInputStream.java:232)
    at com.clickhouse.client.internal.org.apache.commons.io.input.BoundedInputStream.close(BoundedInputStream.java:393)
    at com.clickhouse.client.internal.org.apache.commons.compress.compressors.lz4.FramedLZ4CompressorInputStream.close(FramedLZ4CompressorInputStream.java:164)
    at com.clickhouse.client.api.data_formats.internal.AbstractBinaryFormatReader.close(AbstractBinaryFormatReader.java:1068)
    at com.clickhouse.jdbc.ResultSetImpl.close(ResultSetImpl.java:156)
```

Stack origin matches the patch site exactly: `ResultSetImpl.close():179`, drain path
`FramedLZ4CompressorInputStream.close` → `BoundedInputStream.close` → `ChunkedInputStream.close`.

### Fix works

Patched jar (this branch) against a freshly-restarted server, 2 iterations of the reproducer:

```
iter 0: rows=65412 badNumber=3 badPad=3 tripped=false elapsed=36552ms
iter 1: rows=65412 badNumber=3 badPad=3 tripped=false elapsed=35476ms

FINAL: trips=0 rows_total=130824 badNumber=6 badPad=6
```

Zero `Premature end of chunk` propagations. `try (ResultSet rs = ...) { ... }` completes cleanly.

### No row-content regression

I extended the reproducer to also validate each row delivered before the response is torn down:

* col 0 = `number` — expected sequential 0,1,2,…
* col 1 = `repeat('xyz', 1500)` — expected the 4500-char string

Side-by-side, both jars deliver byte-identical rows:

|                                | Stock (.bak) | Patched (this PR)     |
| ------------------------------ | ------------ | --------------------- |
| `Premature end of chunk` trips | 2 / 2 iters  | 0 / 2 iters           |
| Rows delivered / iter          | 65412        | 65412                 |
| Index of first "bad" row       | 65409        | 65409                 |
| First bad `number` value       | `7305815397312170509` | `7305815397312170509` |
| First bad `pad` head           | `tion__\n<random>…Code:` | `tion__\n<random>…Code:` |

The 3 "bad" rows per iteration are **not** caused by the patch — they are ClickHouse's error trailer (`Code: …. DB::Exception: ParallelFormattingException___<hex>…`) being inlined into the truncated chunked body when `send_timeout=1` fires mid-write, then misread as RowBinary by the format reader. The exact same artifact appears with the stock jar, with identical byte counts and identical leading bytes. The patch only changes whether the cosmetic close-time exception propagates; it does not touch the row-decode path, and the empirical results confirm that.

This artifact also surfaces correctly as a `mid-iter SQLException: Failed to read next row` in both runs, before `close()` runs — so a real iteration-time problem is still propagated. The fix scope (downgrade close-time drain noise only) holds in practice.

### Environment

* ClickHouse server: `clickhouse/clickhouse-server:head` 26.5.1.854
* JDK 25
* JDBC connection settings: `compress=true`, `client.use_http_compression=true`, `socket_timeout=60000`, `clickhouse_setting_send_timeout=1`, `clickhouse_setting_http_response_buffer_size=104857600`, `clickhouse_setting_wait_end_of_query=1`, `clickhouse_setting_max_execution_time=120`
* Query: `SELECT number, repeat('xyz', 1500) AS pad FROM numbers(5000000)`
* Slow client (~50 ms sleep per 100 rows)
* Driver under test built from `jdbc-v2/src/main/java/com/clickhouse/jdbc/ResultSetImpl.java` at the current PR head
* Unit tests `ResultSetImplCloseTest` (6 cases — bare message, `closing chunk expected`, wrapped cause, class-name match without message, unrelated, `null`): all 6 pass

fm4v:
@chernser @mzitnik hi, could you please review

examples/jdbc/src/main/java/com/clickhouse/examples/jdbc/Issue2361Repro.java
```diff
@@ -0,0 +1,127 @@
...
+ * the connection without emitting the terminating zero-length chunk, and
+ * the driver's stream-drain on close trips the exception.
+ *
+ * <p>Run with: {@code java -DchUrl=jdbc:ch://localhost:8123 com.clickhouse.examples.jdbc.Issue2361Repro [iters]}
+ *
+ * <p>The fix in {@code ResultSetImpl.close()} downgrades this drain-time
+ * failure to a debug log instead of a thrown SQLException — close() should
+ * never punish callers for a server-side teardown race after iteration is done.
+ */
+public class Issue2361Repro {
```
chernser: This should be a test. Examples are only for documentation purpose.

jdbc-v2/src/test/java/com/clickhouse/jdbc/ResultSetImplCloseTest.java
```diff
@@ -0,0 +1,72 @@
...
+ * failure (to swallow) or a real error (to propagate).
+ *
+ * Background: see issue #2361. When the server's send_timeout fires mid-write,
+ * Apache HC's ChunkedInputStream.close() trips
+ * `ConnectionClosedException: Premature end of chunk coded message body:
+ * closing chunk expected`. That happens during ResultSetImpl.close(), AFTER
+ * the application has finished iterating. Propagating it punishes well-behaved
+ * try-with-resources callers for a server-side socket race they cannot affect.
+ */
+public class ResultSetImplCloseTest {
```
chernser: Please squash into a fewer tests and move to `ResultSetImplTest`. 
We plan to do another grouping. 
Please mark tests included in `integration` group because for historical and practical reasons we do most tests against ClickHouse instance.

jdbc-v2/src/test/java/com/clickhouse/jdbc/ResultSetImplCloseTest.java
```diff
@@ -0,0 +1,72 @@
...
+
+import static org.testng.Assert.assertFalse;
+import static org.testng.Assert.assertTrue;
+
+/**
+ * Unit tests for ResultSetImpl.isStreamDrainException — the classifier that
+ * decides whether a close-time exception is a benign chunked-stream drain
+ * failure (to swallow) or a real error (to propagate).
+ *
+ * Background: see issue #2361. When the server's send_timeout fires mid-write,
```
chernser: This should document scenario without reference to issue in external system (issue may gone). 
This documentation make more sense in close method of result set impl because we start researching from production code not tests.

jdbc-v2/src/main/java/com/clickhouse/jdbc/ResultSetImpl.java
```diff
@@ -149,14 +149,28 @@ public boolean next() throws SQLException {
     public void close() throws SQLException {
         closed = true;
 
+        // Stream-close exceptions on the response body are expected when the server has
+        // already torn down the connection (e.g. SOCKET_TIMEOUT on the writer side hits
+        // `send_timeout` before the terminating chunk is written). The most common
+        // surface is `ConnectionClosedException: Premature end of chunk coded message
+        // body: closing chunk expected` from Apache HC's `ChunkedInputStream` drain,
```
chernser: please explain in 3 lines comment. 
The idea I think that there are cases where we may hide real exception what cause problematic investigation. One of such cases is premature end of chunk in combination with compression.

jdbc-v2/src/main/java/com/clickhouse/jdbc/ResultSetImpl.java
```diff
@@ -180,6 +196,33 @@ public void close() throws SQLException {
...
+     * actionable from the caller's perspective -- iteration has finished, the
+     * connection is closing anyway, and the server's premature disconnect is the
+     * actual error condition (which, if it occurred during iteration, would have
+     * surfaced through {@code next()}).
+     */
+    // Package-private for unit tests.
+    static boolean isStreamDrainException(Throwable t) {
+        while (t != null) {
+            String msg = t.getMessage();
+            if (msg != null && (msg.contains("Premature end of chunk")
```
chernser: we should detect that it is `org.apache.hc.core5.http.ConnectionClosedException` and log it. 
we avoid building logic on error messages.

jdbc-v2/src/main/java/com/clickhouse/jdbc/ResultSetImpl.java
```diff
@@ -149,14 +149,28 @@ public boolean next() throws SQLException {
...
+        // See https://github.com/ClickHouse/clickhouse-java/issues/2361
         Exception e = null;
         try {
             if (reader != null) {
                 try {
                     reader.close();
                 } catch (Exception re) {
                     log.debug("Error closing reader", re);
-                    e = re;
+                    if (!isStreamDrainException(re)) {
```
chernser: this should be handled in client-v2. 
It could be handled in reader but reader knows nothing about `org.apache.hc.core5.http.ConnectionClosedException`. There are two options 
- just log all exception while closing reader and report them as error or warn
- add `closeResponseStream` to `com.clickhouse.client.api.internal.HttpAPIClientHelper`

chernser:
@fm4v 

Thank you for the contribution! It looks solid. 
Please response to my comment and sync with main so CI pass. 

Thanks you! 

PS: I will work on contributors guide this week. It should make some part clear for future contributions.

examples/jdbc/src/main/java/com/clickhouse/examples/jdbc/Issue2361Repro.java
fm4v: Removed. The Issue2361Repro example has been deleted in the latest commit. The trigger recipe (compress=true, server send_timeout=1, buffered chunked response, slow client read) is now documented inline at the detection site in AbstractBinaryFormatReader.close().

jdbc-v2/src/test/java/com/clickhouse/jdbc/ResultSetImplCloseTest.java
fm4v: Squashed and moved. Since the close-time handling now lives in client-v2 (AbstractBinaryFormatReader.close) rather than jdbc-v2, the tests followed: 6 unit tests collapsed into a single data-driven test (AbstractBinaryFormatReaderCloseTest) in client-v2. Marked the `unit` group, matching the convention of other client-v2 unit tests like SerializerUtilsTest. The new test uses the real `org.apache.hc.core5.http.ConnectionClosedException` (already a client-v2 dep) instead of a stand-in.

jdbc-v2/src/test/java/com/clickhouse/jdbc/ResultSetImplCloseTest.java
fm4v: Done — the test file is gone, replaced by AbstractBinaryFormatReaderCloseTest in client-v2 whose javadoc describes the scenario (server tearing down the connection before the terminating zero-length chunk is written) with no reference to an external issue tracker. The production-side WHY now lives in AbstractBinaryFormatReader.close(), which is the starting point when investigating.

jdbc-v2/src/main/java/com/clickhouse/jdbc/ResultSetImpl.java
fm4v: ResultSetImpl.close() is reverted to its original shape — the long comment is gone entirely. The drain handling moved into client-v2 (AbstractBinaryFormatReader.close), and the 6-line WHY now lives there, at the actual catch site. ResultSetImpl just propagates whatever the reader gives it, like before.

jdbc-v2/src/main/java/com/clickhouse/jdbc/ResultSetImpl.java
fm4v: Done. The detection is now class-based — walks the cause chain and matches `getClass().getName().endsWith(".ConnectionClosedException")`. No message-text matching. The `.endsWith` (with leading dot) tolerates both the unshaded class name (`org.apache.hc.core5.http.ConnectionClosedException`) and the shaded one (`com.clickhouse.shaded.org.apache.hc.core5.http.ConnectionClosedException`) without taking a compile-time dependency on HC from the reader layer.

jdbc-v2/src/main/java/com/clickhouse/jdbc/ResultSetImpl.java
fm4v: Moved to client-v2. Drain handling now lives in `AbstractBinaryFormatReader.close()` — that is where `input.close()` actually triggers `ChunkedInputStream.close()`, so it is the natural catch site. ResultSetImpl.close() is back to its original shape with no special-case logic. The reader detects `ConnectionClosedException` by class-name suffix on the cause chain, so no compile-time HC dependency is added to the reader. Went with your first option (handle in reader) rather than adding `closeResponseStream` to HttpAPIClientHelper since the failure originates inside the reader's stream chain, not at the response level.

fm4v:
@chernser thanks for the careful review. Pushed `4215ce7` addressing all six threads. Headline changes:

- **Moved drain handling from jdbc-v2 into client-v2** (`AbstractBinaryFormatReader.close()`). That is where `input.close()` actually triggers `ChunkedInputStream.close()` in the stack, so it is the natural catch site. `ResultSetImpl.close()` reverts to its original shape — no `isStreamDrainException`, no long comment.
- **Class-based detection**, no message-text matching: walks the cause chain and matches `getClass().getName().endsWith(".ConnectionClosedException")`. Tolerant of both the unshaded and the shaded HC class name without adding a compile-time HC dependency to the reader.
- **Six unit tests → one data-driven unit test** (`AbstractBinaryFormatReaderCloseTest` in client-v2). Uses the real `org.apache.hc.core5.http.ConnectionClosedException`. Marked `unit` group matching the convention of `SerializerUtilsTest` etc.
- **Issue2361Repro example removed**; the trigger recipe is captured in commit messages and the production-code comment.

Branch is already on top of latest `origin/main` (no rebase needed). Local validation: `client-v2` unit tests pass (4/4), `client-v2 compile` + `jdbc-v2 test-compile` + `examples/jdbc compile` all BUILD SUCCESS.

Replies on each individual thread point to the specific change.

client-v2/src/test/java/com/clickhouse/client/api/data_formats/internal/AbstractBinaryFormatReaderCloseTest.java
```diff
@@ -0,0 +1,36 @@
...
+/**
+ * Verifies that the reader's close path recognises Apache HC's
+ * ConnectionClosedException (the "Premature end of chunk coded message body"
+ * surface that fires when the server tore the connection down before writing
+ * the terminating zero-length chunk) and is willing to swallow it on close.
+ * Matched by class-name suffix so the recogniser works against both the
+ * directly-referenced and the shaded copy of the HC class.
+ */
+@Test(groups = {"unit"})
+public class AbstractBinaryFormatReaderCloseTest {
```
chernser, line 20: there was example code that reproduces the issue. Would you please add test to `com.clickhouse.client.HttpTransportTests`. 

Regarding this tests - as I've mentioned, we do not make test for specific functionality or problem. It should be inside reader test suite.

client-v2/src/main/java/com/clickhouse/client/api/data_formats/internal/AbstractBinaryFormatReader.java
```diff
@@ -1086,7 +1086,36 @@ public ClickHouseBitmap getClickHouseBitmap(int index) {
...
+    }
+
+    /**
+     * Walks the cause chain looking for an {@code org.apache.hc.core5.http.ConnectionClosedException}.
+     * Matched by class-name suffix so it works against both directly-referenced and shaded copies of
+     * the HC class without taking a compile-time dependency on it from this layer.
+     */
+    static boolean isConnectionClosedException(Throwable t) {
+        while (t != null) {
+            if (t.getClass().getName().endsWith(".ConnectionClosedException")) {
```
chernser, line 1113: here we should check for class not a name. 
And because class belongs to http client implementation this method should be in `com.clickhouse.client.api.internal.HttpAPIClientHelper` (not a great solution but we will fix it)

chernser (changes requested):

PR #2875, opened 2026-06-11
Trigger Tableau TDVT in connectors on snapshot publish and release (CON-264)

ShimonSte:
Tableau users install the latest ClickHouse JDBC driver manually, so a snapshot or release regression reaches them with no connector release in between. After a publish, fire a `repository_dispatch` to `ClickHouse/connectors` so its Tableau TDVT suite runs against the new driver. Tracking: CON-264.

- **`nightly.yml`:** dispatch `event_type=jdbc-snapshot`; version derived from the pom `<revision>` property (`env.CHC_VERSION` is commented out).
- **`release.yml`:** dispatch `event_type=jdbc-release` with the release `version` input.
- Both are `continue-on-error` so a dispatch failure never fails the publish, and require `secrets.CONNECTORS_DISPATCH_TOKEN` (Contents: read/write on connectors only).

### DIFF
--- a/.github/scripts/mirror_tdvt.py
+++ b/.github/scripts/mirror_tdvt.py
@@ -1,0 +1,198 @@
+#!/usr/bin/env python3
+"""Dispatch Tableau TDVT and mirror its result. Used by nightly.yml and release.yml."""
+
+from __future__ import annotations
+
+import argparse
+import json
+import os
+import re
+import subprocess
+import sys
+import time
+from datetime import datetime, timedelta, timezone
+
+TDVT_REPO = "ClickHouse/clickhouse-tableau-tdvt"
+WORKFLOW = "tableau-tdvt.yml"
+POLL_ATTEMPTS = 24
+POLL_INTERVAL_SEC = 15
+WATCH_INTERVAL_SEC = 30
+CREATED_FILTER_SKEW_SEC = 120
+
+
+def fail(message: str) -> None:
+    print(f"::error::{message}", file=sys.stderr)
+    sys.exit(1)
+
+
+def gh_json(*args: str) -> dict | list:
+    result = subprocess.run(["gh", *args], check=True, capture_output=True, text=True)
+    return json.loads(result.stdout)
+
+
+def gh_api_call(path: str, *fields: str) -> None:
+    args = ["gh", "api", path]
+    for field in fields:
+        flag = "-F" if field.startswith("client_payload[") else "-f"
+        args.extend([flag, field])
+    subprocess.run(args, check=True)
+
+
+def latest_run_id(repo: str) -> int:
+    data = gh_json(
+        "api",
+        f"repos/{repo}/actions/workflows/{WORKFLOW}/runs?event=repository_dispatch&per_page=1",
+    )
+    runs = data.get("workflow_runs", [])
+    return runs[0]["id"] if runs else 0
+
+
+def list_runs(repo: str, created_since: str) -> list[dict]:
+    data = gh_json(
+        "api",
+        f"repos/{repo}/actions/workflows/{WORKFLOW}/runs"
+        f"?event=repository_dispatch&created=>={created_since}&per_page=20",
+    )
+    return data.get("workflow_runs", [])
+
+
+def title_has_version(display_title: str, version: str) -> bool:
+    # TDVT run-name is "TDVT (JDBC {version})" — require the closing ")" boundary.
+    return re.search(r"\(JDBC " + re.escape(version) + r"\)", display_title) is not None
+
+
+def version_matches(runs: list[dict], threshold: int, version: str) -> list[dict]:
+    return [
+        run
+        for run in runs
+        if run.get("id", 0) > threshold
+        and title_has_version(run.get("display_title", ""), version)
+    ]
+
+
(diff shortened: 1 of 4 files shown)


chernser:
@ShimonSte 

Thank you for the PR!

.github/workflows/release.yml
```diff
@@ -88,3 +88,17 @@ jobs:
             clickhouse-r2dbc/target/clickhouse*.jar
             client-v2/target/client-v2*.jar
             jdbc-v2/target/jdbc-v2*.jar
+      # Tableau users install the latest JDBC driver manually, so a release
```
chernser: Release workflow does only push to staging and library may not be accessible for some time after that. 
I think, we need somehow grab latest from maven central and test every week. 

Idea is:
- if something broken - we should catch it on nightly
- if we doing a release and it is a release candidate - then latest should catch but actually we may trigger tests earlier.

ShimonSte: Going back to this (the last thing that needs to be done :-) )
You don't think we should check the release before pushing to staging? can we reuse the jar built locally on the workflow machine before being pushed?

chernser: @ShimonSte 
We should check release before considering release. But release workflow is a separate task that should have minimal dependencies otherwise we may get blocked. 

Reuse jar where?
We cannot build anything locally and publish. Jar must be built from protected branch and signed by key few people can access. Otherwise it is not a trusted binary.

.github/workflows/nightly.yml
```diff
@@ -63,6 +65,12 @@ jobs:
           #   -e 's|^\(    <version>\).*\(</version>\)$|\1${{ env.CHC_VERSION }}-SNAPSHOT\2|' \
           #   -e 's|${parent.groupId}|com.clickhouse|g' -e 's|${project.parent.groupId}|com.clickhouse|g' '{}' \;
           find . -type f -name "simplelogger.*" -exec rm -fv '{}' \;
+      - name: Record published snapshot version
+        id: publish-version
+        run: |
+          VERSION=$(sed -n 's|.*<revision>\(.*\)</revision>.*|\1|p' pom.xml | head -1)
```
chernser, line 71: btw, we have moved to `VERSION` file to simplify such things.

chernser (approved):

.github/workflows/nightly.yml
```diff
@@ -84,3 +92,29 @@ jobs:
...
+          owner: ${{ github.repository_owner }}
+          repositories: "clickhouse-tableau-tdvt"
+      - name: Run Tableau TDVT and mirror its result
+        env:
+          GH_TOKEN: ${{ steps.tdvt-token.outputs.token }}
+        run: |
+          python3 .github/scripts/mirror_tdvt.py \
+            --event-type jdbc-snapshot \
+            --version "${{ needs.nightly.outputs.jdbc_version }}" \
+            --correlation-id "${{ github.repository }}#${{ github.run_id }}-${{ github.run_attempt }}"
```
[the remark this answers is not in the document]
chernser: This is fine :-) 
We need to know it is failed but we cannot block the publish. Will fix it later.

.github/scripts/mirror_tdvt.py
```diff
@@ -0,0 +1,198 @@
...
+    if len(corr) == 1:
+        return corr[0]["id"]
+    if len(corr) > 1:
+        return None
+    if len(matches) > 1:
+        return None
+    if len(matches) == 1:
+        new_runs = [run for run in runs if run.get("id", 0) > threshold]
+        if len(new_runs) > 1:
+            return matches[0]["id"]
```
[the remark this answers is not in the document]
chernser: @ShimonSte what do you think?

.github/workflows/nightly.yml
ShimonSte: I think the nightly creates its version from the revision tag in the `pom.xml` (which is a version behind the one in the `VERSION` file).
Correct me if i'm wrong.

.github/scripts/mirror_tdvt.py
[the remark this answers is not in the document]
ShimonSte: This is an edge case where we run multiple tdvt tests at once (shouldn't happen)
Anyway, i changed the tdvt tests so now we get the ID and compare it exactly.

.github/workflows/release.yml
```diff
@@ -85,3 +85,29 @@ jobs:
...
+          private-key: ${{ secrets.WORKFLOW_AUTH_PUBLIC_PRIVATE_KEY }}
+          owner: ${{ github.repository_owner }}
+          repositories: "clickhouse-tableau-tdvt"
+      - name: Run Tableau TDVT and mirror its result
+        env:
+          GH_TOKEN: ${{ steps.tdvt-token.outputs.token }}
+        run: |
+          python3 .github/scripts/mirror_tdvt.py \
+            --event-type jdbc-release \
+            --version "${{ github.event.inputs.version }}" \
```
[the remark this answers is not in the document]
ShimonSte: This is correct, fixed it

.github/workflows/nightly.yml
chernser: these versions must be synced. but you are right - lets take from true source.

PR #2894, opened 2026-06-25
Fix client v2 varint overflow

romka:
## Summary
Fixes `BinaryStreamReader.readVarInt` in `client-v2` so malformed or overflowing varints are rejected instead of being decoded into corrupted or negative `int` values.

The previous implementation accumulated into an `int` while looping up to 10 bytes. For bytes after the 5th, Java masks `int` shift distances modulo 32, so overlong varints could corrupt lower bits. It also allowed 5-byte values above `Integer.MAX_VALUE`, which can become negative sizes/counts.

This PR covers one of the cases described in the issue https://github.com/ClickHouse/clickhouse-java/issues/2902.

## Changes

- Limit `readVarInt` decoding to 5 bytes.
- Reject fifth-byte payloads that exceed non-negative `int` range or continue past 5 bytes.
- Add focused tests for:
  - `Integer.MAX_VALUE`
  - overflow above `Integer.MAX_VALUE`
  - overlong varint encoding

## Compatibility

This changes handling of malformed/overflowing binary varints from silent decoding to `IOException`. Valid non-negative `int` varints are unchanged.

## Testing

- `mvn -Duser.timezone=UTC -pl client-v2 -am -Dtest=BinaryStreamReaderTests -Dsurefire.failIfNoSpecifiedTests=false test`
- `mvn -Duser.timezone=UTC -DskipITs clean verify`

## Checklist
- [x] Unit tests covering the common scenarios were added

### DIFF
--- a/client-v2/src/main/java/com/clickhouse/client/api/data_formats/internal/BinaryStreamReader.java
+++ b/client-v2/src/main/java/com/clickhouse/client/api/data_formats/internal/BinaryStreamReader.java
@@ -959,16 +959,20 @@ public class BinaryStreamReader {
     public static int readVarInt(InputStream input) throws IOException {
         int value = 0;
 
-        for (int i = 0; i < 10; i++) {
+        for (int i = 0; i < 5; i++) {
             byte b = (byte) readByteOrEOF(input);
+            if (i == 4 && (b & 0xF8) != 0) {
+                throw new IOException("VarInt is too large for int");
+            }
+
             value |= (b & 0x7F) << (7 * i);
 
             if ((b & 0x80) == 0) {
-                break;
+                return value;
             }
         }
 
-        return value;
+        throw new IOException("Malformed VarInt");
     }
 
     /**
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,8 @@
 
 ### Bug Fixes 
 
+- **[client-v2]** Fixed binary varint decoding for length and count fields so overflowing or overlong values fail with an `IOException` instead of being decoded into corrupted or negative `int` values. (https://github.com/ClickHouse/clickhouse-java/issues/2902)
+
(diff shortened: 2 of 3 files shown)


chernser:
@romka 

Thank you for the contribution! We will review the PR. 
Please create an issue so this PR is not orphan. 

Thanks!

client-v2/src/main/java/com/clickhouse/client/api/data_formats/internal/BinaryStreamReader.java
```diff
@@ -959,16 +959,20 @@ private double[][][][] readGeoMultiPolygon() throws IOException {
     public static int readVarInt(InputStream input) throws IOException {
         int value = 0;
 
-        for (int i = 0; i < 10; i++) {
+        for (int i = 0; i < 5; i++) {
```
chernser, line 962: what var int could be in data from server? 
problem may be we need to adjust using 64-bit varint.

romka: Please find more details in the issue: https://github.com/ClickHouse/clickhouse-java/issues/2902.

The tests in this PR demonstrate two cases that are currently decoded incorrectly:
- Overflow: `80 80 80 80 08`,
- Overlong encoding: `80 80 80 80 80 01`.

I agree that if any of these readers are expected to decode protocol-level 64-bit varints, then introducing a dedicated long-returning reader (and updating the relevant call sites) would be a better long-term solution. My goal in this PR was to make the current int-returning implementation fail explicitly instead of silently returning a corrupted value.

chernser:
@romka 
change is fine. but build is failing. Would you please merge main. 
Also we release `0.10.0-rc1` - so changes in the change log should be under `0.11.0` 

Thank you!

chernser (approved):

Issue #2263, opened 2025-04-01
[client-v2] Add support of working with large strings


Paultagoras:
Problem: when data is read from stream it automatically decoded in most suitable form. It works well except with strings because it create overhead. 
Solution: 
- store string as `byte[]` and convert on demand. 
- let read strings as streams (this is for the case when someone storing images in String columns)

AndrewRazumovskiy:
hello, any updates here?

PR #2911, opened 2026-07-06
[v0.10.0-rc1] Binary String Support  

chernser:
## Summary

There are a  few problems with reading binary strings: 
- they are not utf-8 strings and should not be decoded like that 
- reading big binary blobs may cause OOM because currently strings are created on read. 

This PR introduces `StringValue` that acts as `byte[]`  holder to create `String` ad-hoc instead of when read. So strings that suppose to be read via `byte[]` will stay as `byte[]`. 

This affects also JDBC where `ResultSet` has `getBinaryStream` for such cases. Now it is supported operation. 

Part of https://github.com/ClickHouse/clickhouse-java/issues/2263
## Checklist
- [x] Closes #<issue ref>


mzitnik:
@chernser, ClickHouse client benchmark results from a few days ago indicate some ⚠️ regressions. Can you take a look 
We also have failing cloud tests.

client-v2/src/main/java/com/clickhouse/client/api/data_formats/internal/AbstractBinaryFormatReader.java
```diff
@@ -866,6 +889,10 @@ public String[] getStringArray(int index) {
             BinaryStreamReader.ArrayValue array = (BinaryStreamReader.ArrayValue) value;
             if (array.itemType == String.class) {
                 return (String[]) array.getArray();
+            } else if (array.itemType == StringValue.class) {
+                StringValue[] stringValues = (StringValue[]) array.getArray();
```
mzitnik, line 893: Should we check if **stringValues != null**

client-v2/src/main/java/com/clickhouse/client/api/data_formats/internal/SerializerUtils.java
```diff
@@ -931,6 +944,69 @@ public static String convertToString(Object value) {
...
+     * reader may decode to a String, including top-level {@code String}/{@code FixedString} and wrappers such
+     * as {@code SimpleAggregateFunction(String)}). With the binary-string feature enabled the reader produces a
+     * {@link StringValue} holder (a read-time value holder, not a POJO field type); otherwise it produces a
+     * plain {@link String}. A {@code null} value (e.g. a SQL {@code NULL}) is passed through. Used from the
+     * bytecode generated by {@link #compilePOJOSetter(Method, ClickHouseColumn)}.
+     *
+     * @param value value returned by {@code BinaryStreamReader.readValue}
+     * @return the decoded string, or {@code null}
+     */
+    public static String stringValueToString(Object value) {
```
mzitnik, line 954: The naming implies that we convert StringValue to String so why our Object value is not StringValue

client-v2/src/main/java/com/clickhouse/client/api/data_formats/internal/SerializerUtils.java
```diff
@@ -931,6 +944,69 @@ public static String convertToString(Object value) {
...
+    /**
+     * Converts a value read for a string-backed column into the raw bytes for POJO binding. A {@link StringValue}
+     * yields its preserved bytes (no lossy decoding); a plain {@link String} is encoded as UTF-8, matching the
+     * historical behaviour; a value that is already a {@code byte[]} is passed through. A {@code null} value is
+     * passed through. Used from the bytecode generated by {@link #compilePOJOSetter(Method, ClickHouseColumn)}.
+     *
+     * @param value value returned by {@code BinaryStreamReader.readValue}
+     * @return the raw bytes, or {@code null}
+     */
+    public static byte[] stringValueToByteArray(Object value) {
```
mzitnik, line 971: also misleding naming

client-v2/src/main/java/com/clickhouse/client/api/data_formats/internal/SerializerUtils.java
```diff
@@ -931,6 +944,69 @@ public static String convertToString(Object value) {
...
+     * @throws IOException when failed to write to the stream
+     */
+    public static void writeFixedStringBytes(OutputStream stream, byte[] value, int length) throws IOException {
+        if (value.length > length) {
+            throw new IllegalArgumentException("Value of length " + value.length +
+                    " is longer than FixedString(" + length + ")");
+        }
+        stream.write(value);
+        for (int i = value.length; i < length; i++) {
+            stream.write(0);
```
mzitnik, line 1000: Can we write it as batch insted byte by byte

client-v2/src/main/java/com/clickhouse/client/api/data_formats/internal/StringValue.java
```diff
@@ -0,0 +1,165 @@
...
+ * <p>
+ * Backed by a {@link ByteBuffer} for a richer API and future off-heap memory support. The decoded
+ * {@link String} produced by {@link #asString()} is cached.
+ */
+public class StringValue {
+
+    /** Charset used by {@link #asString()} and {@link #toString()} when no charset is provided. */
+    public static final Charset DEFAULT_CHARSET = StandardCharsets.UTF_8;
+
+    private final ByteBuffer buffer;
```
mzitnik, line 37: why we are using ByteBuffer and not storing directly?

mzitnik: I believe performing the logic directly on byte[] can reduce memory footprint and eliminate unnecessary ByteBuffer allocations.

mzitnik (changes requested):

chernser:
@mzitnik client benchmark is not reliable in CI. I will run them locally and post results. By default we are not using BinaryStrings so it should show no regression. 

Updated: 
This is a new one. Will investigate.

client-v2/src/main/java/com/clickhouse/client/api/data_formats/internal/AbstractBinaryFormatReader.java
chernser: `BinaryStreamReader.ArrayValue` will always init array with some value - at least empty array.

client-v2/src/main/java/com/clickhouse/client/api/data_formats/internal/BinaryStreamReader.java
```diff
@@ -661,12 +678,21 @@ public ArrayValue readArrayItem(ClickHouseColumn itemTypeColumn, int len) throws
...
 
+    /**
+     * Reads a value nested inside a container (Array, Map, Tuple, Nested, Variant, JSON). Strings are always
+     * decoded into {@link String} here, regardless of the binary-string feature flag, because nested types
+     * are not expected to carry large/binary strings.
+     */
+    private Object readNestedValue(ClickHouseColumn column) throws IOException {
+        return readValue(column, null, false);
+    }
+
```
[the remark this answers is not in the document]
chernser: True. Let me create issue for this: https://github.com/ClickHouse/clickhouse-java/issues/2928

However it is expected that user can easily reduce list of fields in result set to match POJO

client-v2/src/main/java/com/clickhouse/client/api/data_formats/internal/SerializerUtils.java
chernser: This method is really used in POJO serialization and we do operate with Objects. 
this is similar to `writeDate` and other method in this file. 

PS: I previously commented that this method doesn't make sense in this way but forgot where it was used.

client-v2/src/main/java/com/clickhouse/client/api/data_formats/internal/SerializerUtils.java
chernser: yes, we can but then it will be `byte[]` allocation and then copy to a byte buffer anyway. 
this current code uses stack. 
 
Should we pay more attention to this case? 
I've created https://github.com/ClickHouse/clickhouse-java/issues/2929 to document handling of `FixedString`.

client-v2/src/main/java/com/clickhouse/client/api/data_formats/internal/StringValue.java
chernser: This is a wrapper which has good interface. I want to extend with support of off-heap buffer (but they have a small problem of converting to String because need proper charset handling).

client-v2/src/main/java/com/clickhouse/client/api/data_formats/internal/SerializerUtils.java
chernser: Similar, used in byte code compiler.

client-v2/src/main/java/com/clickhouse/client/api/data_formats/internal/BinaryStreamReader.java
```diff
@@ -242,14 +259,14 @@ public <T> T readValue(ClickHouseColumn column, Class<?> typeHint) throws IOExce
                 case Nothing:
                     return null;
                 case SimpleAggregateFunction:
-                    return (T) readValue(column.getNestedColumns().get(0));
+                    return (T) readValue(column.getNestedColumns().get(0), typeHint, false);
```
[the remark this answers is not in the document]
chernser: This is fine to read aggregate function as String - we do not expect big binary string here.

mzitnik (approved):

chernser:
@mzitnik 
regarding the benchmark tests:
- tests were run in github runner and have very unstable results. I've tested on main and they give different numbers
- I've run locally 5 iterations with 15 time window. Machine was doing only benchmark 
- local tests shown only slight diff withing 8% range (in V1 code too) 
- local tests were done with no fix and with fix but binary string turned off. 

[01-baseline.json](https://github.com/user-attachments/files/30245543/01-baseline.json)
[02-after_fix.json](https://github.com/user-attachments/files/30245555/02-after_fix.json)

image

chernser:
@mzitnik 
when binary enabled no issue found, too
image
[03-after_fix_enabled.json](https://github.com/user-attachments/files/30247285/03-after_fix_enabled.json)

PR #2914, opened 2026-07-07
Integration Spec

chernser:
## Summary

- Describes ways to integrate Client or JDBC 
- References examples and other materials

## Checklist
- [ ] Closes #<issue ref>


docs/integration-common.md
```diff
@@ -0,0 +1,103 @@
...
+
+### When to use the Java Client (recommended for new work)
+
+Choose the Java Client when you:
+
+- Build a new ingestion or analytics pipeline and control the application code
+- Need maximum read/write throughput
+- Want to work with ClickHouse **native or binary formats** (`Native`, `RowBinary`, `Parquet`, `JSONEachRow`, ...)
+- Need typed POJO serialization/deserialization
+- Process data in **column-oriented** way.
```
mshustov: for the row-oriented workload it's still more performant, isn't it?

docs/integration-index.md
```diff
@@ -0,0 +1,103 @@
...
+- Want to work with ClickHouse **native or binary formats** (`Native`, `RowBinary`, `Parquet`, `JSONEachRow`, ...)
+- Need typed POJO serialization/deserialization
+- Process data in **column-oriented** way. 
+- Require fine-grained control over compression, server settings, sessions, and operation-level configuration
+
+The Java Client exposes ClickHouse capabilities directly, with no JDBC abstraction between your code and the wire format. For metadata it offers `Client.getTableSchema(String, String)` as an equivalent to `java.sql.DatabaseMetaData`.
+
+### When to use the JDBC Driver
+
+Choose the JDBC Driver (`com.clickhouse:clickhouse-jdbc`) when you:
```
mshustov, line 66: Do we provide some documentation explaining the differences between CH-flavored JDBC and OLTP-flavored JDBC(no transactions, eventual consistency, etc)? It’s not just a matter of shifting and lifting

docs/integration-index.md
```diff
@@ -0,0 +1,103 @@
+# ClickHouse Java Integration
```
mshustov, line 1: nit: I'd rename the page to `java-integration-index` or similar, since it's an entry points in the docs. In general, it would be cool to see the page topology. are you going to move the docs on the website?

docs/integration-client.md
```diff
@@ -0,0 +1,719 @@
...
+
+---
+
+## Step 1 — Instantiation
+
+**Goal:** decide how many `Client` instances exist, how long they live, and how the internal connection pool is sized.
+
+### What a `Client` is
+
+The [`Client`](https://javadoc.io/doc/com.clickhouse/client-v2/latest/com/clickhouse/client/api/Client.html) is the single entry point for all operations. It owns:
```
mshustov, line 49: should we link a page with Java client docs?

docs/integration-client.md
```diff
@@ -0,0 +1,719 @@
...
+### Instantiation Strategy
+
+- **Share a single instance:** A single, shared `Client` instance suits most use cases. It is thread-safe and designed to be reused across your application.
+- **Long-lived lifecycle:** Build the client once at startup and close it once at shutdown. Creating a new client per request or operation is an anti-pattern because initialization takes time to set up internal structures (like the connection pool and schema cache), which adds latency to your requests.
+- **Serverless functions:** For serverless environments (like AWS Lambda), initialize the client outside the function handler so it can be reused across invocations.
+- **Warm-up (optional):** Calling `client.ping()` at startup can help initialize the connectivity part and verify the endpoint before serving live traffic, though it is not strictly required.
+- **Caching:** The application is responsible for holding the reference to the `Client` instance (e.g., via dependency injection or a singleton). The library does not provide a global static cache.
+
+
+## Step 2 — Authentication
```
mshustov, line 220: is this section specific to a partner building an integration? Should we consider moving it to the Java client docs?

docs/integration-client.md
```diff
@@ -0,0 +1,719 @@
...
+|--------------|-----|
+| [`RowBinaryFormatWriter`](../client-v2/src/main/java/com/clickhouse/client/api/data_formats/RowBinaryFormatWriter.java) | Row-oriented binary insert |
+| [`ClickHouseBinaryFormatWriter`](../client-v2/src/main/java/com/clickhouse/client/api/data_formats/ClickHouseBinaryFormatWriter.java) | General binary insert |
+
+### How to choose
+
+| Goal | Recommended format |
+|------|--------------------|
+| Maximum read throughput | `Native` or `RowBinaryWithNamesAndTypes` |
+| Maximum write throughput | `RowBinary` or `Native` |
```
mshustov: Do we already support `Native` for writing?

docs/integration-client.md
```diff
@@ -0,0 +1,719 @@
...
+| HTTP compression | `InsertSettings.useHttpCompression(true)` | Sets Content-Encoding HTTP header |
+| Pre-compressed data | `InsertSettings.appCompressedData(true, "gzip")` | Informs the client you are providing already-compressed data |
+| Copy buffer size | `InsertSettings.setInputStreamCopyBufferSize(n)` | Stream-to-stream copy buffer size |
+
+> **Server settings**:
+> Server settings control how ClickHouse processes the ingested data. Pass them via `InsertSettings.serverSetting(key, value)`. For example, `serverSetting("async_insert", "1")` enables server-side buffering, and `serverSetting("wait_for_async_insert", "1")` determines whether the client waits for the flush.
+
+### Idempotency — deduplication token
+
+On a `MergeTree`-family table with deduplication enabled, the **`insert_deduplication_token`** lets the server skip duplicate blocks so retries do not create duplicate rows.
```
mshustov, line 777: let's link official docs for the setting and dedup logic

docs/integration-client.md
```diff
@@ -0,0 +1,719 @@
...
+
+**On built-in retries.** `ServerException.isRetryable()` marks transient server codes (timeouts, network, memory-limit, too-many-parts, ...) and the client's own `retry` policy already re-sends those. Do not add a second retry loop on top without accounting for it. Retries behave very differently for reads vs writes — see the per-operation "Errors & how to handle them" sections for the details (in particular, an insert retry must re-send the whole payload).
+
+</error-handling-rules>
+
+---
+
+## Quick reference
+
+A minimal end-to-end program. `Client` implements `AutoCloseable`, so try-with-resources handles shutdown.
```
mshustov: should it live in examples?

docs/integration-jdbc.md
```diff
@@ -0,0 +1,611 @@
...
+### Init configuration — pool sizing and threading
+
+Because each `Connection` wraps a `Client` with its own HTTP pool, the client-level `max_open_connections` (default 10) forwards through. Two layers of pooling exist:
+
+- **JDBC connection pool** (your responsibility) — controls how many `Connection` objects (and thus `Client` instances) exist.
+- **HTTP pool per connection** — `max_open_connections` forwarded to `ClientConfigProperties`.
+
+### Common Pitfalls
+
+<common-pitfalls>
```
mshustov, line 255: it's not a standard markdown, is it?
do we expect a human or an agent to read them? 
for the agent, I'd consider linking from `AGENT.md` or creating a dedicated skill

docs/integration-jdbc.md
```diff
@@ -0,0 +1,611 @@
+# ClickHouse JDBC Integration Guide
```
mshustov, line 1: I'd extend the doc with a testing instruction: how to setup env, run a DB instance locally, etc

docs/integration-client.md
```diff
@@ -0,0 +1,719 @@
...
+```
+
+- Assign a **stable** token per logical batch (file name, Kafka offset, job ID).
+- Use it for retry-safe pipelines and at-least-once sources (Kafka, SQS, file reprocessing).
+- Requires a `MergeTree` engine with deduplication configured. See [`InsertTests.testInsertSettingsDeduplicationToken`](../client-v2/src/test/java/com/clickhouse/client/insert/InsertTests.java).
+
+### Best practices
+
+<best-practices>
+- **Batch large** — thousands to millions of rows per request, never one row per request.
```
mshustov: We can recommend checking `query_log` to understand the server-side performance implications

docs/integration-client.md
```diff
@@ -0,0 +1,719 @@
...
+
+**`connection_pool_enabled` is not recommended to change.** Leave pooling **enabled** (the default). The only reason to disable it is a workload with extreme concurrency where the internal Apache HttpClient connection pool itself becomes a contention point — under very high parallel request rates the pool's own bookkeeping can serialize threads. Disabling pooling avoids that bottleneck at the cost of a fresh connection per operation (higher latency, more sockets), so treat it as a last resort after measuring, not a default.
+
+### Common Pitfalls
+
+<common-pitfalls>
+- **CONSTRAINT:** Do not create a `Client` per request. It destroys pool warm-up and adds latency on every call — the single most common mistake.
+- **Pool too small** (`max_open_connections`) throttles concurrency; size it to peak concurrent operations, not average.
+- **CONSTRAINT:** Always `close()` the client at shutdown to avoid leaking the pool and its threads.
```
mshustov: is it a graceful shutdown?

docs/integration-client.md
```diff
@@ -0,0 +1,719 @@
...
+                throw new RuntimeException("ClickHouse unreachable");
+            }
+
+            // Step 7: write (blocks on the returned future)
+            try (InputStream data = QuickStart.class.getResourceAsStream("/events.jsonl")) {
+                client.insert("my_table", data, ClickHouseFormat.JSONEachRow).get();
+            }
+
+            // Step 6: read a small result
+            List<GenericRecord> rows = client.queryAll("SELECT count() FROM my_table");
```
mshustov: I'd consider adding it as an explicit verification step

docs/integration-client.md
```diff
@@ -0,0 +1,719 @@
...
+</common-pitfalls>
+---
+
+## Step 4 — Connections Configuration
+
+In the Java Client a "connection" is an **HTTP connection borrowed from the internal pool**, not a long-lived database session. Each operation borrows a connection, sends a request, streams the response, and returns the connection to the pool.
+
+### Connection limit (`max_open_connections`)
+
+The pool size depends on your workload — specifically on its **concurrency**, not on how much data it moves. What matters is **how many operations run at the same time**, not the number of rows or bytes any single operation transfers. A pool of 20 connections serves at most 20 simultaneous operations regardless of whether each returns one row or a million. This is the single setting you actually tune. The table below is guidance, not a hard rule: measure under your own load.
```
mshustov: > measure under your own load

How? inspect `query_log`? enable logging / tracing / whatever? use JMH harness?

docs/integration-client.md
```diff
@@ -0,0 +1,719 @@
...
+
+<common-pitfalls>
+- **CONSTRAINT:** Do not create a `Client` per request. It destroys pool warm-up and adds latency on every call — the single most common mistake.
+- **Pool too small** (`max_open_connections`) throttles concurrency; size it to peak concurrent operations, not average.
+- **CONSTRAINT:** Always `close()` the client at shutdown to avoid leaking the pool and its threads.
+
+</common-pitfalls>
+---
+
+## Step 5 — Data formats, readers & writers
```
mshustov, line 457: do we need CH <> Java data type mapping?

docs/integration-client.md
```diff
@@ -0,0 +1,719 @@
...
+    .setDefaultDatabase("analytics")
+    .build();
+```
+
+### Instantiation Strategy
+
+- **Share a single instance:** A single, shared `Client` instance suits most use cases. It is thread-safe and designed to be reused across your application.
+- **Long-lived lifecycle:** Build the client once at startup and close it once at shutdown. Creating a new client per request or operation is an anti-pattern because initialization takes time to set up internal structures (like the connection pool and schema cache), which adds latency to your requests.
+- **Serverless functions:** For serverless environments (like AWS Lambda), initialize the client outside the function handler so it can be reused across invocations.
+- **Warm-up (optional):** Calling `client.ping()` at startup can help initialize the connectivity part and verify the endpoint before serving live traffic, though it is not strictly required.
```
mshustov: should we mention that a Cloud instance might be idling and ping helps to wake it up?

docs/integration-client.md
mshustov: do we have any docs for choosing "right size"? there should be a disclaimer about latency vs part count vs memory trade-off

docs/integration-jdbc.md
```diff
@@ -0,0 +1,611 @@
...
+
+**3. Switch to the Java Client** for per-insert token control via `InsertSettings.setDeduplicationToken(...)`.
+
+See [integration-client.md — deduplication token](integration-client.md#idempotency--deduplication-token) for semantics and requirements.
+
+### Best practices
+
+<best-practices>
+- **Batch inserts** — hundreds to thousands of rows per batch.
+- **Enable the RowBinary beta** for simple inserts — significantly faster than SQL text rendering.
```
mshustov, line 630: what is `beta`?

docs/integration-client.md
```diff
@@ -0,0 +1,719 @@
+# ClickHouse Java Client Integration Guide
```
mshustov, line 1: I'd propose adding: o11y wiring, graceful degradation implementation, capacity planning (hello buffer size) , client upgrade strategy

mshustov (commented):
for partners, I'd consider adding recommended "acceptance criteria" as a separate document (can be done in a follow-up)
my braindump: 
- validate functional correctness: number of rows = intended, data type conversion in a roundtrip, handles schema behaviour gracefully (type mismatch, missing column, extra column, etc)
- failure path: idempotent insert, handles retryable (network) and non-retryable errors, handles failures for long operations (1h+)
- no memory leaks under "long process" or bursty workload
- test against their own latency vs thoroughput vs parts count criteria
- operational readiness: configures client_name, can enable logging
- cloud readiness: handles idle instances, long query (1h+) failures, etc
- has auto tests against supported CH versions

docs/integration-client.md
chernser: Actually very good idea. Also updated project spec.

docs/integration-client.md
chernser: I think, it is more in reference section. I do want to reduce noise here.

docs/integration-client.md
chernser: added.

docs/integration-client.md
chernser: We will contribute some parts to main doc. 
We should cover all features as much as possible because we do not know next partner case. 
Mainly we need to pass all specific details.

docs/integration-client.md
chernser: I'd expect running load tests + doing estimation. But yes, we need to clarify.

docs/integration-client.md
chernser: yes. We need to do something about ungraceful end. But it will be subject for next update.

docs/integration-client.md
chernser: yes. But I have not worked on this step, yet.

docs/integration-client.md
chernser: Yes, but read only.

docs/integration-client.md
chernser: added.

docs/integration-client.md
chernser: ok. Will rework examples.

docs/integration-client.md
chernser: Sorry, I think I did not understand the question.
I will add more examples for sure.

docs/integration-index.md
chernser: renamed.
I hope it will be on the website =) 
But mainly this document is for AI to get as MD file.

docs/integration-common.md
chernser: yes. added.

docs/integration-jdbc.md
chernser: yes, this is recommendation for AI reader. But looks broken so I have removed it.

docs/integration-jdbc.md
chernser: this is mistake. Will correct. Need more work on JDBC part.

docs/integration-client.md
mshustov: that's why I'm asking. ` Maximum write throughput | `RowBinary` or `Native`  - implies `native` is supported for writing

docs/integration-client.md
mshustov: the question was whether examples should be an inlined vs in a separate runnable folder

docs/integration-client.md
```diff
@@ -0,0 +1,711 @@
...
+    // Database credentials (e.g. Option A)
+    .setUsername("default")
+    .setPassword("secret")
+    // Proxy configuration and credentials
+    .addProxy(ProxyType.HTTP, "proxy.example.com", 8080)
+    .setProxyCredentials("proxy_user", "proxy_password")
+    .build();
+```
+
+### Client Identity & Default Database
```
laeg, line 329: This is a great start, Sergey! We should elevate this from sounding optional to "should be configured":

We should also talk about the convention and how it will appear in the document, e.g.:

`partner-name/partner-version ClickHouseJavaClient/version (runtime-info)` lets make it match what we have documented in notion.

We should also add a verification step for partners: 

```
SELECT http_user_agent, count()
FROM system.query_log
WHERE http_user_agent LIKE '%your-name%'
GROUP BY http_user_agent
```

We can then explain why this matters support/triaging issues and user attribution

docs/integration-jdbc.md
```diff
@@ -0,0 +1,611 @@
...
+- **Default type mappings** may not match your expectations for large integers or complex types — override with `jdbc_type_mappings`.
+
+</common-pitfalls>
+---
+
+## JDBC-specific features
+
+| Feature | How to use |
+|---------|------------|
+| Application name | `Connection.setClientInfo("ApplicationName", "my-app")` — surfaces in query log |
```
laeg: see other comment on setting client information

docs/integration-jdbc.md
```diff
@@ -0,0 +1,611 @@
...
+| Schema / database | `Connection.setSchema("analytics")` or the URL path |
+| Query cancellation | `Statement.cancel()` → `KILL QUERY` (optionally `ON CLUSTER` via `jdbc_cluster_name`) |
+| JDBC escape syntax | `{ts '...'}`, `{d '...'}`, `{fn ...}` — translated before execution |
+| Default query settings | `default_query_settings` property |
+| Role management | `SET ROLE` statements (roles remembered by default) |
+| Framework detection | Automatic — logged at connection time (Spark, Flink, NiFi) |
```
laeg: Can we expand this a bit more so that partners can understand:
* How does this detection work?
* What exactly gets logged and where
* How does a partner get their framework added to the auto-detect list if appplicable
* does manual setting override auto-detect

docs/integration-jdbc.md
laeg: Lets add step 8 in with:
* Docker setup for local clickhouse
* cloud environment - partners get credits to use 
* reference to example test suites in our repo

we should make sure to highlight areas like:
* functional correctness: row count matches, data types, schema mismatch handling
* potential failure paths: idempotent inserts, retryable vs non-retryable, long operation failures
* latency vs throughput vs part count
* client_name, logging enabled
* cloud readiness: idle instance handling, long running query failures
* automated testing against the different channels in CH cloud for version support

docs/integration-client.md
laeg: Great work on the error handling and retry guidance! same as my other suggestion for a new step 10.

We should focus on how to verify if they've implemented patterns correctly.

we should also call out where to find the change logs, how to test before upgrading etc

docs/integration-client.md
```diff
@@ -0,0 +1,711 @@
...
+    .build();
+```
+
+Rotate credentials at runtime (thread-safe, non-blocking, applies to newly started requests):
+
+```java
+client.updateUserAndPassword("new_user", "new_password");
+```
+
+### Option B — Token / bearer
```
laeg, line 253: We should probably call out how CH cloud authentication flow works and point people in that direction. How partners will need both layers: driver and cloud-env.

minimum: call out box noting that cloud oauth documentation is located somewhere and that `useBearer...` and `updateBearer...`  are the mechanisms for passing cloud-issues tokens.

docs/integration-client.md
```diff
@@ -0,0 +1,711 @@
...
+These are the only pool-related properties you normally touch; the rest have safe defaults.
+
+| Property (Builder method) | Purpose | Default |
+|---------------------------|---------|---------|
+| `max_open_connections` (`.setMaxConnections()`) | Pool size — set from the parallelism guidance above | 10 |
+| `connection_pool_enabled` (`.enableConnectionPool()`) | Enable/disable pooling (keep enabled) | true |
+| `connection_ttl` (`.setConnectionTTL()`) | Max lifetime of a pooled connection | — |
+| `connection_reuse_strategy` (`.setConnectionReuseStrategy()`) | FIFO or LIFO reuse | — |
+
+**`connection_ttl` against ClickHouse Cloud.** Keep it **relatively small** when the endpoint is a Cloud (or otherwise load-balanced) deployment. A short TTL forces connections to be retired and re-established frequently, so new connections keep going through the load balancer, which lets it **redistribute traffic across nodes** instead of pinning long-lived connections to whichever node they first landed on.
```
laeg, line 444: This is a great practical callout. I think it should be expanded to:

* Auto-upgrades - might want to talk about server settings and pinning compatibility, along with how to test against new version (fast channel)
* Idle instance wake up latency - we mention the this above but can we quantify the latency impact or at least give folks information to use as part of the dialog (its often good to add a note to users explaining first query after idle usage e.g. "note: first queries after an idle period can have a delay" along with telling them to implement warm up strategies or other ux strategies)

docs/integration-client.md
```diff
@@ -0,0 +1,711 @@
...
+    .setDefaultDatabase("analytics")
+    .build();
+```
+
+### Instantiation Strategy
+
+- **Share a single instance:** A single, shared `Client` instance suits most use cases. It is thread-safe and designed to be reused across your application.
+- **Long-lived lifecycle:** Build the client once at startup and close it once at shutdown. Creating a new client per request or operation is an anti-pattern because initialization takes time to set up internal structures (like the connection pool and schema cache), which adds latency to your requests.
+- **Serverless functions:** For serverless environments (like AWS Lambda), initialize the client outside the function handler so it can be reused across invocations.
+- **Warm-up (optional):** Calling `client.ping()` at startup can help initialize the connectivity part and verify the endpoint before serving live traffic, though it is not strictly required. It may also require to wakeup cloud instance.
```
laeg, line 83: Can we show an example for pinging an idle instance?

Do we need to show an example pattern for a readiness probe that accounts for wake-up (like having a retry-backoff upon cant connect exception after going through a wake up scenario)

docs/integration-client.md
```diff
@@ -0,0 +1,711 @@
...
+
+| Setting | Property / method | Notes |
+|---------|-------------------|-------|
+| Client request compression | `InsertSettings.compressClientRequest(true)` | LZ4-compress the insert body before sending |
+| HTTP compression | `InsertSettings.useHttpCompression(true)` | Sets Content-Encoding HTTP header |
+| Pre-compressed data | `InsertSettings.appCompressedData(true, "gzip")` | Informs the client you are providing already-compressed data |
+| Copy buffer size | `InsertSettings.setInputStreamCopyBufferSize(n)` | Stream-to-stream copy buffer size |
+
+> **Server settings**:
+> Server settings control how ClickHouse processes the ingested data. Pass them via `InsertSettings.serverSetting(key, value)`. For example, `serverSetting("async_insert", "1")` enables server-side buffering, and `serverSetting("wait_for_async_insert", "1")` determines whether the client waits for the flush.
```
laeg, line 773: Because of our changes to `async_insert=0` and `wait_end_of_query=0` in `0.10.0-rc1` - Do we need to point people to understanding the complexity of these settings, like row counts might not reflect written data for async and wait for async trades throughput for accuracy?

docs/integration-jdbc.md
laeg: @chernser, let's chat about this more, and if we need to clarify that this had the potential to change if it's beta?

docs/integration-client.md
```diff
@@ -0,0 +1,711 @@
...
+
+- **Standard artifact** (default, no classifier) — the client together with its dependencies declared as ordinary transitive Maven dependencies. Recommended for managed builds in which the application controls the dependency tree.
+- **Shaded artifact** (`all` classifier) — a single self-contained archive that bundles and **relocates** most third-party dependencies (Apache HttpClient, LZ4, RoaringBitmap, ASM, and others) under `com.clickhouse.shaded.*`. Recommended when transitive dependencies cannot be managed, such as self-contained deployment archives or standalone tooling.
+
+**Dependency conflicts to anticipate.** The standard artifact introduces libraries that the application may already depend on at different versions — notably **Guava**, **Apache HttpClient 5**, and **commons-compress**. If the application declares incompatible versions, `NoSuchMethodError` or `LinkageError` may occur at runtime. Two approaches resolve this:
+
+- Use the shaded artifact so that these dependencies are relocated and cannot collide. Note that `slf4j` and `micrometer` are intentionally left unshaded, so logging and metrics continue to bind to the application's own implementations.
```
laeg, line 20: i think this points back to @mshustov's comment on client o11y - we should inform people how to observe what's happening:
* How to configure SLF4J for client-side logging (point to docs if we have it)
* How to inspect `system.query_log` for server-side performance
* OTEL trace context propagation
Micrometer metrics integrations

docs/integration-index.md
laeg: We have a few options here:
1. Agent.md in repo root references these docs
2. Skills that loads these as context
3. Both of the above

If we're expecting AI agents to consume these docs, we might want to aim more for self contained sections with clear decision trees over long prose that require reading the whole document for context

laeg (changes requested):

docs/integration-client.md
chernser: we should have inline examples because reading documentation and jumping to examples is not very productive.

docs/integration-jdbc.md
```diff
@@ -0,0 +1,611 @@
...
+
+| # | Milestone | Core decision |
+|---|-----------|---------------|
+| 1 | [Instantiation strategy](#step-1--instantiation-strategy) | Connection lifecycle and which external pool to use |
+| 2 | [Authentication](#step-2--authentication) | Which auth mechanism and how to configure it via URL/Properties |
+| 3 | [Transport & connectivity](#step-3--transport--connectivity-tls-proxies-timeouts) | TLS/mTLS, proxies, timeouts, health checks |
+| 4 | [Formats under the hood](#step-4--formats-under-the-hood) | What the driver does internally; when JDBC is not enough |
+| 5 | [Read operations & tuning](#step-5--read-operations--tuning) | `ResultSet` streaming; type mapping; heavy-read tuning |
+| 6 | [Write operations & tuning](#step-6--write-operations--tuning) | Batch vs RowBinary beta; heavy-ingest tuning; idempotency |
+| 7 | [Metadata & schema discovery](#step-7--metadata--schema-discovery) | `DatabaseMetaData`, `ResultSetMetaData`, type mapping |
```
mshustov, line 43: Could you add a path for `readonly` user? It's a common patter for BI tools

mshustov: other BI related questions 
-  to how to use [final](https://clickhouse.com/docs/reference/statements/select/from#final-modifier) without modifying a query 🤔  
- how to configure JDBC driver via JDBC URL

I'd suggest adding a BI integrations as separate chapter

mshustov: can be done in a separate follow-up PR

docs/integration-client.md
chernser: Done. corrected.

docs/integration-jdbc.md
chernser: ok. recorded.

docs/integration-index.md
chernser: @laeg 
- agent.md - becomes part of every prompt so we need to keep only very specific things.
- skills - this is for recurring  tasks. Current spec is for building something complete and not expected to repeat very often

This spec will be read by AI or human before writing any code. We might need to think how we can bring updates to this to let users migrate to new features.

chernser: Added to the index : 
```

Reference information should be fetched from official documentation for [Java Client](https://clickhouse.com/docs/integrations/language-clients/java/client) or [JDBC Driver](https://clickhouse.com/docs/integrations/language-clients/java/jdbc). 

This set of documents can be used to one's code over time. Keep checking your implementation at least per release as we going to add more information and provide migration guidance.  

```

docs/integration-client.md
chernser: Added to index page that reference documentation is official one for client and jdbc

docs/integration-client.md
chernser: I will move it to separate document because it is separate task. I see it like this: 
1. spec used to create implementation of initial version functionally correct 
2. tests are implemented with another spec (user provided) 
3. results analyzed with help of ClickHouse skills (look into query_log, check metrics, etc)

docs/integration-jdbc.md
chernser: this is typo - not what I want to mention in the doc. I will correct it.

docs/integration-client.md
chernser: Good point @laeg

docs/integration-client.md
chernser: yes, I think. will add.

docs/integration-jdbc.md
chernser: We need to think if we need to defined step with env setup - it may be not a concern at the stage. 

regarding areas I agree and we mostly highlight it. Some additions will come further.

docs/integration-client.md
chernser: agree. will add in next iteration.

docs/integration-client.md
chernser: yes we have example of ping. will extend in next iteration of the document. but it is quite simple thing that may be learned from main doc.

docs/integration-jdbc.md
chernser: we need to remove this. there is no need to detect framework for partners.

docs/integration-client.md
chernser: added

docs/integration-index.md
chernser: done

docs/integration-client.md
chernser: This will be done later for JWT example

docs/integration-client.md
chernser: Captured - will do in next update

laeg (approved):
approved with follow ups

PR #2920, opened 2026-07-10
fix: ensure nullable binary arrays return boxed type arrays instead of Object[]

### LINKED ISSUE #2846: BinaryStreamReader.ArrayValue wrong return type
## Description
Current implementation of BinaryStreamReader.readArrayItem(ClickHouseColumn itemTypeColumn, int len)
doesn't return correct expected type as it was in client v1.

Current implementation
```
 if (itemTypeColumn.isNullable() || itemTypeColumn.getDataType() == ClickHouseDataType.Variant) {
            array = new ArrayValue(Object.class, len);
            for (int i = 0; i < len; i++) {
                array.set(i, readValue(itemTypeColumn));
            }
        } 
```

The logic here is wrong unless it was done on purpose.
We should add another condition to return an array with the correct type not Object.class

```
if (itemTypeColumn.isNullable() || itemTypeColumn.getDataType() == ClickHouseDataType.Variant) {
            ClickHouseDataType type = column.getDataType();
            array = new ArrayValue( type.getPrimitiveClass().isPrimitive()?Object.class: type.getPrimitiveClass(), len);
            for (int i = 0; i < len; i++) {
                array.set(i, readValue(itemTypeColumn));
            }
        } 
```

### Expected Behaviour
when we have a nullable non primitive type (example NullableDouble array) we should return Double[] instead of Object[]

#### Environment
* Client version: 0.9.8

kishansinghifs1:
## Summary

BinaryStreamReader.readArrayItem used Object.class as the element type for all nullable array columns, so Array(Nullable(Float64)) returned Object[] instead of Double[]. This splits the nullable check from the Variant/Dynamic/Geometry fallback and derives the correct boxed wrapper class via getObjectClass(). Added unit test to verify the component type is Double.class for Array(Nullable(Float64)).

Closes  :  https://github.com/ClickHouse/clickhouse-java/issues/2846

## Checklist
- [x] Closes #2846 


chernser:
@kishansinghifs1 

I will review next week.

kishansinghifs1:
Hiii @chernser,
thank you for the opportunity ,
no worries.

chernser (approved):

Issue #2715, opened 2026-01-19
[jdbc-v2] [output format] SQL Error [HY000]: Only RowBinaryWithNameAndTypes is supported for output format. Please check your query.


ikfid:
## Description
Migrating from clickhouse jdbc driver 0.6.5 to clickhouse jdbc v2 driver 0.9.6, some queries stopped working.

### Steps to reproduce
1. `SELECT 1 as x FORMAT JSON`
### Error Log or Exception StackTrace

```
Exception in thread "main" java.sql.SQLException: Only RowBinaryWithNameAndTypes is supported for output format. Please check your query.
	at com.clickhouse.jdbc.StatementImpl.executeQueryImpl(StatementImpl.java:164)
	at com.clickhouse.jdbc.PreparedStatementImpl.executeQuery(PreparedStatementImpl.java:144)
	at ClickHouseTest.main(ClickHouseTest.java:18)
```

Clickhouse server does not produce any error, and 'clickhouse client' executes the above query without issues, therefore, I assume, there is some issue at the jdbc client side.

### Expected output
```
{                     
 "meta":              
 [                    
  {                   
   "name": "x",       
   "type": "UInt8"    
  }                   
 ],                   
                      
 "data":              
 [                    
  {                   
   "x": 1             
  }                   
 ],                   
                      
 "rows": 1,           
                      
 "statistics":        
 {                    
  "elapsed": 0.001607,
  "rows_read": 1,     
  "bytes_read": 1     
 }                    
}                     
```

### Sample code
implementation "com.clickhouse:clickhouse-jdbc-all:0.9.6"

```
import javax.sql.DataSource;

import java.sql.Connection;
import java.sql.PreparedStatement;
import java.sql.ResultSet;
import java.util.Properties;

public class ClickHouseTest {

    public static void main(String[] args) throws Exception {
        Properties properties = new Properties();
        properties.setProperty("user", "default");
        properties.setProperty("password", "");

        DataSource dataSource = new com.clickhouse.jdbc.DataSourceImpl("jdbc:clickhouse://127.0.0.1:8123", properties);
        Connection connection = dataSource.getConnection();
        PreparedStatement stmt = connection.prepareStatement("SELECT 1 as x FORMAT JSON");
        ResultSet rs = stmt.executeQuery();
        while (rs.next()) {
            String json = rs.getString(1);
            System.out.println(json);
        }
        stmt.close();
        connection.close();
    }
}

```

#### Environment
[...]

PR #2927, opened 2026-07-15
fix(jdbc-v2): support explicit FORMAT JSON queries

014-code:
## Summary

- support explicit `FORMAT JSON` queries in JDBC v2
- expose the server-rendered JSON through a streaming, single-column
  `ResultSet` named `results`
- return one JDBC row for each line in the JSON response
- preserve the existing `FORMAT JSONEachRow` and default
  `RowBinaryWithNamesAndTypes` paths
- update the JDBC v2 feature contract and changelog

## Problem

JDBC v2 rejects queries with an explicit JSON output format:

## Issue

Related #2715.

After migrating from the legacy JDBC driver to JDBC v2, queries that explicitly
request ClickHouse's `JSON` output format fail in `executeQuery()`:

```sql
SELECT 1 AS x FORMAT JSON

### CODE REVIEW

**Comment by chernser** [comment]
@014-code 

thank you for the contribution! 

However we plan to support JSON format in another way because it has metadata and data in one document. 
I've commented in the original issue. 
Current problem is to make some call with internal client and get result. But for this `com.clickhouse.jdbc.ConnectionImpl#getClient` can be used.

**Comment by 014-code** [comment]
> @014-code
> 
> thank you for the contribution!
> 
> However we plan to support JSON format in another way because it has metadata and data in one document. I've commented in the original issue. Current problem is to make some call with internal client and get result. But for this `com.clickhouse.jdbc.ConnectionImpl#getClient` can be used.

Thanks for the clarification.

I understand. I'll take a look at implementing it using ConnectionImpl#getClient() instead of handling the HTTP response directly.

File: docs/client-v2-json-support.md [hunk 1]
```diff
@@ -352,6 +352,34 @@ Notes:
...
+the JDBC connection and consume the `QueryResponse` stream directly.
+
+```java
+try (Connection conn = DriverManager.getConnection(
+        "jdbc:clickhouse://localhost:8123/default", props);
+     QueryResponse response = conn.unwrap(ConnectionImpl.class)
+             .getClient()
+             .query("SELECT 1 AS x FORMAT JSON")
+             .get();
+     BufferedReader reader = new BufferedReader(new InputStreamReader(
```
chernser: This is very inefficient way of parsing JSON - every JSON library supports reading from input stream. 
Please provide example where it is demonstrated for example with Jackson library. 
Besides JSON format (https://clickhouse.com/docs/reference/formats/JSON/JSON) has meta and data separated - so users need some instructions how to handle them.

docs/client-v2-json-support.md
```diff
@@ -352,6 +352,34 @@ Notes:
 
 ## Usage in `jdbc-v2`
 
+### Document-oriented JSON output
```
chernser: It is not document oriented, I'd say. It is just JSON format output with very ClickHouse specific structure https://clickhouse.com/docs/reference/formats/JSON/JSON

jdbc-v2/src/test/java/com/clickhouse/jdbc/ConnectionTest.java
```diff
@@ -921,11 +926,31 @@ public void testUnwrapping() throws Exception {
...
 
+    @Test(groups = { "integration" })
+    public void testRawJSONQueryThroughUnderlyingClient() throws Exception {
+        try (Connection conn = getJdbcConnection();
+             QueryResponse response = conn.unwrap(ConnectionImpl.class).getClient()
+                     .query("SELECT 1 AS x FORMAT JSON")
+                     .get();
+             BufferedReader reader = new BufferedReader(
+                     new InputStreamReader(response.getInputStream(), StandardCharsets.UTF_8))) {
+            assertEquals(response.getFormat(), ClickHouseFormat.JSON);
```
chernser, line 942: we use tests as reference implementation sometimes. This test do not fit this needs. 
Please align it with example in the document.

chernser (changes requested):

014-code:
Thank you for the clarification. I’ve updated the PR accordingly in 2636d7e5.
Reworded the documentation to describe FORMAT JSON as ClickHouse-specific JSON output rather than document-oriented output.
Replaced the line-by-line string collection with a Jackson example that parses QueryResponse#getInputStream() directly.
The example and integration test now handle the meta and data fields separately, including column metadata and row values.
Added a note to use Jackson’s streaming JsonParser for large data arrays.
The JDBC ResultSet behavior remains unchanged; this only documents and verifies access through ConnectionImpl#getClient() as discussed.

chernser (approved):

PR #2986, opened 2026-07-30
Close underlying stream in ClickHouseLZ4InputStream

### LINKED ISSUE #2985: ClickHouseLZ4InputStream does not close underlying stream
### Description

Closing a `ClickHouseLZ4InputStream` does not close its underlying input stream. Its `close()` implementation calls `InputStream.close()`, which is a no-op.

The response path wraps the HTTP entity stream in this class when native ClickHouse LZ4 compression is enabled. As a result, callers that close the stream returned by `QueryResponse.getInputStream()` may leave the underlying HTTP response stream open, particularly when they stop reading before the end of the response.

### Expected behavior

Closing `ClickHouseLZ4InputStream` should close the input stream it wraps, matching standard filtering stream behavior and the other compression wrappers in the client.

fallintoplace:
## Summary

`ClickHouseLZ4InputStream.close()` currently calls `InputStream.close()`, which is a no-op, so closing a native LZ4 response stream does not release its underlying HTTP response stream.

Delegate `close()` to the wrapped input stream and add a focused unit test for the lifecycle contract. This aligns native LZ4 responses with the other compression wrappers and allows partially consumed responses to release transport resources when callers close the stream.

Closes #2985

## Compatibility

No API signatures, compressed data handling, or fully consumed response behavior change. The only behavior change is that closing the wrapper now closes the resource it owns.

## Checklist

- [x] Closes #2985

## Testing

- `mvn -pl client-v2 -Dtest=ClickHouseLZ4InputStreamTest test`
- `TZ=UTC mvn -pl client-v2 test` (530 tests)


client-v2/src/test/java/com/clickhouse/client/api/internal/ClickHouseLZ4InputStreamTest.java
```diff
@@ -0,0 +1,24 @@
...
+import org.mockito.Mockito;
+import org.testng.annotations.Test;
+
+import java.io.IOException;
+import java.io.InputStream;
+
+public class ClickHouseLZ4InputStreamTest {
+
+    @Test(groups = {"unit"})
+    public void testCloseClosesUnderlyingStream() throws IOException {
```
chernser: It should be an integration test.
Current test is good only for coverage but it doesn't test suggested in the issue scenario.

client-v2/src/main/java/com/clickhouse/client/api/internal/ClickHouseLZ4InputStream.java
```diff
@@ -177,6 +177,6 @@ static void setInt32(byte[] bytes, int offset, int value) {
 
     @Override
     public void close() throws IOException {
-        super.close();
+        in.close();
```
chernser, line 180: this will close response stream however we technically can read body twice because we create new wrapping stream. 
So there should be some documentation for QueryResponse explaining how it works.

CHANGELOG.md
```diff
@@ -44,6 +44,9 @@
 
 ### Bug Fixes 
 
+- **[client-v2]** Fixed native LZ4 response streams not closing their underlying HTTP response stream. Closing the
```
chernser: This is not native LZ4 response - it is just LZ4 input stream. Word `native` is very confusing here because we have native protocol and native format.

CHANGELOG.md
```diff
@@ -44,6 +44,9 @@
...
+  stream returned by `QueryResponse.getInputStream()` now releases the wrapped transport stream, including when a
+  response is only partially read. (https://github.com/ClickHouse/clickhouse-java/issues/2985)
```
chernser: when response is partially read nothing makes input stream to be closed - so this is not true.

chernser (changes requested):
see comments.

chernser (approved):

PR #3046, opened 2026-08-08
fix(jdbc-v2): render array contents for named tuples

### LINKED ISSUE #3045: Regression in JDBC 0.9.8 when reading Array(Tuple(...)) with named tuple elements.
## Description
Regression in JDBC 0.9.8 when reading Array(Tuple(...)) with named tuple elements.

With the official ClickHouse JDBC driver 0.9.8, IntelliJ IDEA displays Array(Tuple(id UUID)) values as:
`com.clickhouse.jdbc.types.Array@54c3b772`
instead of the array contents.

The same query works correctly with JDBC 0.6.0.
The issue only affects named tuples. Arrays of unnamed tuples are displayed correctly.

### Steps to reproduce
1.    Connect to ClickHouse using clickhouse-jdbc:0.9.8.
2.  Execute `SELECT [('550e8400-e29b-41d4-a716-446655440000')];` Result is displayed correctly.
3.  Execute `SELECT [('550e8400-e29b-41d4-a716-446655440000')::Tuple(id UUID)];` Result is displayed as com.clickhouse.jdbc.types.Array@...
No exception is thrown.

The value is rendered as:
```
com.clickhouse.jdbc.types.Array@54c3b772
```

### Expected Behaviour
Array(Tuple(id UUID)) should be returned through JDBC in a way that clients can display its contents, the same as in JDBC 0.6.0 and the same as Array(Tuple(UUID))

### Code Example

```sql
SELECT [('550e8400-e29b-41d4-a716-446655440000')];

SELECT [('550e8400-e29b-41d4-a716-446655440000')::Tuple(id UUID)];

```

#### Client Configuration
Default JDBC configuration.
```java

```

#### Environment
* [ ] Cloud
* Client version:    clickhouse-jdbc 0.9.8
* Language version:     N/A (observed in IntelliJ Database Viewer)
* OS:     macOS
* IDE: IntelliJ IDEA 2026.2.0.1 

#### ClickHouse Server
* ClickHouse Server version: 26.3.17.56
* ClickHouse Server non-default settings, if any: None
* `CREATE TABLE` statements for tables involved:     Not required (reproducible with literal values only)

It is unclear whether this is a JDBC driver regression or an incompatibility with IntelliJ’s database viewer, but the behavior changed between JDBC 0.6.0 and 0.9.8 and is reproducible with the SQL above.

014-code:
## Summary

- Make JDBC `Array` values display their nested contents.
- Support readable rendering of `Array(Tuple(...))` values with named tuple elements.
- Preserve the existing `java.sql.Array` return type and `getArray()` behavior.
- Add unit and integration regression tests.
- Update `CHANGELOG.md`.

## Problem

When JDBC v2 returned an array containing a named tuple, clients such as IntelliJ IDEA could display the value as:

com.clickhouse.jdbc.types.Array@54c3b772

For example:

SELECT [('550e8400-e29b-41d4-a716-446655440000')::Tuple(id UUID)];

The actual contents were available through `java.sql.Array#getArray()`, but the JDBC wrapper inherited `Object#toString()`, which produced an unreadable class name and identity hash.

Fixes #3045.

## Changes

`com.clickhouse.jdbc.types.Array#toString()` now uses deep array formatting via `Arrays.deepToString(...)`.

The value is now rendered as:

[[550e8400-e29b-41d4-a716-446655440000]]

Nested arrays and tuple values are also rendered correctly.

## Compatibility

No public API signatures, JDBC return types, or array contents were changed.

This change only improves the textual representation returned by `com.clickhouse.jdbc.types.Array#toString()`.

## Tests

Added:

- `ArrayTest#testToStringForArrayOfNamedTuples`
  - Verifies deep formatting of `Array(Tuple(id UUID))`.
  - Passed: 1 test, 0 failures.

- `ArrayTupleIntegrationTest#testArrayOfNamedAndUnnamedTuplesToString`
  - Runs against ClickHouse using Docker.
  - Covers both named and unnamed tuple arrays.
  - Verifies `ResultSet#getObject()` returns `java.sql.Array`.
  - Verifies `toString()` matches the contents returned by `Array#getArray()`.
  - Passed: 1 test, 0 failures.


mshustov:
@claude review

jdbc-v2/src/main/java/com/clickhouse/jdbc/types/Array.java
```diff
@@ -127,6 +128,16 @@ private void ensureValid() throws SQLException {
...
     }
 
+    /**
+     * Returns a readable representation of the array contents. JDBC clients such as IDE database viewers
+     * commonly use {@code toString()} when rendering values returned by {@link java.sql.ResultSet#getObject}.
+     * Deep formatting is required for arrays of tuples and nested arrays.
+     */
+    @Override
+    public String toString() {
+        return Arrays.deepToString(array);
```
chernser: this can be on someones hot path and will result performance degradation. 
All rendering such kind should be done by standalone class. It would be more safe and useful. 
Besides here we may not control form of the result so approach will be useful only for some cases. 
please implement Array to string formatter if needed.

chernser:
@014-code 
However initial issue seems need more research. I do not want  to make `toString()` traverse deep arrays. It may be another issue.

014-code:
> @014-code Thank you for the contribution! However initial issue seems need more research. I do not want to make `toString()` traverse deep arrays. It may be another issue.

Thank you for the clarification.

I have withdrawn the `Array#toString()` deep-formatting approach, including its tests and changelog entry. I agree that traversing nested arrays from `toString()` may affect a hot path, and the resulting representation would not establish a sufficiently clear or controllable JDBC display contract.

I also merged the latest `main` into this branch and pushed the update.

I will close this PR for now. The original behavior in #3045 needs further investigation and a separately agreed approach before introducing any array or tuple display formatting.

PR #3067, opened 2026-08-19
Fix JDBC parser parameter counting for empty comments and SELECT * EXCEPT

### LINKED ISSUE #3052: jdbc-v2 0.9.7: SQL parser silently reports argCount=0 for two constructs (bare `--` comment; `SELECT * EXCEPT (...)`), causing ArrayIndexOutOfBoundsException on bind
## Description

With `clickhouse-jdbc` **0.9.7** (jdbc-v2), two SQL constructs make the client-side parser return `argCount == 0` even though the statement contains a `?` placeholder. Binding then fails with:

```
java.lang.ArrayIndexOutOfBoundsException: Index 0 out of bounds for length 0
	at com.clickhouse.jdbc.PreparedStatementImpl.setInt(PreparedStatementImpl.java:174)
```

Both statements are accepted by the server without complaint — the failure is purely in the client-side parser.

This looks like the same family as #2461 / #2478 / #2568 (all `area:sql-parser`, fixed in 0.9.1 / 0.9.3), but these two triggers are still present in 0.9.7.

Interestingly the two triggers affect **different parser implementations**, so which one bites depends on which parser is selected.

### Trigger A — a bare `--` comment line breaks `ANTLR4Parser`

A line comment consisting of exactly two dashes and **nothing after them** (not even a space):

```sql
--
select count(*) from numbers(10) where number = ?
```

`-- ` (with a trailing space), `--x`, and `/**/` all parse fine. Position does not matter — the same happens when the empty `--` line is inside a subquery:

```sql
select count(*) from (
--
select 1 as a) x where x.a = ?
```

### Trigger B — `SELECT * EXCEPT (...)` breaks `ANTLR4AndParamsParser`

```sql
select count(*) from ( select * EXCEPT (b), b as c from ( select 1 as a, 2 as b ) ) x where x.a = ?
```

## Results

Measured by calling the parser directly (see below), `argCount` for each case:

| SQL | `ANTLR4Parser` | `ANTLR4AndParamsParser` |
|---|---|---|
| `select count(*) from numbers(10) where number = ?` (baseline) | 1 | 1 |
| bare `--` comment line | **0** | 1 |
| `-- ` (dash dash space) | 1 | 1 |
| `--x` | 1 | 1 |
| `/**/` | 1 | 1 |
| bare `--` inside a subquery | **0** | 1 |
| `SELECT * EXCEPT (...)` | 1 | **0** |

## How this was measured

No server needed — the parser is client-side, so it can be driven directly:

```java
Class<?> c = Class.forName("com.clickhouse.jdbc.internal.SqlParserFacade$ANTLR4Parser");
Constructor<?> ct = c.getDeclaredConstructor(boolean.class);
ct.setAccessible(true);
Object p = ct.newInstance(true);
Method parse = c.getMethod("parsePreparedStatement", String.class);
parse.setAccessible(true);
Method argCount = Class.forName("com.clickhouse.jdbc.internal.ParsedPreparedStatement")
        .getMethod("getArgCount");

System.out.println(argCount.invoke(parse.invoke(p, "--\nselect count(*) from numbers(10) where number = ?")));
// prints 0
```

## Why this is painful to diagnose

The parse failure is silent. `argCount` is simply 0, and nothing is logged at the point of failure. The symptom only appears later, at bind time, as an `ArrayIndexOutOfBoundsException` from `PreparedStatementImpl`, which gives no hint that the SQL failed to parse.

In our case the SQL came from a Hibernate `@Subselect` view definition: queries **without** a `WHERE` parameter worked fine, and the moment any filter was added the request failed with the AIOOBE above. That combination points investigators at the condition-building code rather than at the driver.

Two suggestions, independent of the parse bugs themselves:

1. When the parser fails or finds no parameters while the raw SQL clearly contains `?`, log a warning (or throw a `SQLException` describing the parse failure) instead of silently producing `argCount = 0`.
2. `PreparedStatementImpl.setXxx` could raise a `SQLException` with the parameter index and the detected count, rather than letting the raw `ArrayIndexOutOfBoundsException` escape.

## Environment

- `com.clickhouse:clickhouse-jdbc:0.9.7` (all)
- JDK 25 (GraalVM CE 25.0.4)
- Reproduced offline against the parser only, so server version is not involved

n200534:
## Summary

- fix single-line comment scanning when a bare `--` is followed immediately by a newline, so later JDBC parameter markers are still discovered
- extend the bundled ANTLR4 grammar to accept `SELECT * EXCEPT (...)`, including identifier-list and regular-expression forms
- add regression coverage for top-level and nested empty comments and `SELECT * EXCEPT (...)` across all three SQL parser backends
- document the user-visible fix in `CHANGELOG.md`

Closes #3052

## Root cause

`ClickHouseUtils.skipSingleLineComment` treated a newline located exactly at its start index as if no newline existed, causing the parameter scanner to skip the rest of the SQL. Separately, the bundled ANTLR4 grammar accepted `*` in a select list but did not model ClickHouse's `EXCEPT` modifier, so the parse-tree-based backend never reached a later `?` marker.

## User impact and compatibility

Prepared statements containing either construct now report the correct parameter count and can bind normally instead of failing with `ArrayIndexOutOfBoundsException`. This does not change public APIs, configuration, wire protocols, or binary compatibility.

## Validation

- `mvn -pl jdbc-v2 -am -DskipTests install` — passed under JDK 17
- `mvn -pl jdbc-v2 test` — 1,331 tests passed
- `mvn -pl clickhouse-data test` — 1,554 tests passed, 113 conditionally skipped, 0 failures/errors

## Checklist

- [x] Closes #3052
- [x] Unit tests cover the affected parser scenarios
- [x] A human-readable changelog entry was added

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -102,6 +102,9 @@
   inserted data or failing with a server-side `SYNTAX_ERROR`. Escape sequences are now recognized only outside of quoted
   text, and a `{fn ...}` escape is unwrapped at its matching closing brace, so nested braces (e.g. a `{name:Type}` query
   parameter or a nested escape) stay balanced. (https://github.com/ClickHouse/clickhouse-java/issues/2995)
+- **[jdbc-v2]** Fixed prepared statements losing parameter markers after an empty `--` comment line or after
+  `SELECT * EXCEPT (...)`, which caused parameter binding to fail with `ArrayIndexOutOfBoundsException` for the
+  affected SQL parser backends. (https://github.com/ClickHouse/clickhouse-java/issues/3052)
 - **[client-v2]** Fixed LZ4 input streams not closing their underlying HTTP response stream. Closing an LZ4 stream
   returned by `QueryResponse.getInputStream()` now releases the wrapped transport stream, including after a partial
   read. (https://github.com/ClickHouse/clickhouse-java/issues/2985)
--- a/clickhouse-data/src/main/java/com/clickhouse/data/ClickHouseUtils.java
+++ b/clickhouse-data/src/main/java/com/clickhouse/data/ClickHouseUtils.java
@@ -1150,7 +1150,7 @@ public final class ClickHouseUtils {
      */
     public static int skipSingleLineComment(String args, int startIndex, int len) {
         int index = args.indexOf('\n', startIndex);
-        return index > startIndex ? index + 1 : len;
+        return index >= startIndex ? index + 1 : len;
     }
 
     /**
--- a/clickhouse-data/src/test/java/com/clickhouse/data/ClickHouseUtilsTest.java
+++ b/clickhouse-data/src/test/java/com/clickhouse/data/ClickHouseUtilsTest.java
@@ -239,6 +239,9 @@ public class ClickHouseUtilsTest {
                 args.indexOf('\n') + 1);
         Assert.assertEquals(ClickHouseUtils.skipSingleLineComment(args, args.indexOf("--", 11), args.length()),
                 args.length());
+
+        args = "--\nselect 1";
+        Assert.assertEquals(ClickHouseUtils.skipSingleLineComment(args, 2, args.length()), 3);
     }
 
     @Test(groups = { "unit" })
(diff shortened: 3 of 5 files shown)


mshustov:
@cursor review

chernser (approved):

PR #3096, opened 2026-09-02
[DOCS] Observability 

### LINKED ISSUE #2147: Document metrics usage
We need to document how people can use [Micrometer](https://micrometer.io/) (or other metrics) in some way - whether that be java code examples, or pure documentation.

We should definitely provide copy + paste-able code (because that's the easiest to ingest), but it doesn't necessarily need to be 100% code.

chernser:
## Summary
- Adds `examples/demo-spring-service` as showcase for observability 
- Adds Observability section to the main documentation 
- Adds Observability section to the partner spec 
- Adds `docs/integration-ops.md` as part of partner spec  for operations 

Closes: https://github.com/ClickHouse/clickhouse-java/issues/2147

## Checklist
- [ ] Closes #<issue ref>


docs/integration-client.md
```diff
@@ -966,6 +967,172 @@ A [`Session`](../client-v2/src/main/java/com/clickhouse/client/api/Session.java)
 
 ---
 
+## Observability & Monitoring
+
+The ClickHouse Java Client (V2) provides built-in observability features designed around OpenTelemetry semantic conventions and Micrometer metrics. The client enables monitoring operational durations, throughput, retries, server execution statistics, connection pool status, and distributed tracing spans.
+
+For a complete production-style Spring Boot application demonstrating Client V2 configuration, connection pooling, trace context propagation across `@Async` boundaries, and full telemetry integration, see the `examples/demo-spring-service` module in this repository.
+
+### Client Builder Observability API
```
mshustov, line 1036: This overlaps with `docs/clickhouse-docs/client.mdx` with similar content. I suggest consolidating it into a single document and referencing other pages. I’m concerned about content discrepancies over time

docs/integration-ops.md
```diff
@@ -0,0 +1,269 @@
...
+
+import java.util.concurrent.TimeUnit;
+
+public Client createOptimizedClient() {
+    return new Client.Builder()
+        .addEndpoint("http://localhost:8123")
+        .setUsername("default")
+        .setPassword("secret")
+        .setMaxConnections(50)
+        .setConnectionRequestTimeout(5, TimeUnit.SECONDS)
```
mshustov: not `ChronoUnit`? as per  https://github.com/ClickHouse/clickhouse-java/blob/40464dde1c7f3de958a765a756ef6f4c70e98ad3/client-v2/src/test/java/com/clickhouse/client/HttpTransportTests.java#L244

examples/demo-spring-service/README.md
```diff
@@ -0,0 +1,221 @@
+# IoT Ingest Demo Service
```
mshustov, line 1: we already have another Spring demo service https://github.com/ClickHouse/clickhouse-java/tree/40464dde1c7f3de958a765a756ef6f4c70e98ad3/examples/demo-service. Should we delete ore rename the old one?

docs/integration-ops.md
```diff
@@ -0,0 +1,269 @@
...
+
+Because both layers share the same Apache HttpClient HTTP transport stack, connection pooling, metrics collection, and distributed tracing work consistently across both direct Java Client usage and JDBC driver connections.
+
+---
+
+## Connection Pools & Resource Management
+
+### HTTP Connection Pooling (`client-v2`)
+
+The `Client` instance owns an internal Apache HttpClient 5 connection pool. It manages persistent HTTP connections to ClickHouse endpoints.
```
mshustov: should we add a link to `Apache HttpClient 5` docs? one time is enough

docs/integration-ops.md
```diff
@@ -0,0 +1,269 @@
...
+}
+```
+
+### JDBC Pooling Considerations (HikariCP & Frameworks)
+
+When using the JDBC driver (`clickhouse-jdbc`) with an external connection pooler like **HikariCP**:
+
+- **Layering:** HikariCP manages JDBC `Connection` instances, while each JDBC connection wraps a `Client` instance with its own internal HTTP socket pool.
+- **Sizing Alignment:** Avoid over-allocating HikariCP connections. Because ClickHouse processes HTTP requests concurrently over pooled sockets, a smaller HikariCP pool (e.g., 10–20 connections) paired with a properly sized `Client` HTTP pool is typically optimal.
+- **Connection Lifecycle:** Configure HikariCP's `maxLifetime` slightly shorter than any network or load-balancer idle timeout to prevent stale socket exceptions.
```
mshustov: > than any network or load-balancer idle timeout to prevent stale socket exceptions.

Do we have any anecdotal values to recommend?

docs/integration-ops.md
```diff
@@ -0,0 +1,269 @@
...
+
+The `Client` instance owns an internal Apache HttpClient 5 connection pool. It manages persistent HTTP connections to ClickHouse endpoints.
+
+#### Key Pool Settings
+
+| Setting | Builder Method | Default | Description |
+|---------|----------------|---------|-------------|
+| Max Connections | `setMaxConnections(int)` | `10` | Maximum open HTTP connections per server endpoint. |
+| Connection TTL | `setConnectionTTL(long, TimeUnit)` | `-1` (disabled) | Time-to-live after which an active connection is closed and recreated. |
+| Keep-Alive Timeout | `setKeepAliveTimeout(long, TimeUnit)` | Server default | HTTP Keep-Alive duration for idle pooled connections. |
```
mshustov: any recommendations for aligning the value with server config value? AFAIK it should be lower than https://clickhouse.com/docs/reference/settings/server-settings/settings/other#keep_alive_timeout to avoid an attempt to reuse a server-closed connection

docs/integration-ops.md
```diff
@@ -0,0 +1,269 @@
...
+- Monitor the `httpcomponents.httpclient.pool.total.pending` gauge in Grafana / Prometheus. A non-zero or spiking pending count indicates thread contention for HTTP sockets.
+- Compare `httpcomponents.httpclient.pool.total.connections{state="leased"}` against `httpcomponents.httpclient.pool.total.max`.
+
+**Remediation:**
+- Increase `Client.Builder.setMaxConnections(...)` to match concurrent thread demand.
+- Ensure all query and insert response objects (`QueryResponse`, `InsertResponse`) are closed promptly using `try-with-resources`.
+
+### 2. Correlating Application Traces with ClickHouse Server Logs (`system.query_log`)
+
+**Symptom:** Need to trace an expensive or failing query from APM traces down to ClickHouse server execution logs.
```
mshustov: not only APM traces - any query, no?

docs/clickhouse-docs/client.mdx
```diff
@@ -899,6 +899,173 @@ try (QueryResponse response = client.query("SELECT * FROM my_table").get()) {
...
+        long durationMs = metrics.getMetric(ClientMetrics.OP_DURATION).getLong();
+    }
+}
+```
+
+### Spans Reported {#v2-o11y-spans}
+
+Tracing in Client V2 produces two types of spans arranged in a parent-child hierarchy:
+
+1. **Operation Span**: Represents a high-level client operation (`query <database>` or `insert <database>.<table_name>`). Started under the current thread's trace context (`Context.current()`) so that database operations join the caller's active trace span.
```
mshustov: > (`query <database>` or `insert <database>.<table_name>`)

not related to this PR: there is inconsistency across language clients that we need to address. see https://github.com/ClickHouse/clickhouse-rs/pull/390

mshustov (approved):

docs/integration-client.md
chernser: @mshustov 
I agree with you. Initially I was thinking this will be good to have some duplication so no context is mixed. 
Will put here specific part and reference to the client document.

examples/demo-spring-service/README.md
chernser: yes, we need to delete old one later. 
https://github.com/ClickHouse/clickhouse-java/issues/3111

docs/integration-ops.md
chernser: Why not. 
I still think it will overflow context in some way and LLM will anyway search documentation by name but we will prune while testing.

docs/integration-ops.md
chernser: good point. I will compose some example.

docs/integration-ops.md
chernser: we need to recommend to not align actually because client should auto-config in this case. 
I will record it for troubleshooting guide.

docs/integration-ops.md
chernser: yes.

docs/clickhouse-docs/client.mdx
chernser: This is very related to this PR because we need to fix it right now to avoid changing names in the future. 
We can easily fix insert and make it like in RS client.  
The problem is with query because `[10] db.operation.name: The operation name SHOULD NOT be extracted from db.query.text.` and on the client it is problematic. 
In RS client query seems has no value https://github.com/ClickHouse/clickhouse-rs/pull/390/changes#diff-f0c63cae3ed0328ebbe14ac22d86b43bc8a40146944b03c8326681e4f3802443R177 

So we can remove it from our code also - in Java client user can customize observability attributes if needed

chernser: Will fix the naming and will remove `db.operation.name` from execute and query spans . Will decide later.

docs/integration-ops.md
chernser: Fixed.

docs/clickhouse-docs/client.mdx
chernser: Fixed.

docs/integration-client.md
chernser: Fixed.

docs/integration-ops.md
chernser: fixed.

docs/integration-ops.md
chernser: I will remove this section because has dedicated issue for that

docs/integration-ops.md
chernser: Fixed this section.

docs/integration-ops.md
chernser: done. moved recommendation to integration-client

PR #3101, opened 2026-09-04
[jdbc-v2,client-v2] Fix format selection and document how to 

### LINKED ISSUE #3086: [docs] Update regarding priority of `X-ClickHouse-Format` header
### Description 
There is upcoming changes in ClickHouse about priority of `X-ClickHouse-Format` and `FORMAT` clause. 
This changes should be very well highlighted in documentation and specs.

chernser:
## Summary

- Fixes client to allow override format in query
- Sets default format on client level and not on operation level to allow override it by query 
- Updates all documentation and tests

Closes: https://github.com/ClickHouse/clickhouse-java/issues/3086
## Checklist
- [x] Closes #<issue ref>


client-v2/src/main/java/com/clickhouse/client/api/ClientConfigProperties.java
```diff
@@ -349,7 +349,7 @@ public Object parseValue(String value) {
         if (valueType.isEnum()) {
             Object[] constants = valueType.getEnumConstants();
             for (Object constant : constants) {
-                if (constant.toString().equals(value)) {
+                if (constant.toString().equalsIgnoreCase(value.trim())) {
```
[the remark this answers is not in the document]
chernser: fixed - empty string is becoming null.

chernser:
@cursor review

docs/integration-jdbc.md
```diff
@@ -402,26 +402,75 @@ public boolean checkConnectionHealth(Connection conn, int timeoutSeconds) throws
...
 | Batch INSERT | Multi-row SQL rewrite or RowBinary | Depends on statement shape |
 
+### Format Selection and SQL `FORMAT` Clauses
+
+The response format can be configured using the `format` connection property (`ClientConfigProperties.INPUT_OUTPUT_FORMAT` or `"format"`).
+
+**Important for ClickHouse 26.8+:**
+- On ClickHouse 26.8+, the request format header sent by the driver (`X-ClickHouse-Format`) takes priority over a `FORMAT` clause written in the SQL query string.
+- By default, the driver sends `format=RowBinaryWithNamesAndTypes`.
+- To use a SQL `FORMAT` clause (such as `SELECT ... FORMAT JSONEachRow`) with ClickHouse 26.8+, set `format=JSONEachRow` in connection properties or set `format=` (to `""` empty string) so the default binary format header is omitted and ClickHouse honors the query's `FORMAT` clause.
```
Marais: Recommending `format=` (empty) for a JDBC connection has a wider effect than this bullet says. With no header the server falls back to `default_format` (TabSeparated) for every statement that has no `FORMAT` clause, and `StatementImpl` rejects text formats. Verified on 26.8.1 with a scratch test on a connection configured with `format=`:

- `execute` / `executeUpdate` for DDL and `INSERT ... VALUES`: work
- `executeQuery("SELECT 1")` and `execute("SELECT 1")`: `SQLException: Only RowBinaryWithNameAndTypes and JSONEachRow are supported for output format`
- `DatabaseMetaData.getTables` / `getColumns` / `getSchemas`: same `SQLException`

So a connection configured this way only works for statements that carry an explicit `FORMAT` clause, and any tool that touches metadata breaks. Suggest either stating this explicitly here (and in `docs/features.md` line 110 and the CHANGELOG entry), or recommending `format=JSONEachRow` as the JDBC way to read JSON and treating `format=` as an expert-only setting.

Related: the "Export to a file format" row at line 473 has the same issue. Only `RowBinaryWithNamesAndTypes` and `JSONEachRow` are consumable by a `ResultSet`; `format=CSV` makes every query fail, which `testUnsupportedFormat` shows. A small follow-up that would make this failure self-explanatory: have the `StatementImpl` error name the format actually received and point at the `format` property.

client-v2/src/main/java/com/clickhouse/client/api/Client.java
```diff
@@ -1301,6 +1300,17 @@ public Builder setMetricsRecorder(MetricsRecorder metricsRecorder) {
             return this;
         }
 
+        /**
+         * Sets default format used when no format is specified in {@code QuerySettings}.
+         *
+         * @param format - valid ClickHouse format
+         * @return this instance of builder
+         */
+        public Builder queryFormat(String format) {
```
Marais, line 1316: Since this becomes permanent public API in 0.11.0, I'd settle its shape before merging:

- `ClickHouseFormat.valueOf` is case-sensitive, so `queryFormat("csv")` throws while `setOption("format", "csv")` now succeeds thanks to the case-insensitive parsing added in this PR.
- `null` throws an NPE, so the method cannot express the new "send no format header" mode this PR introduces; callers have to fall back to `setOption(INPUT_OUTPUT_FORMAT.getKey(), null)`.
- An invalid name surfaces the raw `No enum constant ...` message rather than a `ClientMisconfigurationException`.

Suggestion: take `ClickHouseFormat` instead of `String`, with `null` meaning "no header", e.g. `queryFormat(ClickHouseFormat format)` storing `format == null ? null : format.name()`. Alternatively route the string through `INPUT_OUTPUT_FORMAT.parseValue(...)` so both entry points agree. Either way a test for `null` and for an invalid value would be good; the only coverage today is the `CSV` happy path in `ClientTests.testDefaultSettings`.

Marais (approved):

docs/integration-jdbc.md
chernser: Thank you for the comment - I will make changes in this code and documentation.

client-v2/src/main/java/com/clickhouse/client/api/Client.java
chernser: Using `ClickHouseFormat` require us to update code each time new format added to ClickHouse. 
So I'm thinking to fix the issue with that.

docs/integration-jdbc.md
chernser: Fixed the metadata and setting the RowBinary
Also mentioned in the documentation and migration guide

client-v2/src/main/java/com/clickhouse/client/api/Client.java
chernser: fixed checks and case sensitivity.

client-v2/src/main/java/com/clickhouse/client/api/ClientConfigProperties.java
```diff
@@ -347,9 +347,20 @@ public Object parseValue(String value) {
...
+            String configValue = value.trim();
+            if (configValue.isEmpty()) {
+                return null;
+            }
+            if (valueType.equals(ClickHouseFormat.class)) {
+                try {
+                    return ClickHouseFormat.fromString(configValue);
+                } catch (IllegalArgumentException e) {
+                    return configValue;
+                }
```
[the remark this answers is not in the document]
chernser: Fixed.

jdbc-v2/src/main/java/com/clickhouse/jdbc/metadata/DatabaseMetaDataImpl.java
```diff
@@ -1833,7 +1851,7 @@
                 "FROM system.functions " +
                 "WHERE name LIKE '" + (functionNamePattern == null ? "%" : functionNamePattern) + "'";
         try {
-            return connection.createStatement().executeQuery(sql);
+            return createStatement().executeQuery(sql);
```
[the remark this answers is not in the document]
chernser: this is planned work.

PR #3142, opened 2026-09-18
Fix ArrayResultSet after-last cursor state

### LINKED ISSUE #3143: [jdbc-v2] Fix ArrayResultSet after-last cursor state
### Description 

ResultSet that is built around an array handles `next()` call incorrectly. JDBC documentation states that call should progress further and if `next()` returns `false` then `isAfterLast` also should return false.

fallintoplace:
## Summary

**What**
- Move a non-empty `ArrayResultSet` to after-last when `next()` returns `false`.

**Why**
- The cursor stayed on the last row after exhaustion.
- `isAfterLast()`, `isLast()`, `getRow()`, and `previous()` could report the wrong state.

**Implementation**
- Set `pos` to `length` on non-empty exhaustion.
- Keep empty arrays before-first.
- Add regression coverage for the cursor state and moving back to the last row.

Closes: https://github.com/ClickHouse/clickhouse-java/issues/3143


jdbc-v2/src/main/java/com/clickhouse/jdbc/types/ArrayResultSet.java
```diff
@@ -99,6 +99,9 @@ private static ClickHouseColumn elementColumn(ClickHouseColumn column) {
     @Override
     public boolean next() throws SQLException {
         if (pos + 1 >= length || length == 0) {
+            if (length > 0) {
```
chernser: this is the problem not only for nonEmpty arrays.

chernser (changes requested):
Please correct for all cases - documentation says that next() should progress unconditionally. And when next() returns false - result set should be in state `AfterLast` 

> Moves the cursor forward one row from its current position. A ResultSet cursor is initially positioned before the first row; the first call to the method next makes the first row the current row; the second call makes the second row the current row, and so on.
When a call to the next method returns false, the cursor is positioned after the last row. Any invocation of a ResultSet method which requires a current row will result in a SQLException being thrown. If the result set type is TYPE_FORWARD_ONLY, it is vendor specified whether their JDBC driver implementation will return false or throw an SQLException on a subsequent call to next.

chernser:
@fallintoplace 

chernser (approved):

chernser (approved):

